PR #8, opened 2017-09-20
Add two helper methods: isDirectory and list

mertemin:
This PR adds two helper methods `isDirectory` and `list` to `SparkFileHelper` class. These methods would be useful to check is a given path is directory or not and for directories list resources under.


src/test/java/org/openstreetmap/atlas/checks/persistence/SparkFileHelperTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        // Try to delete directory
+        TEST_HELPER.deleteDirectory(path);
+    }
+
+    @Test
+    public void testIsDirectory() throws IOException
+    {
+        // Test a file
+        final File tempFile = File.createTempFile("test", ".tmp");
```
MikeGost: Nit: You could use `FileSuffix.TEMPORARY` value instead of `".tmp"` String for these tests.

PR #12, opened 2017-11-15
Making unique identifer function in BaseCheck more robust. Updating S…

mgcuthbert:
…park helper functions.

For the unique identifier function, updating so that it takes an AtlasObject object instead of AtlasEntity. If the object is of type AtlasEntity then it will use the type otherwise will use the simplified class name.

he SparkFileHelper was extended with a commit() method that checks if the given output directory exists, renaming the individual files when it does.

A timestamp is appended to the metrics filename to create a unique metrics file for each run.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
@@ -292,31 +292,48 @@ public abstract class BaseCheck<T> implements Check, Serializable
     }
 
     /**
-     * Generates a unique identifier given an {@link AtlasEntity}. OSM/Atlas objects with different
+     * Generates a unique identifier given an {@link AtlasObject}. OSM/Atlas objects with different
      * types can share the identifier (way 12345 - node 12345). This method makes sure we generate a
      * truly unique identifier, based on the OSM identifier, among different types for an
-     * {@link AtlasEntity}.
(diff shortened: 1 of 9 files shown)


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -292,31 +292,48 @@ protected void markAsFlagged(final T identifier)
...
-     *            {@link AtlasEntity} to generate unique identifier for
+     * @param object
+     *            {@link AtlasObject} to generate unique identifier for
      * @return unique object identifier among different types
      */
-    protected String getUniqueObjectIdentifier(final AtlasEntity entity)
+    protected String getUniqueObjectIdentifier(final AtlasObject object)
     {
-        return String.format("%s%s", entity.getType().toShortString(), entity.getIdentifier());
+        if (object instanceof AtlasEntity)
```
MikeGost: The separation between Type and Class exists to support ComplexEntity type of checks, correct? If so, do you think it's simpler to use `object.getClass().getSimpleName()` for all unique identifiers or is there another reason for this?

mgcuthbert: Yes it is to support ComplexEntities. The only reason to differentiate between the two is because the majority of these checks are going to be on AtlasEntity objects and in that case the `toShortString()` method will return a single character which in general is going to keep that value short and clean. However other than that there isn't any other reason. However I would like to keep it that way for that reason.

PR #15, opened 2017-12-05
Adding instruction localization for Atlas Check Flags

mgcuthbert:
An example configuration
```
"BuildingRoadIntersectionCheck":
{
"enabled":true,
  "locale":"es",
  "flags":{
    "en":["Building (id-{0}) intersects road (id-{1})"],
    "es":["Edificio(id-{0}) cruza calle(id-{1})"]
  }
}
```
Results in Instructions like:
"1. Edificio(id-465465839) cruza calle(id-443627253)"

Where a configuration like this:
```
"BuildingRoadIntersectionCheck":
{
"enabled":true,
  "locale":"en",
  "flags":{
    "en":["Building (id-{0}) intersects road (id-{1})"],
    "es":["Edificio(id-{0}) cruza calle(id-{1})"]
  }
}
```
If no flags are provided in the configuration then `getLocalizedInstruction(int, Object…)` in BaseCheck will return the instruction provided by `getFallbackInstructructions()`. If a specific language set of instructions does not have all instructions then the fallback instructions will be used. If the specified language does not have instructions at all, then the default language instructions will be used. Default Locale (as specified by language) currently set to English. 

When there are multiple Instructions, they are all added to a list whose key is the two letter language abbreviation(as seen above). Instructions that have the same meaning in different languages should be in the same spot of the list for all languages. The instructions themselves are formatted using the MessageFormat class, where the number in the bracket refers to an index in an varargs object array passed to the `getLocalizedInstruction(int, Object…)` method.

Note: Many changes are simply IDE formatting.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
@@ -1,8 +1,10 @@
 package org.openstreetmap.atlas.checks.base;
 
 import java.io.Serializable;
+import java.text.MessageFormat;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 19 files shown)


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -259,6 +307,14 @@ protected CheckFlag createFlag(final Set<AtlasObject> objects, final String inst
 
     protected abstract Optional<CheckFlag> flag(AtlasObject object);
 
+    /**
+     * Method to implement for inheriting checks to return the default set of instruction formats
+     * that will be the last resort in {@link BaseCheck#getLocalizedInstruction(int, Object[])}
+     *
+     * @return The set of instructions to fall back to if configuration results in none.
+     */
+    protected abstract List<String> getFallbackInstructions();
```
mertemin: Let's implement this as a default method and return empty list. I don't think we need to force checks to implement a fallback instruction as this is the last resort to get a instruction.

mgcuthbert: On principle I certainly agree, however currently most of the instructions are set in the check itself, so unless we plan on enforcing the instructions to be set through configuration I think we have to enforce the check to have a fallback in the code. If we do enforce it in configuration then we could just get rid of this function altogether. Either way I think this is a good discussion to have but maybe not for this PR.

PR #18, opened 2018-01-08
Fixed ID formatting

jklamer:
MessageFormat by default formats all `long` types with comma seperated thousands, which is an incorrect representation of an OSM ID. Changed the instruction formats to specify the formatting of any OSM IDs.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/OrphanNodeCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/OrphanNodeCheck.java
@@ -18,8 +18,8 @@ import org.openstreetmap.atlas.utilities.configuration.Configuration;
  */
 public class OrphanNodeCheck extends BaseCheck<Long>
 {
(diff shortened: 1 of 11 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/points/OrphanNodeCheck.java
```diff
@@ -18,8 +18,8 @@
  */
 public class OrphanNodeCheck extends BaseCheck<Long>
 {
-    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
-            .asList("Node with OSM ID {0} is an orphan, no tags and not connected to any ways.");
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            "Node with OSM ID {0,number,#} is an orphan, no tags and not connected to any ways.");
```
MikeGost: Nit: Was looking at this check recently. The instruction reads strangely. Maybe "... is an orphan - it doesn't have any tags, is not connected to any ways and is not part of a Relation".

PR #20, opened 2018-02-08
Configuration overrides for countries and groups of countries in Integrity Checks

jklamer:
The PR utilizes keyword overrides to create per country/group configuration options. The check resource loader initializes a set of checks for every country using a configuration specific to the countries overrides and the groups overrides.

The new group interaction can be found as an example below

```
{
  "CheckResourceLoader.scanUrls": ["org.openstreetmap.atlas.checks.base.checks"],
  "groups":{
    "beginningAlphabet":["ABC","DEF"],
    "alphabetEnds":["ABC","XYZ"]
  },
  "CheckResourceLoaderTestCheck":{
    "enabled": true,
    "var1":1,"var2":"Hi",
    "override.beginningAlphabet.var1":2,
    "override.alphabetEnds.var2":"Bye"
    }
}
```
The above configuration will override var1 for ABC and DEF, and var2 for ABC and XYZ. The test can be found in the PR.

**Dependent on https://github.com/osmlab/atlas/pull/77**

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
@@ -8,10 +8,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
@@ -114,9 +111,10 @@ public class CheckResourceLoader
         Configuration specializedConfiguration = this.configuration
                 .configurationForKeyword(country);
 
-        if (this.countryGroups.containsKey(country))
+        final List<String> groups = this.countryGroups.get(country);
+        if (groups != null)
         {
-            for (final String group : this.countryGroups.get(country))
+            for (final String group : groups)
             {
                 specializedConfiguration = specializedConfiguration.configurationForKeyword(group);
             }
@@ -139,16 +137,15 @@ public class CheckResourceLoader
      *
      * @param isEnabled
      *            {@link Predicate} used to determine if a check is enabled
-     * @param specificConfiguration
-     *            {@link Configuration} used to loadChecks if not the one used to initialize
-     *            {@link CheckResourceLoader}
+     * @param configuration
+     *            {@link Configuration} used to loadChecks {@link CheckResourceLoader}
      * @param <T>
      *            check type
      * @return a {@link Set} of checks
      */
     @SuppressWarnings("unchecked")
     public <T extends Check> Set<T> loadChecks(final Predicate<Class> isEnabled,
-            final Configuration specificConfiguration)
+            final Configuration configuration)
     {
         final Set<T> checks = new HashSet<>();
         final Time time = Time.now();
@@ -170,7 +167,7 @@ public class CheckResourceLoader
                                 try
                                 {
                                     check = checkClass.getConstructor(Configuration.class)
-                                            .newInstance(specificConfiguration);
+                                            .newInstance(configuration);
                                 }
                                 catch (final InvocationTargetException oops)
                                 {
@@ -220,18 +217,15 @@ public class CheckResourceLoader
         return loadChecks(this::isEnabledByConfiguration, this.configuration);
     }
 
-    public <T extends Check> Map<String, Set<T>> loadChecksForCountries(
-            final Predicate<Class> isEnabled, final Iterable<String> countries)
+    public <T extends Check> Set<T> loadChecksForCountry(final Predicate<Class> isEnabled,
+            final String country)
     {
-        return StreamSupport.stream(countries.spliterator(), false).collect(Collectors.toMap(
-                Function.identity(),
-                country -> loadChecks(isEnabled, this.getConfigurationForCountry(country))));
+        return loadChecks(isEnabled, this.getConfigurationForCountry(country));
     }
 
-    public <T extends Check> Map<String, Set<T>> loadChecksForCountries(
-            final Iterable<String> countries)
+    public <T extends Check> Set<T> loadChecksForCountry(final String country)
     {
-        return loadChecksForCountries(this::isEnabledByConfiguration, countries);
+        return loadChecksForCountry(this::isEnabledByConfiguration, country);
     }
 
     private boolean isEnabledByConfiguration(final Class checkClass)
--- a/src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
@@ -5,12 +5,18 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.maps.MultiMap;
 import org.openstreetmap.atlas.utilities.time.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,19 +44,20 @@ import com.google.common.reflect.ClassPath;
  * </pre>
  *
  * @author brian_l_davis
+ * @author jklamer
  */
 public class CheckResourceLoader
 {
     private static final String DEFAULT_ENABLED_KEY_TEMPLATE = "%s.enabled";
-    private static final Logger logger = LoggerFactory.getLogger(CheckResourceLoader.class);
     private static final String DEFAULT_PACKAGE = "org.openstreetmap.atlas.checks";
     private static final String DEFAULT_TYPE = Check.class.getName();
-
-    private final Configuration configuration;
+    private static final Logger logger = LoggerFactory.getLogger(CheckResourceLoader.class);
     private final Class<?> checkType;
-    private final Set<String> packages;
+    private final Configuration configuration;
+    private final MultiMap<String, String> countryGroups = new MultiMap<>();
     private final Boolean enabledByDefault;
     private final String enabledKeyTemplate;
+    private final Set<String> packages;
 
     /**
      * Default constructor
@@ -64,6 +71,12 @@ public class CheckResourceLoader
         this.packages = Collections.unmodifiableSet(Iterables.asSet((Iterable<String>) configuration
                 .get("CheckResourceLoader.scanUrls", Collections.singletonList(DEFAULT_PACKAGE))
                 .value()));
+        final Map<String, List<String>> groupCountries = configuration
+                .get("groups", Collections.emptyMap()).value();
+        groupCountries.keySet().forEach(group ->
+        {
+            groupCountries.get(group).forEach(country -> this.countryGroups.add(country, group));
+        });
 
         final ClassLoader loader = Thread.currentThread().getContextClassLoader();
         try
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -70,39 +70,35 @@
     private static final Switch<String> COUNTRIES = new Switch<>("countries",
             "Comma-separated list of country ISO3 codes to be processed", StringConverter.IDENTITY,
             Optionality.REQUIRED);
+    // Indicator key for ignored countries
+    private static final String IGNORED_KEY = "Ignored";
```
mertemin: Can you disable your IDE's parameter reordering and retry? Stylistic changes are making it harder to understand the actual difference between files.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -64,6 +71,12 @@ public CheckResourceLoader(final Configuration configuration)
         this.packages = Collections.unmodifiableSet(Iterables.asSet((Iterable<String>) configuration
                 .get("CheckResourceLoader.scanUrls", Collections.singletonList(DEFAULT_PACKAGE))
                 .value()));
+        final Map<String, List<String>> groupCountries = configuration
+                .get("groups", Collections.emptyMap()).value();
+        groupCountries.keySet().forEach(group ->
+        {
```
mgcuthbert: The extra braces here are unnecessary.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -403,9 +404,11 @@ protected SwitchList switches()
      *            set of {@link BaseCheck}s to execute
      * @return {@code true} if sanity check passes, {@code false} otherwise
      */
-    private boolean isValidInput(final StringList countries, final Set<BaseCheck> checksToExecute)
+    private boolean isValidInput(final StringList countries,
+            final Map<String, Set<BaseCheck>> checksToExecute)
     {
-        if (countries.size() == 0 || checksToExecute.size() == 0)
+        if (countries.size() == 0
+                || checksToExecute.values().stream().anyMatch(checks -> checks.size() == 0))
```
mertemin: I'd suggest avoiding the bulk load operation. That would be hard to debug. Consider loading configuration without any overrides and checking if that is valid or not in this method. Then, per country read in group/country specific configuration and use that for execution.

mertemin: I'd suggest avoiding the bulk load operation. That would be hard to debug. Consider loading configuration without any overrides and checking if that is valid or not in this method. Then, per country read in group/country specific configuration and use that for execution.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -206,24 +207,24 @@ public void start(final CommandMap commandMap)
...
+                .forEach(country -> countryCheckTuples
+                        .add(new Tuple2<>(country, checksToExecute.get(country))));
 
         // Log countries and integrity
         logger.info("Initialized countries: {}", countryCheckTuples.stream().map(tuple -> tuple._1)
                 .collect(Collectors.joining(",")));
-        logger.info("Initialized checks: {}", checksToExecute.stream().map(BaseCheck::getCheckName)
-                .collect(Collectors.joining(",")));
+        logger.info("Initialized checks: {}", checksToExecute.values().stream().flatMap(Set::stream)
+                .map(BaseCheck::getCheckName).distinct().collect(Collectors.joining(",")));
```
mertemin: This log message is not as useful any more. Due to overrides `checksToExecute` set is in a different state.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -64,6 +71,12 @@ public CheckResourceLoader(final Configuration configuration)
         this.packages = Collections.unmodifiableSet(Iterables.asSet((Iterable<String>) configuration
                 .get("CheckResourceLoader.scanUrls", Collections.singletonList(DEFAULT_PACKAGE))
                 .value()));
+        final Map<String, List<String>> groupCountries = configuration
```
mertemin: Variable name should be more generic.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -83,29 +96,59 @@ public CheckResourceLoader(final Configuration configuration)
...
      *
-     * @param <T>
-     *            check type
-     * @return a {@link Set} of checks
+     * @param country
+     *            country string
+     * @return {@link Configuration}
      */
-    public <T extends Check> Set<T> loadChecks()
+    public Configuration getConfigurationForCountry(final String country)
```
mertemin: Method should be more generic. Group doesn't have to be a country.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -83,29 +96,59 @@ public CheckResourceLoader(final Configuration configuration)
...
+     *            {@link Configuration} used to loadChecks if not the one used to initialize
+     *            {@link CheckResourceLoader}
      * @param <T>
      *            check type
      * @return a {@link Set} of checks
      */
     @SuppressWarnings("unchecked")
-    public <T extends Check> Set<T> loadChecks(final Predicate<Class> isEnabled)
+    public <T extends Check> Set<T> loadChecks(final Predicate<Class> isEnabled,
+            final Configuration specificConfiguration)
```
mertemin: `specificConfiguration` name is misleading. In what way it is specific?

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -165,6 +208,32 @@ public CheckResourceLoader(final Configuration configuration)
...
+     *            check type
+     * @return a {@link Set} of checks
+     */
+    public <T extends Check> Set<T> loadChecks()
+    {
+        return loadChecks(this::isEnabledByConfiguration, this.configuration);
+    }
+
+    public <T extends Check> Map<String, Set<T>> loadChecksForCountries(
+            final Predicate<Class> isEnabled, final Iterable<String> countries)
```
mertemin: Method name, parameter name for keys are misleading.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: ```groups``` work?

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: In this case this function is only designed for countries. The group keywords are retrieved using the country string passed in here.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: It is the configuration made specifically for a country.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
jklamer: This will still log all the checks loaded for at least one country. Which I agree is no longer that useful, but what would be a good alternative? Should we log at each county distributed job?

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: What would be a better name? This function loads the checks for each country in the string list.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
mertemin: This method is loading checks for given configuration. It is oblivious to overrides. So I'd suggest leaving naming as-is.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -196,34 +196,38 @@ public void start(final CommandMap commandMap)
...
+        final Set<BaseCheck> preOverriddenChecks = checkLoader.loadChecks();
+        if (!isValidInput(countries, preOverriddenChecks))
         {
             logger.error("No countries supplied or checks enabled, exiting!");
             return;
         }
 
+        // Load checks
+        final Map<String, Set<BaseCheck>> checksToExecutePerCountry = checkLoader
+                .loadChecksForCountries(countries);
```
mertemin: Are collecting configuration for all the groups necessary here? I'd suggest updating `loadChecksForCountries` method to `loadChecksForGroup` to load checks with overridden group key. We should not hold onto such data structure here. Create as needed and pass in.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -196,34 +196,38 @@ public void start(final CommandMap commandMap)
...
-                .value();
+                .get("priority.countries", Collections.EMPTY_LIST).value();
 
         // Create a list of Country to Check tuples
         // Add priority countries first if they are supplied by parameter
         final List<Tuple2<String, Set<BaseCheck>>> countryCheckTuples = new ArrayList<>();
-        countries.stream().filter(priorityCountries::contains)
-                .forEach(country -> countryCheckTuples.add(new Tuple2<>(country, checksToExecute)));
+        countries.stream().filter(priorityCountries::contains).forEach(country -> countryCheckTuples
+                .add(new Tuple2<>(country, checksToExecutePerCountry.get(country))));
```
mertemin: Same here. Just load the overridden configuration.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -254,6 +258,9 @@ public void start(final CommandMap commandMap)
             final String country = tuple._1();
             final Set<BaseCheck> checks = tuple._2();
 
+            logger.info("Initialized checks for {}: {}", country,
+                    checks.stream().map(BaseCheck::getCheckName).collect(Collectors.joining(",")));
```
mertemin: Nice!

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -83,29 +96,59 @@ public CheckResourceLoader(final Configuration configuration)
...
-    public <T extends Check> Set<T> loadChecks()
+    public Configuration getConfigurationForCountry(final String country)
     {
-        return loadChecks(this::isEnabledByConfiguration);
+        Configuration specializedConfiguration = this.configuration
+                .configurationForKeyword(country);
+
+        if (this.countryGroups.containsKey(country))
+        {
+            for (final String group : this.countryGroups.get(country))
```
mertemin: Although hash table is small here, you are doing lookup twice if given country exists in the map: one with `containsKey` and another with `get` call. I'd suggest using `get` and checking against null values.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
mertemin: I'd suggest keeping it generic and using `group` instead of `country`.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -64,6 +71,12 @@ public CheckResourceLoader(final Configuration configuration)
         this.packages = Collections.unmodifiableSet(Iterables.asSet((Iterable<String>) configuration
                 .get("CheckResourceLoader.scanUrls", Collections.singletonList(DEFAULT_PACKAGE))
                 .value()));
+        final Map<String, List<String>> groups = configuration.get("groups", Collections.emptyMap())
+                .value();
+        groups.keySet().forEach(group ->
+        {
+            groups.get(group).forEach(country -> this.countryGroups.add(country, group));
```
mertemin: Just curious - what is the value of using `MultiMap` over `Map<String, List<String>>`?

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
mertemin: I'd suggest keeping it generic and aligned with override definition - groups. Is this method used anywhere though?

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: Thats not what I use it for. The Groups are taken into account within the function. The configuration will get an override for whatever group it is a part of. As defined by the base configuration.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: This version of the function is used within the more Generic ```loadChecksForCountries(Iterable<String> countries)``` which is then used within ```IntegrityCheckSparkJob```

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
jklamer: I don't have to handle initializing the list when I add key value pairs

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
jklamer: Oh I see just load that during tuple creation.

PR #21, opened 2018-02-22
Polygon filter Per Check

jklamer:
Adds a [AtlasEntityPolygonsFilter](https://github.com/osmlab/atlas/pull/85) for every check and is implemented in the BaseCheck class. Each Check configuration body can now specify polygons in which it will/or will not run. 

Dependent on https://github.com/osmlab/atlas/pull/85 before merging.


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -97,15 +100,23 @@ public BaseCheck(final Configuration configuration)
                     .registerTypeAdapter(Challenge.class, new ChallengeDeserializer()).create();
             this.challenge = gson.fromJson(gson.toJson(challengeMap), Challenge.class);
         }
+        this.polygonFilter = AtlasEntityPolygonsFilter.forConfigurationValues(
+                configurationValue(configuration, AtlasEntityPolygonsFilter.INCLUDED_POLYGONS_KEY,
+                        new HashMap<>()),
```
mertemin: Please use `Collections.emptyMap()`

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -97,15 +100,23 @@ public BaseCheck(final Configuration configuration)
...
+                configurationValue(configuration,
+                        AtlasEntityPolygonsFilter.INCLUDED_MULTIPOLYGONS_KEY, new HashMap<>()),
+                configurationValue(configuration, AtlasEntityPolygonsFilter.EXCLUDED_POLYGONS_KEY,
+                        new HashMap<>()),
```
mertemin: Same here

PR #22, opened 2018-02-24
Remove SparkFileHelper, Output, and FilePath and references the ones in atlas-generator

AlexHsieh22:
As part of https://github.com/osmlab/atlas-generator/pull/13 , SparkFileHelper, Output, and FilePath were ported over to atlas-generator.  With this PR, we're referencing the new atlas-generator version and updating all references of SparkFileHelper, Output, and FilePath to the atlas-generator ones.


src/test/java/org/openstreetmap/atlas/checks/event/MetricFileGeneratorTest.java
```diff
@@ -26,30 +26,30 @@
...
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
     @Test
-    public void testFileName() throws IOException
+    public void testFileLabelWithKnownSuffix() throws IOException
     {
-        final File tempDirectory = tempFolder.newFolder();
-        final MetricFileGenerator generator = new MetricFileGenerator("some-metrics-log.csv",
+        final File tempDirectory = this.tempFolder.newFolder();
+        final MetricFileGenerator generator = new MetricFileGenerator("some.metrics.txt",
```
mgcuthbert: Why the change in the filename? I understand it is just a test, just curious as to the reasoning behind it.

AlexHsieh22: This is part of the reordering.  Eclipse is reordering all method to be alphabetical.  So it's not a change in filename, it's that the ordering of the tests are different.  The ```testFileName()``` test case is now found on line 62.

PR #23, opened 2018-02-26
SelfIntersectingPolylineCheck enhancements

danielduhh:
The main goal of this PR is to improve the clarity of flags by classifying various cases of self-intersection.

- Line features w/ building tags
- Area features w/ duplicate Edges
- Area features w/ invalid geometries 
- Roads w/ invalid geometries

This also removes all flags containing following tags Highway tags: _PROPOSED, CONSTRUCTION, FOOTWAY, PATH_, as well as all Waterway tags.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
@@ -4,7 +4,9 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 
+import org.openstreetmap.atlas.checks.atlas.predicates.TagPredicates;
 import org.openstreetmap.atlas.checks.base.BaseCheck;
 import org.openstreetmap.atlas.checks.flag.CheckFlag;
 import org.openstreetmap.atlas.checks.validation.GeometryValidator;
@@ -12,10 +14,15 @@ import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.Location;
 import org.openstreetmap.atlas.geography.PolyLine;
 import org.openstreetmap.atlas.geography.Polygon;
+import org.openstreetmap.atlas.geography.Segment;
 import org.openstreetmap.atlas.geography.atlas.items.Area;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
 import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Line;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.Taggable;
+import org.openstreetmap.atlas.tags.WaterwayTag;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,10 +37,25 @@ public class SelfIntersectingPolylineCheck extends BaseCheck<Long>
 {
     private static final String INSTRUCTION_SHORT = "Self-intersecting polyline for feature {0,number,#}";
     private static final String INSTRUCTION_LONG = INSTRUCTION_SHORT + " at {1}";
+    private static final String AREA_INSTRUCTION = "Feature {0,number,#} has invalid geometry at {1}";
+    private static final String AREA_BUILDING_INSTRUCTION = "Feature {0,number,#} is an incomplete "
+            + "building at {1}";
+    public static final String DUPLICATE_EDGE_INSTRUCTION = "Feature {0,number,#} has duplicate Edges";
     private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(INSTRUCTION_SHORT,
-            INSTRUCTION_LONG);
+            INSTRUCTION_LONG, AREA_INSTRUCTION, AREA_BUILDING_INSTRUCTION,
+            DUPLICATE_EDGE_INSTRUCTION);
     private static final Logger logger = LoggerFactory
             .getLogger(SelfIntersectingPolylineCheck.class);
+    // Excluded Highway tags
+    private static final Predicate<Taggable> INELIGIBLE_HIGHWAY_TAGS = object -> Validators
+            .isOfType(object, HighwayTag.class, HighwayTag.PROPOSED)
+            || Validators.isOfType(object, HighwayTag.class, HighwayTag.CONSTRUCTION)
+            || Validators.isOfType(object, HighwayTag.class, HighwayTag.FOOTWAY)
+            || Validators.isOfType(object, HighwayTag.class, HighwayTag.PATH);
+    // Excluded Waterway tags
+    private static final Predicate<Taggable> WATERWAY_TAGS = object -> Validators
+            .hasValuesFor(object, WaterwayTag.class);
+    public static final Integer THREE = 3;
     private static final long serialVersionUID = 2722288442633787006L;
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
```diff
@@ -25,15 +33,29 @@
...
+    private static final String AREA_INSTRUCTION = "Feature {0,number,#} has invalid geometry at {1}";
+    private static final String POLYLINE_BUILDING_INSTRUCTION = "Feature {0,number,#} is a incomplete "
+            + "building at {1}";
+
+    private static final String DUPLICATE_EDGE_INSTRUCTION = "Feature {0,number,#} has a duplicate "
+            + "Edge at {1}";
+    private static final String POLYLINE_INSTRUCTION = "Self-intersecting polyline for feature "
+            + "{0,number,#} at {1}";
+    // Excluded Highway tags
+    private static final Predicate<Taggable> INELIGIBLE_HIGHWAY_TAGS = object -> Validators
```
mgcuthbert: I think it would be better if we controlled this in the configuration using the tagFilter like:
```
"tags.filter": "highway->!proposed&highway->!construction&highway->!footway&highway->!path"
```

This way if someone wanted to include those they could just modify the configuration instead of having to actually modify the code.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
```diff
@@ -61,27 +83,41 @@ public SelfIntersectingPolylineCheck(final Configuration configuration)
...
-                || object instanceof Line;
+        // Master edges excluding ineligible highway tags
+        return ((object instanceof Edge && ((Edge) object).isMasterEdge()
+                && !INELIGIBLE_HIGHWAY_TAGS.test(object))
+                // Areas
+                || (object instanceof Area)
+                // Lines excluding ineligible highway tags
+                || (object instanceof Line && !INELIGIBLE_HIGHWAY_TAGS.test(object)))
+                // Exclude waterway tags
+                && !Validators.hasValuesFor(object, WaterwayTag.class);
```
mgcuthbert: What is the reasoning behind excluding waterways?

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
danielduhh: Great idea, am I able to list multiple values w/ the `!` operator? : 

```
"tags.filter": "highway->!proposed,!construction,!footway,!path"
```

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
danielduhh: - OSM has no guidelines on handling Waterway self intersections
- Waterways have unique modeling techniques that warrant it's own inspection of geometry/tag combinations
- Waterways are not used for navigation

Unless there's an issue with Atlas handling self intersecting Waterway geometries, I'd suggest keeping them out of this check

danielduhh:
Closing for further updates

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
danielduhh: The original check uses `tags.Filter` to eliminate all features not containing `highway`, `building`, or `waterway` tags. I intend to keep that initial filter, however, I'm unsure how to include 
```
highway->!construction&highway->!footway&highway->!path
``` 
into the original `highway->*|building->*` clause in order to make tag filtering fully dynamic.

Features must contain highway or building tags & if they contain highway tags, they cannot be one of the following: construction, footway, or path.

mgcuthbert: Sorry for replying so late, basically "&"'s are grouped together. So to do what you want, you can do this: 
```highway->*&highway->!construction&highway->!footway&highway->!path|building->*```

This will kind of treat it like this:
```(highway->*&highway->!construction&highway->!footway&highway->!path)|building->*```

Note the braces should not be included, just giving you an idea of precedence. Also the specific highway tags need to be included after the wildcard highway tag.

Comment (danielduhh) [reply src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java]: Awesome. Thanks for the clarification... re opening w/ this update

**Comment by danielduhh** [comment]
Also refactored unit tests for extensibility

File: src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java [hunk 3]
```diff
@@ -144,4 +170,36 @@ else if (object instanceof Area)
...
+            {
+                // If adjacent segment is the same value as our current, duplicate exists
+                if (adjacentEdge < segments.size() && segment.equals(segments.get(adjacentEdge)))
+                {
+                    duplicates.add(segment);
+                }
+            }
+        }
+
+        return duplicates.size() > 0;
```
mertemin: Test against empty should be sufficient here

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
```diff
@@ -144,4 +170,36 @@ else if (object instanceof Area)
...
+        final List<Segment> duplicates = new ArrayList<>();
+
+        // Loop through Polyline Segments
+        for (int i = 0; i < segments.size(); i++)
+        {
+            final Segment segment = segments.get(i);
+            final int adjacentEdge = i + 2;
+
+            // Check if segment exists elsewhere in List
+            if (segments.indexOf(segment) != segments.lastIndexOf(segment))
```
mertemin: What is the advantage of checking `segments.indexOf(segment) != segments.lastIndexOf(segment)` here. If there is a duplicate segment, then `adjacentEdge < segments.size() && segment.equals(segments.get(adjacentEdge))` would be true anyways. If `adjacentEdge < segments.size() && segment.equals(segments.get(adjacentEdge))` is true, then `segments.indexOf(segment) != segments.lastIndexOf(segment)` must be true. So `adjacentEdge < segments.size() && segment.equals(segments.get(adjacentEdge))` should be sufficient for duplicate check.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
mertemin: Wait you don't need to keep track of `duplicates`. Whenever a duplicate is found inside `for` loop, just return true. Otherwise, false could be returned after `for` loop.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
```diff
@@ -144,4 +170,36 @@ else if (object instanceof Area)
...
+     *            the {@link PolyLine} being examined
+     * @return {@code true} if the any set of adjacent edges have identical geometries
+     */
+    private boolean hasDuplicateEdges(final PolyLine polyline)
+    {
+        final List<Segment> segments = polyline.segments();
+        final List<Segment> duplicates = new ArrayList<>();
+
+        // Loop through Polyline Segments
+        for (int i = 0; i < segments.size(); i++)
```
mertemin: Please avoid using single letter variables to keep code style consistent with the rest of the code base.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/SelfIntersectingPolylineCheck.java
danielduhh: Yes, good find. `segments.indexOf(segment) != segments.lastIndexOf(segment)` returns true if the segment exists elsewhere. In that case, I should be able to:

```java
if (segments.indexOf(segment) != segments.lastIndexOf(segment))
{
    return true;
}
```

PR #26, opened 2018-03-14
Roundabout Valence Check

savannahostrowski:
Brief summary of the check:

This check identifies roundabouts that have an unusual valence. By default, unusual valence is defined as less than 2 or greater than 14. However, these values are configurable, and can be
adjusted by setting the connections.minimum and connections.maximum in the configuration file.

This PR includes: documentation, logic, and associated unit tests.


src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheckTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Tests for {@link RoundaboutValenceCheck}
+ *
+ * @author savannahostrowski
+ */
+public class RoundaboutValenceCheckTest
+{
+
+    private final RoundaboutValenceCheck check = new RoundaboutValenceCheck(
```
jklamer: Is this used to test against? If not, you can use it for a test with different configuration values which would be nice to have.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,187 @@
...
+    public static final String WRONG_VALENCE_INSTRUCTIONS = "This roundabout, {0, number, #}, "
+            + "has the wrong valence. It has a valence of {1, number, integer}.";
+    public static final String VALENCE_OF_ONE_INSTRUCTIONS = "This feature, {0,number,#},"
+            + " should not be labelled as a roundabout. "
+            + "This feature should be a turning loop or turning circle.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(WRONG_VALENCE_INSTRUCTIONS, VALENCE_OF_ONE_INSTRUCTIONS);
+
+    private static final double LOWER_VALENCE_THRESHOLD_DEFAULT = 2.0;
+    private static final double UPPER_VALENCE_THRESHOLD_DEFAULT = 10.0;
```
jklamer: Can you update this to match the description? Or is this as intended?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,187 @@
...
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Edge edge = (Edge) object;
+        final Map<Long, Edge> roundaboutEdges = new HashMap<>();
+
+        // Get all edges in the roundabout
+        getAllRoundaboutEdges(edge, roundaboutEdges);
+        final int totalRoundaboutValence;
```
jklamer: You can simplify this code by using some Java8 streams in this case. Something along the lines of
```
final Set<Edge> connectedEdges =  roundaboutEdges.values().stream()
.flatMap(roundaboutEdge -> roundaboutEdge.connectedEdges().stream()).filter(HighwayTag::isCarNavigableHighway).filter(Edge::isMasterEdge)
.filter(e -> ! JunctionTag.isRoundabout(e)).filter(e -> ! roundaboutEdges.values().contains(e)).collect(Collectors.toSet());
final int totalRoundaboutValence = connectedEdges.size();
```

One thing to note:  ```.filter(e -> ! JunctionTag.isRoundabout(e))``` and 
```.filter(e -> ! roundaboutEdges.values().contains(e))``` effectively due the same thing so long as ```getAllRoundaboutEdges``` always returns edges the only follow the ```JunctionTag.isRoundabout(e)``` condition. I believe the stronger design however would be to return a Set<Edge> from ```getAllRoundaboutEdges``` and then use the filter ```.filter(e -> ! roundaboutEdges.contains(e))``` in place of either of the two above. The advantages of this are 
<li>generating a values() set from the returned map every time in the filter is a performance cost </li>
<li> you don't have to rely on keeping parity with ```getAllRoundaboutEdges``` for what qualifies as a roundabout edge</li>

File: src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java [hunk 4]
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * This method gets all edges in a roundabout given one edge in that roundabout
+     * 
+     * @param edge
+     *            An Edge object known to be a roundabout edge
+     * @param roundaboutEdges
+     *            A map that contains key, value pairs of Edge identifiers and all associated Edge
+     *            data
+     */
+    private void getAllRoundaboutEdges(final Edge edge, final Map<Long, Edge> roundaboutEdges)
```
jklamer: For the reasons above can you refactor this to ```private Set<Edge> getAllRoundaboutEdges(final Edge edge)```

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,187 @@
...
+        // Mark the current node as visited and enqueue it
+        this.markAsFlagged(edge.getIdentifier());
+        queue.add(edge);
+
+        // As long as the queue is not empty
+        while (queue.size() != 0)
+        {
+            // Dequeue a connected edge and add it to the roundaboutEdges
+            final Edge e = queue.poll();
+            if (e.getIdentifier() < 0)
```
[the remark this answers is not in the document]
savannahostrowski: Yeah, that's correct. Is using `!e.isMasterEdge()` what you mean?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,187 @@
...
+            roundaboutEdges.put(e.getIdentifier(), e);
+
+            // Get the edges connected to the edge e as an iterator
+            final Iterator<Edge> iterator = e.connectedEdges().iterator();
+            while (iterator.hasNext())
+            {
+                final Edge connectedEdge = iterator.next();
+                final Long edgeId = connectedEdge.getIdentifier();
+
+                if (JunctionTag.isRoundabout(connectedEdge))
```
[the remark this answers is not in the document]
savannahostrowski: Sweet. I dig it.

docs/checks/roundaboutValenceCheck.md
```diff
@@ -0,0 +1,103 @@
...
+                while (iterator.hasNext())
+                {
+                    final Edge connectedEdge = iterator.next();
+                    final Long edgeId = connectedEdge.getIdentifier();
+    
+                    if (JunctionTag.isRoundabout(connectedEdge))
+                    {
+                        if (!roundaboutEdges.containsKey(edgeId)
+                            && edgeId != edge.getIdentifier()
+                            && !this.isFlagged(edgeId))
```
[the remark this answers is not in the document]
savannahostrowski: Updated.

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheckTest.java
savannahostrowski: Whoops, forgot to actually use that configuration 😺 . Added some unit tests using an increased number of maximum connections. Let me know if you had something different in mind.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
savannahostrowski: Can you clarify? @jklamer

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
savannahostrowski: Oh, I like this. So succinct. 👍

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
savannahostrowski: Refactored. Let me know if this wasn't what you were thinking!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
jklamer: In the description it says "By default, unusual valence is defined as less than 2 or greater than 14". And I just wanted to make sure the ```UPPER_VALENCE_THRESHOLD_DEFAULT = 10``` is intended.

savannahostrowski: Oh yeah, that's intended.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // We check that the object is an instance of Edge
+        return object instanceof Edge
+                // And that the Edge has not already been marked as flagged
+                && !this.isFlagged(object.getIdentifier())
+                // Make sure that the edges are instances of roundabout
+                && JunctionTag.isRoundabout(object);
```
mertemin: If an edge is encountered, then whether it is roundabout or not should be checked first. Because, check doesn't work on non-roundabout edges. This condition needs to filter first. Otherwise, many edges will be checked whether they are flagged or not first before checking if they are roundabout or not.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+     *            An Edge object known to be a roundabout edge
+     * @return A set of edges in the roundabout
+     */
+    private Set<Edge> getAllRoundaboutEdges(final Edge edge)
+    {
+        final Set<Edge> roundaboutEdges = new HashSet<>();
+        // Initialize a queue to add yet to be processed connected edges to
+        final Queue<Edge> queue = new LinkedList<>();
+
+        // Mark the current node as visited and enqueue it
```
mertemin: Comment needs to be updated.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+        this.markAsFlagged(edge.getIdentifier());
+        queue.add(edge);
+
+        // As long as the queue is not empty
+        while (queue.size() != 0)
```
mertemin: Check if queue is empty or not via `isEmpty` method.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+        {
+            // Dequeue a connected edge and add it to the roundaboutEdges
+            final Edge e = queue.poll();
```
mertemin: Please avoid using single letter variable names.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
mertemin: You can further filter if an edge is master or not here.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+
+            // Check to make sure that we are only adding master edges into the set
+            if (!e.isMasterEdge())
```
mertemin: Queue can keep only master edges. Edges could be filtered before being added to the queue. Then, this check could be removed.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+            {
+                continue;
+            }
+
+            roundaboutEdges.add(e);
+
+            // Get the edges connected to the edge e as an iterator
+            final Iterator<Edge> iterator = e.connectedEdges().iterator();
```
mertemin: I'd suggest using a for loop unless there is an advantage of using iterators here. `connectedEdges` method gives you a set to iterator over right away.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+    {
+        final Edge edge = (Edge) object;
+
+        // Get all edges in the roundabout
+        final Set<Edge> roundaboutEdges = getAllRoundaboutEdges(edge);
+
+        final Set<Edge> connectedEdges = roundaboutEdges.stream()
+                .flatMap(roundaboutEdge -> roundaboutEdge.connectedEdges().stream())
+                .filter(HighwayTag::isCarNavigableHighway).filter(Edge::isMasterEdge)
+                .filter(e -> !JunctionTag.isRoundabout(e)).filter(e -> !roundaboutEdges.contains(e))
```
mertemin: Please avoid using single letter variables to keep styling consistent within code base.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
```diff
@@ -0,0 +1,167 @@
...
+                .collect(Collectors.toSet());
+        final int totalRoundaboutValence = connectedEdges.size();
+
+        // If the totalRoundaboutValence is less than the minimum configured number of connections
+        // or greater than or equal to the maximum configured number of connections
+        if (totalRoundaboutValence < this.minimumValence
+                || totalRoundaboutValence >= this.maximumValence)
```
mertemin: I'd suggest keeping border conditions consistent with each other. Either treat them inclusive or not, but do the same for both.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java
savannahostrowski: So you mean that `JunctionTag.isRoundabout(object)` and `!this.isFlagged(object.getIdentifier())` should be switched?

Nice! I like the idea of adding master edge filtering here.

mertemin: Yes, filter roundabouts first.

PR #27, opened 2018-03-20
New Check: AreasWithHighwayTag

mgcuthbert:
### Overview
Checks to see if an Area is attributed with a highway tag. “Area” in this context refers to the Atlas Area Object, not the explicit OSM area definition.  _The tag combination of highway=PEDESTRIAN and area=YES (or any Atlas-defined Area with the highway=PEDESTRIAN tag) is the one Valid exception._  

### Requirements
* Rule 1: An Area cannot have a _highway_ tag attached to it.
* Rule 2: Any case that is flagged for Rule 1 should have the _highway_ tag removed.
* Rule 3: The tag combination of _highway=PEDESTRIAN_ and _area=YES_ is **VALID**, and this should be the only exception.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/areas/AreasWithHighwayTagCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/areas/AreasWithHighwayTagCheck.java
@@ -1,0 +1,67 @@
+package org.openstreetmap.atlas.checks.validation.areas;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.atlas.items.Area;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+import org.openstreetmap.atlas.tags.AreaTag;
+import org.openstreetmap.atlas.tags.HighwayTag;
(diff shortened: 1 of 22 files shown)


danielduhh:
I suggest including
```json
"tag.filters":"highway->!pedestrian&area->!yes"
```
as part of `AreasWithHighwayTag ` configuration

mgcuthbert:
Yes great point, thanks.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/AreasWithHighwayTagCheck.java
```diff
@@ -0,0 +1,67 @@
...
+        return object instanceof Area;
+    }
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        // See configuration for initial check filter; highway=pedestrian & area=yes is a valid
+        // Area/Highway tag combination.
+        final Area area = (Area) object;
+        final Optional<String> highwayTag = area.getTag(HighwayTag.KEY);
```
jklamer: you can check for this in the validCheckForObject method with 
```object instances Area && ((Area) object).getTag(HighwayTag.KEY).isPresent()```. That way all the validity checking is happening in that function and only the flagging is happening in this function.

mgcuthbert: I am going to let a separate PR be dealt with related to this comment.

PR #28, opened 2018-03-22
Address Point Match Check

savannahostrowski:
This check identifies Point objects in OSM that have a specified street number (addr:housenumber) 
but no specified street name (addr:street). No specified street name refers to either having a null
value for the street name key, or no street name key present at all.

This pull request includes: all check logic, unit tests, and documentation.


danielduhh:
impressive!!!!


Will review by the end of the week....

PR #31, opened 2018-03-27
Address Point Match Check

savannahostrowski:
This check identifies Point objects in OSM that have a specified street number (addr:housenumber)
but no specified street name (addr:street). No specified street name refers to either having a null
value for the street name key, or no street name key present at all.

This pull request includes: all check logic, unit tests, and documentation.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
@@ -1,0 +1,136 @@
+package org.openstreetmap.atlas.checks.validation.points;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.Rectangle;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.tags.AddressHousenumberTag;
+import org.openstreetmap.atlas.tags.AddressStreetTag;
+import org.openstreetmap.atlas.tags.names.NameTag;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * Tests for {@link AddressPointMatchCheck}
+ *
+ * @author savannahostrowski
+ */
+
+public class AddressPointMatchCheck extends BaseCheck
+{
+    private static final long serialVersionUID = 1L;
+    public static final String NO_STREET_NAME_POINT_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. The street name should likely "
+            + "be one of {1}. These names were derived from nearby Points.";
+    public static final String NO_STREET_NAME_EDGE_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. The street name should likely "
+            + "be one of {1}. These names were derived from nearby Edges.";
+    public static final String NO_SUGGESTED_NAMES_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. No suggestions names were found.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            NO_STREET_NAME_POINT_INSTRUCTIONS, NO_STREET_NAME_EDGE_INSTRUCTIONS,
+            NO_SUGGESTED_NAMES_INSTRUCTIONS);
+    private static final String ADDRESS_STREET_NUMBER_KEY = AddressHousenumberTag.KEY;
+    private static final String POINT_STREET_NAME_KEY = AddressStreetTag.KEY;
+    private static final String EDGE_STREET_NAME_KEY = NameTag.KEY;
+    private static final double BOUNDS_SIZE_DEFAULT = 150.0;
+
+    private final Distance boundsSize;
+
+    @Override
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    public AddressPointMatchCheck(final Configuration configuration)
+    {
+        super(configuration);
+        this.boundsSize = Distance.meters(
+                (Double) configurationValue(configuration, "bounds.size", BOUNDS_SIZE_DEFAULT));
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // Object is an instance of Point
+        return object instanceof Point
+                // And has a street number specified
+                && object.getTag(ADDRESS_STREET_NUMBER_KEY).isPresent()
+                // And if the street name key has a value of null
+                && (!object.getTag(POINT_STREET_NAME_KEY).isPresent()
+                        // Or if the street name key is not present
+                        || !object.getTags().containsKey(POINT_STREET_NAME_KEY));
+    }
+
+    /**
+     * This is the actual function that will check to see whether the object needs to be flagged.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return an optional {@link CheckFlag} object that
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Point point = (Point) object;
+        // Get a bounding box around the Point of interest
+        final Rectangle box = point.getLocation().boxAround(boundsSize);
+
+        // Get all Points inside the bounding box
+        final Iterable<Point> interiorPoints = point.getAtlas().pointsWithin(box);
+        // Convert the Iterable into a Stream for filtering, remove Points that have null as their
+        // street name as they cannot be candidates for a street for the Point of interest
+        final Set<Point> points = StreamSupport.stream(interiorPoints.spliterator(), false)
+                .filter(interiorPoint -> interiorPoint.getTag(POINT_STREET_NAME_KEY).isPresent())
+                .collect(Collectors.toSet());
+
+        // Get all Edges that intersect or are contained within the bounding box
+        final Iterable<Edge> interiorEdges = point.getAtlas().edgesIntersecting(box);
+        // Convert the Iterable into a Stream for filtering, remove Edge that have null as their
+        // street name as they cannot be candidates for a street for the Point of interest
+        final Set<Edge> edges = StreamSupport.stream(interiorEdges.spliterator(), false)
+                .filter(interiorEdge -> interiorEdge.getTag(EDGE_STREET_NAME_KEY).isPresent())
+                .collect(Collectors.toSet());
+
+        final Set<String> streetNameList = new HashSet<>();
+
+        // If there are no Points or Edges in the bounding box
+        if (points.isEmpty() && edges.isEmpty())
+        {
+            // Flag Point with instruction indicating that there are are no suggestions
+            return Optional.of(this.createFlag(point,
+                    this.getLocalizedInstruction(2, point.getOsmIdentifier())));
+        }
+        // If there are Points in the bounding box
+        else if (!points.isEmpty())
+        {
+            // Add all interior Point street names to the list of candidate street names
+            points.forEach(interiorPoint -> streetNameList
+                    .add(interiorPoint.getTags().get(POINT_STREET_NAME_KEY)));
+            return Optional.of(this.createFlag(point,
+                    this.getLocalizedInstruction(0, point.getOsmIdentifier(), streetNameList)));
+
+        }
+        // If there are Edges intersecting or contained by the bounding box
+        else
+        {
+            // Add all interior Edge street names to the list of candidate street names
+            edges.forEach(interiorEdge -> streetNameList
+                    .add(interiorEdge.getTags().get(EDGE_STREET_NAME_KEY)));
+            return Optional.of(this.createFlag(point,
+                    this.getLocalizedInstruction(1, point.getOsmIdentifier(), streetNameList)));
+        }
+
+    }
+}
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
@@ -8,20 +8,27 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import org.apache.directory.api.util.Strings;
 import org.openstreetmap.atlas.checks.base.BaseCheck;
 import org.openstreetmap.atlas.checks.flag.CheckFlag;
 import org.openstreetmap.atlas.geography.Rectangle;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
 import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.geography.atlas.items.Relation;
 import org.openstreetmap.atlas.tags.AddressHousenumberTag;
 import org.openstreetmap.atlas.tags.AddressStreetTag;
+import org.openstreetmap.atlas.tags.RelationTypeTag;
 import org.openstreetmap.atlas.tags.names.NameTag;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
 import org.openstreetmap.atlas.utilities.scalars.Distance;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas;
 
 /**
- * Tests for {@link AddressPointMatchCheck}
+ * This check identifies Point objects in OSM that have a specified street number (addr:housenumber)
+ but no specified street name (addr:street). No specified street name refers to either having a null
+ value for the street name key, or no street name key present at all.
  *
  * @author savannahostrowski
  */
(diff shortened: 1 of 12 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // Object is an instance of Point
+        return object instanceof Point
+                // And has a street number specified
+                && object.getTag(ADDRESS_STREET_NUMBER_KEY).isPresent()
+                // And if the street name key has a value of null
+                && (!object.getTag(POINT_STREET_NAME_KEY).isPresent()
+                        // Or if the street name key is not present
+                        || !object.getTags().containsKey(POINT_STREET_NAME_KEY));
```
mertemin: The condition (`!object.getTag(POINT_STREET_NAME_KEY).isPresent() || !object.getTags().containsKey(POINT_STREET_NAME_KEY)`) checking for street name tag could be simplified by `Strings.isEmpty(object.tag(POINT_STREET_NAME_KEY))`.

`object.tag(POINT_STREET_NAME_KEY)` would return `null` if tag does not exist. If it exists, it's value would be returned. `Strings.isEmpty` would return true if given value is null or empty string.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+
+        // Get all Edges that intersect or are contained within the bounding box
+        final Iterable<Edge> interiorEdges = point.getAtlas().edgesIntersecting(box);
+        // Convert the Iterable into a Stream for filtering, remove Edge that have null as their
+        // street name as they cannot be candidates for a street for the Point of interest
+        final Set<Edge> edges = StreamSupport.stream(interiorEdges.spliterator(), false)
+                .filter(interiorEdge -> interiorEdge.getTag(EDGE_STREET_NAME_KEY).isPresent())
+                .collect(Collectors.toSet());
+
+        final Set<String> streetNameList = new HashSet<>();
```
mertemin: This doesn't have to be declared here.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+            // Flag Point with instruction indicating that there are are no suggestions
+            return Optional.of(this.createFlag(point,
+                    this.getLocalizedInstruction(2, point.getOsmIdentifier())));
+        }
+        // If there are Points in the bounding box
+        else if (!points.isEmpty())
+        {
+            // Add all interior Point street names to the list of candidate street names
+            points.forEach(interiorPoint -> streetNameList
+                    .add(interiorPoint.getTags().get(POINT_STREET_NAME_KEY)));
```
mertemin: Street names could be collected by `points.forEach(point -> point.tag(POINT_STREET_NAME_KEY)).collect(Collectors.toSet())`.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+            return Optional.of(this.createFlag(point,
+                    this.getLocalizedInstruction(0, point.getOsmIdentifier(), streetNameList)));
+
+        }
+        // If there are Edges intersecting or contained by the bounding box
+        else
+        {
+            // Add all interior Edge street names to the list of candidate street names
+            edges.forEach(interiorEdge -> streetNameList
+                    .add(interiorEdge.getTags().get(EDGE_STREET_NAME_KEY)));
```
mertemin: Please see comment above about collecting street names.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @author savannahostrowski
+ */
+
+public class AddressPointMatchCheck extends BaseCheck
+{
+    private static final long serialVersionUID = 1L;
+    public static final String NO_STREET_NAME_POINT_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. The street name should likely "
+            + "be one of {1}. These names were derived from nearby Points.";
```
mertemin: OSM doesn't have points. I'd suggest referring them as nodes.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    public static final String NO_STREET_NAME_EDGE_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. The street name should likely "
+            + "be one of {1}. These names were derived from nearby Edges.";
```
mertemin: OSM doesn't have edges. I'd suggest referring them as ways.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    public static final String NO_SUGGESTED_NAMES_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. No suggestions names were found.";
```
mertemin: I'd suggest mentioning the tag key that is missing the value here.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            NO_STREET_NAME_POINT_INSTRUCTIONS, NO_STREET_NAME_EDGE_INSTRUCTIONS,
+            NO_SUGGESTED_NAMES_INSTRUCTIONS);
+    private static final String ADDRESS_STREET_NUMBER_KEY = AddressHousenumberTag.KEY;
+    private static final String POINT_STREET_NAME_KEY = AddressStreetTag.KEY;
+    private static final String EDGE_STREET_NAME_KEY = NameTag.KEY;
+    private static final double BOUNDS_SIZE_DEFAULT = 150.0;
```
mertemin: This seems a bit high. This would create a rectangle with ~300m long sides, which would cover a relatively big area. I'd suggest running with lower values and checking the results. I'd bet with a value like 25 here you'd be able to generate the same results.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Point point = (Point) object;
+        // Get a bounding box around the Point of interest
+        final Rectangle box = point.getLocation().boxAround(boundsSize);
+
+        // Get all Points inside the bounding box
+        final Iterable<Point> interiorPoints = point.getAtlas().pointsWithin(box);
```
mertemin: Name `interiorPoints` is a bit misleading (geometries have interior points). I'd suggest renaming it to `nearbyPoints`.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+        // Convert the Iterable into a Stream for filtering, remove Points that have null as their
+        // street name as they cannot be candidates for a street for the Point of interest
+        final Set<Point> points = StreamSupport.stream(interiorPoints.spliterator(), false)
```
mertemin: You can say `Iterable.stream(point.getAtlas().pointsWithin(box))` to stream points within an area.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+                .filter(interiorPoint -> interiorPoint.getTag(POINT_STREET_NAME_KEY).isPresent())
+                .collect(Collectors.toSet());
+
+        // Get all Edges that intersect or are contained within the bounding box
+        final Iterable<Edge> interiorEdges = point.getAtlas().edgesIntersecting(box);
```
mertemin: Same concern here. I'd suggest renaming it to `nearbyEdges`.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+        // Convert the Iterable into a Stream for filtering, remove Edge that have null as their
+        // street name as they cannot be candidates for a street for the Point of interest
+        final Set<Edge> edges = StreamSupport.stream(interiorEdges.spliterator(), false)
```
mertemin: In a similar sense, `Iterables.stream` could be utilized here.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,136 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.tags.AddressHousenumberTag;
+import org.openstreetmap.atlas.tags.AddressStreetTag;
+import org.openstreetmap.atlas.tags.names.NameTag;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * Tests for {@link AddressPointMatchCheck}
```
mertemin: Needs an update.

src/test/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheckTestRule.java
```diff
@@ -0,0 +1,98 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Node;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Point;
+
+/**
+ * {@link AddressPointMatchCheckTest} data generator
+ *
+ * @author savannahostrowski
+ */
+
+public class AddressPointMatchCheckTestRule extends CoreTestRule
```
mertemin: I'd suggest writing tests with empty street name tag value for the initial point.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
savannahostrowski: @mertemin Is there a nice way to collect an Iterables object into a Set to retain unique street names? Also, is there a concise way of checking whether Iterables are empty? I know Guava has this functionality but in straight Java, I don't think there's a pretty way of checking/doing those things.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
isabelleh: In some cases where an address belongs to a campus/hospital/large building it can be quite far from any nearby named roads so I think 25 would be a low value. Something closer to 75 might improve performance but catch more of the large building cases.

kasims44: I did a sensitivity analysis on different values and found that 75m yields the best results as a default value for the bounding box size. Any lower and the number of street recommendations drops significantly while increasing the size beyond 75 yields less and less significant results the larger the box becomes.

savannahostrowski:
Updates to add Associated Street Relation functionality (as mentioned by @isabelleh , and address @mertemin's comments).

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+            + "be one of {1}. These names were derived from nearby ways.";
+    public static final String NO_SUGGESTED_NAMES_INSTRUCTIONS = "This node, {0,number,#}, has "
+            + "no street name specified in the address. No suggestions names were found as there were no"
+            + "nearby nodes or ways with street name key tags.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            NO_STREET_NAME_POINT_INSTRUCTIONS, NO_STREET_NAME_EDGE_INSTRUCTIONS,
+            NO_SUGGESTED_NAMES_INSTRUCTIONS);
+    private static final String ADDRESS_STREET_NUMBER_KEY = AddressHousenumberTag.KEY;
+    private static final String POINT_STREET_NAME_KEY = AddressStreetTag.KEY;
+    private static final String EDGE_STREET_NAME_KEY = NameTag.KEY;
```
mertemin: Please consider not redefining these static values. `*.KEY` format is already short and clean.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final String STREET_RELATION_ROLE = "street";
+    private static final String ASSOCIATED_STREET_RELATION = "associatedStreet";
+    private static final double BOUNDS_SIZE_DEFAULT = 150.0;
```
mertemin: There was a comment thread about this value and 75m was found to be the optimum value. This value is still 150m, which makes the size width/height of a bounding box ~300m.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        // Object is an instance of Point
+        return object instanceof Point
+                // And does not have an Associated Street Relation
+                && !hasAssociatedStreetRelation(object)
+                // And has a street number specified
+                && object.getTag(ADDRESS_STREET_NUMBER_KEY).isPresent()
+                // And if the street name key has a value of null or if the street name key is not
+                // present
+                && Strings.isEmpty(object.tag(POINT_STREET_NAME_KEY));
```
mertemin: Please consider dropping the previous condition (`object.getTag(ADDRESS_STREET_NUMBER_KEY).isPresent()`) and replacing `Strings.isEmpty(object.tag(POINT_STREET_NAME_KEY))` with `Strings.isNullOrEmpty(object.tag(POINT_STREET_NAME_KEY))`

`Strings.isNullOrEmpty(object.tag(POINT_STREET_NAME_KEY))` will return
- true, if tag does not exist
- true, if tag exists, but null
- true, if tag exists, but empty
- false, if tag exists and has non-empty value

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+        // Initialize the rolePresent flag to false as we have not yet determined whether it is
+        // Part of the AssociatedStreet relation
+        boolean rolePresent = false;
+        final Point point = (Point) object;
+
+        // Get all relations that the Point is associated with, filter to keep only relations
+        // Where type=associatedStreet
+        final Set<Relation> relations = point.relations().stream().filter(
+                relation -> relation.tag(RelationTypeTag.KEY).equals(ASSOCIATED_STREET_RELATION))
+                .collect(Collectors.toSet());
```
mertemin: Please consider replacing the logic of this method with something like:
```
point.relations().stream().filter(relation -> relation.tag(RelationTypeTag.KEY).equals(ASSOCIATED_STREET_RELATION))
.anyMatch(relation -> relation.members().stream().anyMatch(member -> member.getRole().equals(STREET_RELATION_ROLE) && member.getEntity().getType().equals(ItemType.EDGE)))
```

Such call would return `true` if point is a part of a relation where relation has an `Edge` member with `street` role.

Java's `Stream` has powerful methods like `anyMatch` that could be used to find if any item is matching the condition in a stream of items. Collecting items into sets here is not necessary.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // Get a bounding box around the Point of interest
+        final Rectangle box = point.getLocation().boxAround(boundsSize);
+
+        // Get all Points in the bounding box, remove Points that have null as their
+        // street name or do not have the street name key tag, and get a set of candidate street
+        // names
+        final Set<String> points = Iterables.stream(point.getAtlas().pointsWithin(box))
+                .filter(nearbyPoint -> nearbyPoint.getTag(POINT_STREET_NAME_KEY).isPresent())
+                .map(nearbyPoint -> nearbyPoint.tag(POINT_STREET_NAME_KEY)).collectToSet();
```
mertemin: `nearbyPoint`'s street name tag is accessed twice here if it exists. I'd suggest mapping street name tags first (via `tag` method) and then filtering out nulls later and finally collecting them into a set.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // Get all Edges intersecting the bounding box, remove Edges that have null as their
+        // street name or do not have the street name key tag, and get a set of candidate street
+        // names
+        final Set<String> edges = Iterables.stream(point.getAtlas().edgesIntersecting(box))
+                .filter(nearbyEdge -> nearbyEdge.getTag(EDGE_STREET_NAME_KEY).isPresent())
+                .map(nearbyEdge -> nearbyEdge.tag(EDGE_STREET_NAME_KEY)).collectToSet();
```
mertemin: Same concern here. Tag is looked up and accessed twice if it exists.

src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressPointMatchCheck.java
```diff
@@ -0,0 +1,166 @@
...
+ * but no specified street name (addr:street) and are not part of an associated street Relation. No
+ * specified street name refers to either having a null value for the street name key, or no street
+ * name key present at all.
+ *
+ * @author savannahostrowski
+ */
+
+public class AddressPointMatchCheck extends BaseCheck
+{
+    private static final long serialVersionUID = 1L;
```
mertemin: Please update with a generated UID.

savannahostrowski:
Refactoring to `hasAssociatedStreetRelation()`, and finding points and edges in bounding box to avoid a double call to get the tag on each object. Also updated documentation accordingly, and added a serial version UID.

PR #32, opened 2018-03-27
Duplicate Ways Check

savannahostrowski:
The Duplicate Ways check flags Ways in OpenStreetMap which are either partially or completely duplicated.

This pull request covers the logic, unit tests, documentation, and configuration for the Duplicate Ways check.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
@@ -1,0 +1,109 @@
+package org.openstreetmap.atlas.checks.validation.linear.edges;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.Segment;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.tags.AreaTag;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * This check looks for partially or completely duplicated Ways via Edges.
+ *
+ * @author savannahostrowski
+ */
+public class DuplicateWaysCheck extends BaseCheck
+{
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
+    public static final String DUPLICATE_EDGE_INSTRUCTIONS = "This way, {0,number,#}, "
+            + "has at least one duplicate segment. ";
+
+    public static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(DUPLICATE_EDGE_INSTRUCTIONS);
+    public static final int ZERO_LENGTH = 0;
+    public static final String AREA_KEY = AreaTag.KEY;
+
+    // A map of segments and list of Edge identifiers
+    private final Map<Segment, Set<Long>> globalSegments = new HashMap<>();
+
+    @Override
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    public DuplicateWaysCheck(final Configuration configuration)
+    {
+        super(configuration);
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge
+                // Check to see that the edge is car navigable
+                && HighwayTag.isCarNavigableHighway(object)
+                // The edge is not part of an area
+                && !object.getTags().containsKey(AREA_KEY)
+                // The edge has not already been seen
+                && !this.isFlagged(((Edge) object).getMasterEdgeIdentifier());
+    }
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+
+        // Get current edge object
+        final Edge edge = (Edge) object;
+
+        // Get the edge identifier
+        final long identifier = edge.getIdentifier();
+
+        // Get all Segments in Edge
+        final List<Segment> edgeSegments = edge.asPolyLine().segments();
+
+        // For each Segment in the Edge
+        for (final Segment segment : edgeSegments)
+        {
+            // Make sure that we aren't flagging duplicate nodes
+            if (!segment.length().isGreaterThan(Distance.meters(ZERO_LENGTH)))
+            {
+                continue;
+            }
+
+            // Check if the Segment is in globalSegments
+            if (globalSegments.containsKey(segment))
+            {
+                // Add identifier to the list of identifiers with that segment
+                globalSegments.get(segment).add(identifier);
+
+                if (!this.isFlagged(edge.getMasterEdgeIdentifier()))
+                {
+                    this.markAsFlagged(edge.getMasterEdgeIdentifier());
+                    return Optional.of(this.createFlag(edge,
+                            this.getLocalizedInstruction(0, edge.getOsmIdentifier())));
+                }
+            }
+            else
+            {
+                // If it doesn't already exist, then add the segment and list with one identifier
+                final Set<Long> identifiers = new HashSet<>();
+                identifiers.add(identifier);
+                globalSegments.put(segment, identifiers);
+            }
+        }
+        return Optional.empty();
+    }
+}
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
@@ -29,7 +29,7 @@ public class DuplicateWaysCheck extends BaseCheck
 
     private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
             .asList(DUPLICATE_EDGE_INSTRUCTIONS);
-    private static final Distance ZERO_LENGTH = Distance.ZERO;
+    private static final Distance ZERO_DISTANCE = Distance.ZERO;
 
     @Override
     protected List<String> getFallbackInstructions()
@@ -53,7 +53,7 @@ public class DuplicateWaysCheck extends BaseCheck
                 // Check to see that the edge is car navigable
                 && HighwayTag.isCarNavigableHighway(object)
                 // The edge is not part of an area
-                && !object.getTags().containsKey(AreaTag.KEY);
+                && !object.getTag(AreaTag.KEY).isPresent();
     }
 
     @Override
@@ -68,6 +68,7 @@ public class DuplicateWaysCheck extends BaseCheck
         final Iterable<Edge> edgesInBounds = edge.getAtlas().edgesIntersecting(bounds,
                 Edge::isMasterEdge);
 
+
         for (final Edge edgeInBounds : edgesInBounds)
         {
             // If the Edge found in the bounds has an area tag or if the Edge has a length of 0
@@ -75,7 +76,7 @@ public class DuplicateWaysCheck extends BaseCheck
             // Edge that the bounds was drawn with then continue because we don't want to
             // Flag area Edges, duplicate Nodes, already flagged Edges, or the Edge of interest
             if (edgeInBounds.getTag(AreaTag.KEY).isPresent()
-                    || edgeInBounds.asPolyLine().length().equals(ZERO_LENGTH)
+                    || edgeInBounds.asPolyLine().length().equals(ZERO_DISTANCE)
                     || this.isFlagged(edgeInBounds.getIdentifier())
                     || edge.getIdentifier() == edgeInBounds.getIdentifier())
             {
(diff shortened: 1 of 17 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,109 @@
...
+    public static final String DUPLICATE_EDGE_INSTRUCTIONS = "This way, {0,number,#}, "
+            + "has at least one duplicate segment. ";
+
+    public static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(DUPLICATE_EDGE_INSTRUCTIONS);
+    public static final int ZERO_LENGTH = 0;
+    public static final String AREA_KEY = AreaTag.KEY;
+
+    // A map of segments and list of Edge identifiers
+    private final Map<Segment, Set<Long>> globalSegments = new HashMap<>();
```
mertemin: Is some cases, this data structure has a potential to fill the memory, cause OOM exceptions. I'd suggest finding a less memory-intensive approach to find duplicate ways.

savannahostrowski:
I've refactored the logic for this check based on @mertemin's comments surrounding memory usage. Using a HashSet within the flag method scope(as opposed to the original global HashMap), I think this is more along the lines of what we're looking for. I've also added in another preliminary filtering conditional to ensure that we are only checking master Edges in the Duplicate Ways check.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+{
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
+    public static final String DUPLICATE_EDGE_INSTRUCTIONS = "This way, {0,number,#}, "
+            + "has at least one duplicate segment.";
+
+    public static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(DUPLICATE_EDGE_INSTRUCTIONS);
+    public static final int ZERO_LENGTH = 0;
+    public static final String AREA_KEY = AreaTag.KEY;
```
mertemin: These statics don't have to be exposed publicly from the class.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge
+                // Check to see that the Edge is a master Edge
+                && ((Edge) object).isMasterEdge()
+                // Check to see that the edge has not already been seen
+                && !this.isFlagged(((Edge) object).getMasterEdgeIdentifier())
```
mertemin: If this condition is being checked, then `((Edge) object).isMasterEdge()` must be true, as well as `object instanceof Edge`. So you don't have to cast `object` here. Just get it's identifer and check if it is flagged or not.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        super(configuration);
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge
+                // Check to see that the Edge is a master Edge
+                && ((Edge) object).isMasterEdge()
```
mertemin: Please consider using [`Edge.isMasterEdgeIdentifier` method](https://github.com/osmlab/atlas/blob/1dbcbe5818d018f7c919b51d8ea82902ec1a5fa7/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java#L30).

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
mertemin: Using `AreaTag.KEY` might be cleaner than using `AREA_KEY`.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+public class DuplicateWaysCheck extends BaseCheck
+{
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
+    public static final String DUPLICATE_EDGE_INSTRUCTIONS = "This way, {0,number,#}, "
+            + "has at least one duplicate segment.";
+
+    public static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(DUPLICATE_EDGE_INSTRUCTIONS);
+    public static final int ZERO_LENGTH = 0;
```
mertemin: Please reuse [`Distance.ZERO`](https://github.com/osmlab/atlas/blob/1dbcbe5818d018f7c919b51d8ea82902ec1a5fa7/src/main/java/org/openstreetmap/atlas/utilities/scalars/Distance.java#L24).

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+        // Out the non-master Edges as the bounds Edges are not guaranteed to be uni-directional
+        final Iterable<Edge> edgesInBounds = Iterables
+                .stream(edge.getAtlas().edgesIntersecting(bounds)).filter(Edge::isMasterEdge);
+
+        for (final Edge edgeInBounds : edgesInBounds)
+        {
+            // If the Edge found in the bounds has an area tag or if the Edge has a length of 0
+            // Or if the Edge has already been flagged before then continue because we don't want to
+            // Flag area Edges or duplicate Nodes
+            if (edgeInBounds.getTags().containsKey(AREA_KEY)
```
mertemin: `getTags` is fetching all the tags. Please rely on `getTag` method.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+                    || edgeInBounds.asPolyLine().length().equals(Distance.meters(ZERO_LENGTH))
+                    || this.isFlagged(edgeInBounds.getMasterEdgeIdentifier()))
```
mertemin: `edgeInBounds` is verified that it is master edge on line 73. No need to call `getMasterEdgeIdentifier`.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+            {
+                continue;
+            }
+
+            // If the Set of Edges does not contain the Edge found in the bounds
+            if (!allEdgeSegments.containsAll(edgeInBounds.asPolyLine().segments()))
```
mertemin: Please consider using [`PolyLine.intersects` method](https://github.com/osmlab/atlas/blob/c0965af0fd1563844fff505a2a80cc79b95f655e/src/main/java/org/openstreetmap/atlas/geography/PolyLine.java#L592) to check for an overlap. You can also utilize [`PolyLine.intersections` method](https://github.com/osmlab/atlas/blob/c0965af0fd1563844fff505a2a80cc79b95f655e/src/main/java/org/openstreetmap/atlas/geography/PolyLine.java#L570) to get overlapping locations.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+                }
+                // Add all Segments flattened to the Set of Segments
+                allEdgeSegments.addAll(edgeInBounds.asPolyLine().segments());
+            }
+            // A full duplicate Edge was found in the Set of allEdgePolyLines
+            else
+            {
+                this.markAsFlagged(edgeInBounds.getMasterEdgeIdentifier());
+                return Optional.of(this.createFlag(edgeInBounds,
+                        this.getLocalizedInstruction(0, edgeInBounds.getOsmIdentifier())));
```
mertemin: A full overlap has the same instruction. I'd suggest not worrying about the full overlap. Just flag whenever an overlap is found. Atlas has edges vs OSM has ways. Edges by definition are partial OSM ways, so full overlap or not, you'll almost always flag partial OSM ways anyways.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,115 @@
...
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Set<Segment> allEdgeSegments = new HashSet<>();
+        final Edge edge = (Edge) object;
+
+        final Rectangle bounds = edge.asPolyLine().bounds();
+        // Get Edges which are contained by or intersect the bounds, and then filter
+        // Out the non-master Edges as the bounds Edges are not guaranteed to be uni-directional
+        final Iterable<Edge> edgesInBounds = Iterables
+                .stream(edge.getAtlas().edgesIntersecting(bounds)).filter(Edge::isMasterEdge);
```
mertemin: You can possibly say `final Iterable<Edge> edgesInBounds = edge.getAtlas().edgesIntersecting(bounds, Edge::isMasterEdge)`. `edgesIntersecting` optionally accepts a `Predicate`.

savannahostrowski:
Additionally, for my own knowledge, was the concern here mostly just to do with the use of the HashMap to store Segments and associated Edge IDs as a global variable?  Or is the concern more specific to how Atlas Checks are being run (is it being run on a country-by-country/region-by-region basis or is the whole world's data being processed in one fell swoop? 😄 )

mertemin:
@savannahostrowski Atlas-checks is by design flexible to accept any given part of the world to execute (or the whole world for that matter). The concern with accumulating state and using it as you go validate more features is that results will be determined by the order of features processed and memory footprint of a such solution could be big (based on the area of interest). Atlas could be utilized and possibly overlapping edges could be retrieved without additional memory tracking.

Here is an example:
edge-x (A B C D)
edge-y (H I K L)
edge-z (A B C)
edge-t (C D)

If you process x, y, z and finally t, then you'd flag z and t for overlap. If you process, t, z, y and finally x, then x will be flagged only.

savannahostrowski:
Addressed @mertemin comments regarding reuse of constants, and fine tuning preliminary filtering. Opted to use [.overlapsShapeOf()](https://github.com/osmlab/atlas/blob/c0965af0fd1563844fff505a2a80cc79b95f655e/src/main/java/org/openstreetmap/atlas/geography/PolyLine.java#L830) to detect duplicate Ways. Let me know if this still isn't sufficient!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,99 @@
...
+public class DuplicateWaysCheck extends BaseCheck
+{
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
+    private static final String DUPLICATE_EDGE_INSTRUCTIONS = "This way, {0,number,#}, "
+            + "has at least one duplicate segment.";
+
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(DUPLICATE_EDGE_INSTRUCTIONS);
+    private static final Distance ZERO_LENGTH = Distance.ZERO;
```
mertemin: Consider using `ZERO_DISTANCE` to be consistent with the type.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return object instanceof Edge
+                // Check to see that the Edge is a master Edge
+                && Edge.isMasterEdgeIdentifier(object.getIdentifier())
+                // Check to see that the edge has not already been seen
+                && !this.isFlagged(object.getIdentifier())
+                // Check to see that the edge is car navigable
+                && HighwayTag.isCarNavigableHighway(object)
+                // The edge is not part of an area
+                && !object.getTags().containsKey(AreaTag.KEY);
```
mertemin: Consider using `object.getTag(AreaTag.KEY)` to be consistent (see line #77).

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/DuplicateWaysCheck.java
```diff
@@ -0,0 +1,99 @@
...
+            {
+                continue;
+            }
+
+            final PolyLine edgeInBoundsPoly = edgeInBounds.asPolyLine();
+
+            // Checks that either the edgeInBounds PolyLine overlaps the edgePoly or
+            // That the edgePoly overlaps the edgeInBoundsPoly
+            if (edgeInBoundsPoly.overlapsShapeOf(edgePoly)
+                    || edgePoly.overlapsShapeOf(edgeInBoundsPoly))
```
mertemin: `overlapsShapeOf` is comparing individual segments. If there is overlap, then that would be found with the first call. If there is no overlap, then it wouldn't be found in both overlap checks. I'd suggest dropping one of the overlap calls, as it seems redundant.

PR #36, opened 2018-04-12
Sink Island Check Enhancements

savannahostrowski:
The enhancements made to the Sink Island check filter out Ways from flagging candidates which contain an amenity tag with the value of either Parking, Parking Space, Motorcycle Parking, or Parking Entrance on the end node.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
@@ -155,19 +155,9 @@ public class SinkIslandCheck extends BaseCheck<Long>
         // process all those edges
         explored.forEach(marked -> this.markAsFlagged(marked.getIdentifier()));
 
-        // We want to filter out any Edges whose end Node has an amenity tag equal to one of the
-        // amenityValuesToExclude
-        final Set<AtlasObject> filteredExploredEdges = explored.stream()
-                .filter(edge -> validEdge(edge)).collect(Collectors.toSet());
-
-        if (!filteredExploredEdges.isEmpty())
-        {
-            // Create the flag if and only if the empty flag value is not set to false
-            return emptyFlag ? Optional.empty()
-                    : Optional
-                            .of(createFlag(filteredExploredEdges, this.getLocalizedInstruction(0)));
-        }
-        return Optional.empty();
+        // Create the flag if and only if the empty flag value is not set to false
+        return emptyFlag ? Optional.empty()
+                : Optional.of(createFlag(explored, this.getLocalizedInstruction(0)));
     }
 
     @Override


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
```diff
@@ -146,9 +154,19 @@ public boolean validCheckForObject(final AtlasObject object)
         // process all those edges
         explored.forEach(marked -> this.markAsFlagged(marked.getIdentifier()));
 
-        // Create the flag if and only if the empty flag value is not set to false
-        return emptyFlag ? Optional.empty()
-                : Optional.of(createFlag(explored, this.getLocalizedInstruction(0)));
+        // We want to filter out any Edges whose end Node has an amenity tag equal to one of the
+        // amenityValuesToExclude
+        final Set<AtlasObject> filteredExploredEdges = explored.stream()
+                .filter(edge -> !endNodeHasAmenityTypeToExclude(edge)).collect(Collectors.toSet());
```
mertemin: I'd suggest using `validEdge` and baking this logic in there. That method is used to filter out possible false positives.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
```diff
@@ -28,13 +30,19 @@
...
-            .asList("Road is impossible to get out of.");
+    private static final float LOAD_FACTOR = 0.8f;
+    private static final long TREE_SIZE_DEFAULT = 50;
+    private static final List<String> FALLBACK_INSTRUCTIONS = Collections
+            .singletonList("Road is impossible to get out of.");
+    private static final String MOTORCYCLE_PARKING_AMENITY = "MOTORCYCLE_PARKING";
+    private static final String PARKING_ENTRANCE_AMENITY = "PARKING_ENTRANCE";
+    private static final List<String> amenityValuesToExclude = Arrays.asList(
+            AmenityTag.PARKING.toString(), AmenityTag.PARKING_SPACE.toString(),
+            MOTORCYCLE_PARKING_AMENITY, PARKING_ENTRANCE_AMENITY);
```
mertemin: Please consider opening an issue on atlas project to add these tags to `AmenityTag`.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
```diff
@@ -180,4 +198,22 @@ private boolean validEdge(final AtlasObject object)
...
+     * @return {@code true} if the end node of the end has one of the amenityValuesToExclude, and
+     *         {@code false} otherwise
+     */
+    private boolean endNodeHasAmenityTypeToExclude(final AtlasObject object)
+    {
+        final Edge edge = (Edge) object;
+        final Optional<String> tagValue = edge.end().getTag(AmenityTag.KEY);
+
+        return tagValue.isPresent()
+                && amenityValuesToExclude.contains(tagValue.get().toUpperCase());
```
mertemin: If `AmenityTag` has the values we are looking for there, `Validators.isOfType` method could be utilized here.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
savannahostrowski: Okay, so I'll leave that as is for now, and open that issue.

mgcuthbert:
An update in PR #39 has resolved the build failure, please merge from upstream so that it will build correctly for your PR.

savannahostrowski:
@mgcuthbert Merged from upstream but it looks like my problem is due to the fact that I am relying on some new enums I added to Atlas. Is there something I need to do to get the most up-to-date code from Atlas?

MikeGost:
Update dependencies.gradle to use atlas 5.0.14 as the version, that should pull in the latest enum changes.

savannahostrowski:
@MikeGost THANKS!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
```diff
@@ -146,9 +155,19 @@ public boolean validCheckForObject(final AtlasObject object)
...
+                .filter(edge -> validEdge(edge)).collect(Collectors.toSet());
+
+        if (!filteredExploredEdges.isEmpty())
+        {
+            // Create the flag if and only if the empty flag value is not set to false
+            return emptyFlag ? Optional.empty()
+                    : Optional
+                            .of(createFlag(filteredExploredEdges, this.getLocalizedInstruction(0)));
+        }
+        return Optional.empty();
```
mertemin: I don't think changes between lines 158 and 170 are needed. Since `validEdge` method is updated, now edges whose end node has an amenity tag will be ignored.

See line 109, which collects `outEdge` with `validEdge` filter.

PR #37, opened 2018-04-13
Duplicate Point Check

savannahostrowski:
The Duplicate Point check flags Points in OSM that share the same location. 

This PR includes flag logic, documentation, configuration, and unit tests.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
@@ -1,0 +1,72 @@
+package org.openstreetmap.atlas.checks.validation.points;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.Location;
+import org.openstreetmap.atlas.geography.Rectangle;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+
+import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * This check looks for two or more {@link Point}s that are in the exact same location.
+ *
+ * @author savannahostrowski
+ */
+public class DuplicatePointCheck extends BaseCheck<Location>
+{
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList("Duplicate Node {0,number,#} at {1}");
+    private static final Distance ZERO_DISTANCE = Distance.ZERO;
+    private static final long serialVersionUID = 8624313405718452123L;
+
+    @Override
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    /**
+     * Default constructor
+     *
+     * @param configuration
+     *            the JSON configuration for this check
+     */
+    public DuplicatePointCheck(final Configuration configuration)
+    {
+        super(configuration);
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
```diff
@@ -0,0 +1,71 @@
...
+        return object instanceof Point;
+    }
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Point point = (Point) object;
+        if (!this.isFlagged(point.getLocation()))
+        {
+            final Rectangle box = point.getLocation().boxAround(ZERO_DISTANCE);
```
mertemin: There was a recent update to `DuplicateNodeCheck` to utilize Atlas API better. Please take a look at PR #34  and update this check accordingly. ([link to specific changes](https://github.com/osmlab/atlas-checks/pull/34#diff-152719b3cb1cd62b0b5bc262719a4acb))

savannahostrowski: Oh nice! Thanks @mertemin 😄

src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
```diff
@@ -0,0 +1,72 @@
...
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Point;
+    }
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Point point = (Point) object;
+        this.markAsFlagged(point.getLocation());
```
MikeGost: Would it make sense to flag only if we find duplicates? Otherwise, we're flagging at each location even if there is no duplicate there. If there's no duplicate there, we will never find another Point at that location. This would minimize memory impact. Maybe worth playing with.

src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * This check looks for two or more {@link Point}s that are in the exact same location.
+ *
+ * @author savannahostrowski
+ */
+public class DuplicatePointCheck extends BaseCheck<Location>
+{
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList("Nodes {0} are duplicates at {1}");
+    private static final Distance ZERO_DISTANCE = Distance.ZERO;
```
MikeGost: This can be removed :)

savannahostrowski: Nice catch! Thanks.

src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
savannahostrowski: Actually, I'd agree. Will do.

src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicatePointCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Point point = (Point) object;
+
+        final List<Point> duplicates = Iterables
+                .asList(object.getAtlas().pointsAt(point.getLocation()));
+        if (duplicates.size() > 1)
+        {
+            duplicates.forEach(duplicate -> this.markAsFlagged(point.getLocation()));
```
mertemin: Since points are duplicates, flagging just the initial point should be sufficient.

savannahostrowski: Haha, wooow. My bad. I can tell it's Friday.

PR #39, opened 2018-04-23
Update flagged point

jklamer:
Updating flagged point to include properties of the flagged location item.


src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -36,6 +51,30 @@ public FlaggedPoint(final Location point)
...
+            }
+            else if (locationItem instanceof Point)
+            {
+                tags.put(ITEM_TYPE_TAG, POINT_TAG);
+            }
+            return tags;
+        }
+        catch (final Exception exception)
+        {
+            return Collections.EMPTY_MAP;
```
MikeGost: This seems odd, what else could the `LocationItem` be besides a `Node` or `Point`? At minimum, I'd log if this ever happens, but throwing might be an option too.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -12,18 +15,30 @@
...
     private final Location point;
+    private final Map<String, String> properties;
 
     /**
      * Default constructor
      * 
-     * @param point
-     *            the {@code point} {@link Location} to flag
+     * @param locationItem
+     *            the {@code locationItem} {@link Location} to flag
```
MikeGost: Nit: `locationItem` --> `LocationItem`

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -12,18 +15,30 @@
  */
 public class FlaggedPoint extends FlaggedObject
 {
+    private static final String NODE_TAG = "Node";
```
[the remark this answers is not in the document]
jklamer: They are, oops.

build.gradle
```diff
@@ -40,7 +40,7 @@ repositories
     // For Spark CDH
     maven { url "https://repository.cloudera.com/content/repositories/releases/" }
     // For jetty (through spark)
-    maven { url "http://repo.spring.io/plugins-release/" }
+    maven { url "http://repository.cloudera.com/cloudera/cloudera-repos/" }
```
mgcuthbert: What necessitated this particular change?

jklamer: Change in the dependent location causing the build to fail.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java
```diff
@@ -15,6 +15,14 @@
 public abstract class FlaggedObject implements Serializable
 {
     protected static final String COUNTRY_MISSING = "NA";
+    protected static final String AREA_TAG = "Area";
+    protected static final String EDGE_TAG = "Edge";
+    protected static final String OSM_IDENTIFIER_TAG = "osmid";
```
mertemin: Key name is inconsistent with other key names.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -36,6 +51,27 @@ public FlaggedPoint(final Location point)
...
+
+    @SuppressWarnings("unchecked")
+    private Map<String, String> initProperties(final LocationItem locationItem)
+    {
+        final Map<String, String> tags = locationItem.getTags();
+        tags.put(ITEM_IDENTIFIER_TAG, locationItem.getIdentifier() + "");
+        tags.put(OSM_IDENTIFIER_TAG, locationItem.getOsmIdentifier() + "");
+        if (locationItem instanceof Node)
+        {
+            tags.put(ITEM_TYPE_TAG, NODE_TAG);
```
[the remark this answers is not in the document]
MikeGost: @mertemin - I don't think this is the case. `getTags()` return a copy of the map by default. `Taggable` returns a new Map of the tags and the PackedTagStore will return a new map with all the key/value pairs for a given index (Node, Point). I wrote a quick test to get an AtlasEntity, get its tags, update an existing tag value and create a new tag key/value pair. I then grabbed the tags from the AtlasEntity again and its tags were not updated.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java
```diff
@@ -15,6 +15,14 @@
...
+    protected static final String ITEM_IDENTIFIER_TAG = "ItemId";
+    protected static final String ITEM_TYPE_TAG = "ItemType";
+    protected static final String LINE_TAG = "Line";
+    protected static final String NODE_TAG = "Node";
+    protected static final String POINT_TAG = "Point";
```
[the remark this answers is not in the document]
MikeGost: +1 here

PR #40, opened 2018-04-26
One Member Relation Check

savannahostrowski:
The OneMemberRelation Check flags Relations that only have one Member.

This pull request contains logic, documentation, configuration, and unit tests.

OSM Examples:
https://www.openstreetmap.org/relation/6077987
https://www.openstreetmap.org/relation/7214599

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
@@ -41,7 +41,7 @@ public class OneMemberRelationCheck extends BaseCheck
     @Override
     public boolean validCheckForObject(final AtlasObject object)
     {
-        return object instanceof Relation && !this.isFlagged(object);
+        return object instanceof Relation;
     }
 
     @Override
@@ -55,8 +55,7 @@ public class OneMemberRelationCheck extends BaseCheck
         if (members.size() == 1 && (!relation.isMultiPolygon() || members.iterator().next()
                 .getRole().equalsIgnoreCase(RelationTypeTag.MULTIPOLYGON_ROLE_INNER)))
         {
-
-            this.markAsFlagged(relation);
+            
             return Optional.of(createFlag(
                     relation.members().stream().map(RelationMember::getEntity)
                             .collect(Collectors.toSet()),
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
@@ -1,6 +1,6 @@
 package org.openstreetmap.atlas.checks.validation.relations;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -11,7 +11,6 @@ import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
 import org.openstreetmap.atlas.geography.atlas.items.Relation;
 import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
 import org.openstreetmap.atlas.geography.atlas.items.RelationMemberList;
-import org.openstreetmap.atlas.tags.RelationTypeTag;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
 
 /**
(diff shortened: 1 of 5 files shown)


mgcuthbert:
An update in PR #39 has resolved the build failure, please merge from upstream so that it will build correctly for your PR.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public OneMemberRelationCheck(final Configuration configuration)
+    {
+        super(configuration);
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Relation && !this.isFlagged(object.getOsmIdentifier());
```
mertemin: Is there a need to flag relation OSM identifiers? As far as I know there is one-to-one mapping between OSM relations and Atlas relations.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
```diff
@@ -0,0 +1,83 @@
...
+        // If the relation is a multipolygon
+        if (relation.isMultiPolygon())
+        {
+            // Determine if there are members with the role:inner
+            final boolean isInner = relation.members().stream().anyMatch(member -> member.getRole()
+                    .equalsIgnoreCase(RelationTypeTag.MULTIPOLYGON_ROLE_INNER));
+
+            // If the only member of the relation has the role:inner then we want to flag the
+            // relation
+            if (isInner && relation.members().size() == 1)
```
mertemin: For every multi-polygon relation this method will be looking for a member with `role:inner`. However, that should be only needed when the number of members is 1. I'd suggest updating logic here to avoid doing redundant work. Also `.members()` call is returning a new `Set` every time it is being called. I'd suggest caching it.

A conditional like `members.size() == 1 && (!relation.isMultiPolygon() || members.iterator().next().getRole().equalsIgnoreCase(RelationTypeTag.MULTIPOLYGON_ROLE_INNER))` should be able to filter relations that needs to be flagged.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
savannahostrowski: No, you're right. I think that OSM relation identifiers and Atlas relation identifiers are one-to-one.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
savannahostrowski: Wow, love it. So slick. Thanks @mertemin.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
mertemin: You probably do not need to flag them then. Because you won't hit the same relation twice.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Relation relation = (Relation) object;
+        final RelationMemberList members = relation.members();
+
+        // If the number of members in the relation is 1, and the relation is either not a
+        // Multipolygon relation or the sole member is role:inner
+        if (members.size() == 1 && (!relation.isMultiPolygon() || members.iterator().next()
```
jklamer: So while this is slick, I would say that that even when the relation is a multi-polygon with one outer member it is worth flagging. While there is nothing about that is prohibited by the OSM spec, a relation is designated to defined relationship between multiple osm elements. Also, a multi-polygon implies multiple polygons. I think the preferred data modeling of this case is just one osm way with all the tags of the relation and no relation.

To make this distinction vs other case it might be worth adding a second flag instruction to cover this.

savannahostrowski: @jklamer I agree, sounds good.

PR #42, opened 2018-04-30
Malformed Roundabout Check

savannahostrowski:
This check flags roundabouts where the directionality is opposite to what it should be, where
the roundabout is multi-directional, or where the roundabout has incorrect geometry (concave).

This PR includes flag logic, documentation, configuration, and unit tests.

**OSM Examples:**
Roundabout [id:242413354](https://www.openstreetmap.org/way/242413354) is multi-directional and has some segment going the wrong way. This is incorrect and should be flagged.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
@@ -193,8 +193,9 @@ public class MalformedRoundaboutCheck extends BaseCheck
             final Edge edge2 = roundaboutEdges.get((i + 1) % roundaboutEdges.size());
 
             // Get the cross product and then the direction of the roundabout
-            double crossProduct = getCrossProduct(edge1, edge2);
-            RoundaboutDirection direction = crossProduct < 0 ? RoundaboutDirection.COUNTERCLOCKWISE
+            final double crossProduct = getCrossProduct(edge1, edge2);
+            final RoundaboutDirection direction = crossProduct < 0
+                    ? RoundaboutDirection.COUNTERCLOCKWISE
                     : (crossProduct > 0) ? RoundaboutDirection.CLOCKWISE
                             : RoundaboutDirection.UNKNOWN;
 
@@ -231,7 +232,7 @@ public class MalformedRoundaboutCheck extends BaseCheck
      *            An Edge entity in the roundabout adjacent to edge1
      * @return A double corresponding to the cross product between two edges
      */
-    private static Double getCrossProduct(Edge edge1, Edge edge2)
+    private static Double getCrossProduct(final Edge edge1, final Edge edge2)
     {
 
         // Get the nodes' latitudes and longitudes to use in deriving the vectors
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
@@ -68,7 +68,7 @@ public class MalformedRoundaboutCheck extends BaseCheck
     {
         super(configuration);
         this.leftDrivingCountries = new HashSet<>(
-                Arrays.asList((String[]) configurationValue(configuration, "left.driving.countries",
+                Arrays.asList((String[]) configurationValue(configuration, "traffic.countries.left",
                         LEFT_DRIVING_COUNTRIES_DEFAULT)));
     }
 
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
@@ -28,20 +28,21 @@ import org.openstreetmap.atlas.utilities.configuration.Configuration;
 public class MalformedRoundaboutCheck extends BaseCheck
 {
     private static final long serialVersionUID = -3018101860747289836L;
-    public static final String WRONG_WAY_INSTRUCTIONS = "This roundabout, {0,number,#},is going the"
+    private static final String WRONG_WAY_INSTRUCTIONS = "This roundabout, {0,number,#},is going the"
             + " wrong direction, or has been improperly tagged as a roundabout.";
-    public static final String MULTIDIRECTIONAL_INSTRUCTIONS = "This roundabout, {0,number,#}, is"
+    private static final String MULTIDIRECTIONAL_INSTRUCTIONS = "This roundabout, {0,number,#}, is"
             + " multi-directional, or the roundabout has improper angle geometry.";
-    public static final Set<String> LEFT_DRIVING_COUNTRIES = new HashSet<>(
-            Arrays.asList("AIA", "ATG", "AUS", "BGD", "BHS", "BMU", "BRB", "BRN", "BTN", "BWA",
-                    "CCK", "COK", "CXR", "CYM", "CYP", "DMA", "FJI", "FLK", "GBR", "GGY", "GRD",
-                    "GUY", "HKG", "IDN", "IMN", "IND", "IRL", "JAM", "JEY", "JPN", "KEN", "KIR",
-                    "KNA", "LCA", "LKA", "LSO", "MAC", "MDV", "MLT", "MOZ", "MSR", "MUS", "MWI",
-                    "MYS", "NAM", "NFK", "NIU", "NPL", "NRU", "NZL", "PAK", "PCN", "PNG", "SGP",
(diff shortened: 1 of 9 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // We check that the object is an instance of Edge
+        return object instanceof Edge
+                // Make sure that the object has an iso_country_code
+                && object.getTag(ISOCountryTag.KEY).isPresent()
+                // Make sure that the edges are instances of roundabout
+                && JunctionTag.isRoundabout(object)
+                // Is not two-way
+                && !OneWayTag.isExplicitlyTwoWay(object)
```
mertemin: Just curious is it possible to have two way roundabouts? If not, it might worth flagging them as well.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+ * This check flags roundabouts where the directionality is opposite to what it should be, where the
+ * roundabout is multi-directional, or where the roundabout has incorrect geometry (concave).
+ *
+ * @author savannahostrowski
+ */
+
+public class MalformedRoundaboutCheck extends BaseCheck
+{
+    private static final long serialVersionUID = -3018101860747289836L;
+    private static final String WRONG_WAY_INSTRUCTIONS = "This roundabout, {0,number,#},is going the"
```
mertemin: nit: `{0,number,#},is` -> `{0,number,#}, is`

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+        final String isoCountryCode = edge.tag(ISOCountryTag.KEY).toUpperCase();
+
+        // Get all edges in the roundabout
+        final List<Edge> roundaboutEdges = getAllRoundaboutEdges(edge);
+
+        // Get the direction of the roundabout
+        final RoundaboutDirection direction = findRoundaboutDirection(roundaboutEdges);
+
+        // Determine if the roundabout is in a left or right driving country
+        final boolean isLeftDriving = leftDrivingCountries.contains(isoCountryCode);
```
mertemin: This could be moved after the first `if` block below. It is not used between lines 115-119.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+            final Edge currentEdge = queue.poll();
+
+            roundaboutEdges.add(currentEdge);
+
+            // Get the edges connected to the edge as an iterator
+            final Set<Edge> connectedEdges = currentEdge.connectedEdges();
+
+            for (final Edge connectedEdge : connectedEdges)
+            {
+                final Long edgeId = connectedEdge.getIdentifier();
```
mertemin: This is only needed if condition for the `if` block below evaluates to true. I'd suggest removing it and just calling `this.markAsFlagged(connectedEdge.getIdentifier())`.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+     *         products respectively, MULTIDIRECTIONAL if multiple directions are found in the same
+     *         roundabout, and UNKNOWN if all edge cross products are 0 or if the roundabout's
+     *         geometry is malformed
+     */
+    private static RoundaboutDirection findRoundaboutDirection(final List<Edge> roundaboutEdges)
+    {
+        // Initialize the directionSoFar to UNKNOWN as we have no directional information yet
+        RoundaboutDirection directionSoFar = RoundaboutDirection.UNKNOWN;
+
+        for (int i = 0; i < roundaboutEdges.size(); i++)
```
mertemin: Not a fan of this rule, but we are trying to avoid single-letter variables.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+     * @see "https://en.wikipedia.org/wiki/Cross_product"
+     * @param edge1
+     *            An Edge entity in the roundabout
+     * @param edge2
+     *            An Edge entity in the roundabout adjacent to edge1
+     * @return A double corresponding to the cross product between two edges
+     */
+    private static Double getCrossProduct(final Edge edge1, final Edge edge2)
+    {
+
```
mertemin: nit: Some method have this extra blank space in the beginning, some don't.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -0,0 +1,265 @@
...
+            // Get the edges connected to the edge as an iterator
+            final Set<Edge> connectedEdges = currentEdge.connectedEdges();
+
+            for (final Edge connectedEdge : connectedEdges)
+            {
+                final Long edgeId = connectedEdge.getIdentifier();
+
+                if (JunctionTag.isRoundabout(connectedEdge)
+                        && !roundaboutEdges.contains(connectedEdge))
+
```
mertemin: nit: extra line. A cleanup of extra lines would be nice.

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheckTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        this.verifier.verifyEmpty();
+    }
+
+    @Test
+    public void testClockwiseRoundaboutRightDrivingAtlas()
+    {
+        this.verifier.actual(this.setup.clockwiseRoundaboutRightDrivingAtlas(), check);
+
+        this.verifier.globallyVerify(flags -> Assert.assertEquals(1, flags.size()));
```
mertemin: I'd suggest checking more than just the count. Maybe the feature flagged.

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheckTest.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Tests for {@link MalformedRoundaboutCheck}
+ *
+ * @author savannahostrowski
+ */
+public class MalformedRoundaboutCheckTest
+{
+
+    private final MalformedRoundaboutCheck check = new MalformedRoundaboutCheck(
+            ConfigurationResolver.emptyConfiguration());
```
mertemin: This check is not stateless (processed features are flagged). Therefore, the order of processing could change the result. I'd suggest passing in a new instance of the check per unit test.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
savannahostrowski: Yeah, you're right. I don't think that it's valid to have a two-way roundabout. I'll remove that.

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheckTest.java
savannahostrowski: @mertemin Do you mean having each unit test verify the identifier of the edge that's been flagged? If that's the case, I'm not sure how that would improve the unit tests as it'll always be the first edge in the test Atlas, as that edge is always the starting edge for the roundabout?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
mertemin: Cool. The craziest roundabout I know is in England: https://www.openstreetmap.org/#map=19/51.56265/-1.77127

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheckTest.java
mertemin: Cool, I am OK with leaving as-is.

PR #43, opened 2018-05-10
Update README and gradlew

hallahan:
Hey!

I just got atlas checks running on my machine, and I have a small detail in the README that explains that the user does indeed need to put a PBF in the data dir. Also, I updated the gradlew, which pointed to a very old gradle. This way, users can call gradlew and be assured that it will be using the version of gradle we intend.

cc/ @savannahostrowski

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 **/out/
 /bin/
 **/ignored/
+/data/*.pbf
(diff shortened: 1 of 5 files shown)


mgcuthbert:
So this is true, as in the case in the README file does indeed require something in the data directory, whether it is a PBF or Atlas file. However you may want to include a link to the [Standalone Docs](docs/standalone.md) which includes information on how to essentially download pbf using http.

hallahan:
That link is already in there, I think?

mgcuthbert:
Ah yes, completely missed that. Then ignore my previous comment. :)

PR #44, opened 2018-05-14
Waterbody And Island Size Check

danielduhh:
This check identifies waterbodies that are either too small, or too large in size. Max/Min surface areas values are set in the configuration.

Islands are either Atlas Area's tagged `place=island`, `place=islet`, or `place=archipelago`, or inner members of an MultiPolygon water Relation (`type=multipolygon` and `natural=water`).

Waterbodies are Atlas areas tagged `natural=water` or outer members of an MultiPolygon water Relation.

cc: @medercc


src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+ * @author danielbaah
+ */
+public class WaterbodyAndIslandSizeCheck extends BaseCheck
+{
+
+    private static final long serialVersionUID = 4105386144665109331L;
+    private static String WATERBODY_MIN_INSTRUCTION = "Waterbody with OSM ID {0,number,#}, has an surface area of {1,number,#.##} meters squared which is smaller than the expected minimum of {2} meters squared.";
+    private static String WATERBODY_MAX_INSTRUCTION = "Waterbody with OSM ID {0,number,#}, has an surface area of {1,number,#.##} kilometers squared which is larger than the expected maximum of {2} kilometers squared.";
+    private static String ISLAND_MIN_INSTRUCTION = "Island with OSM ID {0,number,#}, has an surface area of {1,number,#.##} meters squared which is smaller than the expected minimum of {2} meters squared.";
+    private static String ISLAND_MAX_INSTRUCTION = "Island with OSM ID {0,number,#}, has an surface area of {1,number,#.##} kilometers squared which is larger than the expected maximum of {2} kilometers squared.";
```
mertemin: These min/max instructions could be combined into one for easier maintenance. Min and max could be put into instruction rather than having one instruction for min and another for max.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+                {
+                    return Optional.of(this.createFlag(object, this.getLocalizedInstruction(FOUR,
+                            object.getOsmIdentifier(), surfaceAreaMeters, this.isletMinimumArea)));
+                }
+                else if (surfaceAreaKilometers > this.isletMaximumArea)
+                {
+                    return Optional.of(this.createFlag(object,
+                            this.getLocalizedInstruction(FIVE, object.getOsmIdentifier(),
+                                    surfaceAreaKilometers, this.isletMaximumArea)));
+                }
```
mertemin: As mentioned before instructions for min/max could be combined, which would make this part shorter and cleaner.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+        // that fail the size check
+        else if (object instanceof Relation)
+        {
+            final Relation relation = (Relation) object;
+            // Get Relation members that are an Area, have either an inner or outer member,
+            // and have yet to be examined
+            final Set<RelationMember> relationMembers = relation.members().stream()
+                    .filter(member -> member.getEntity() instanceof Area)
+                    .filter(member -> !this.isFlagged(member.getEntity().getIdentifier()))
+                    .filter(this::isInnerOrOuterMember).collect(Collectors.toSet());
```
mertemin: `filter` calls could be combined into one. There is no need to call `filter` method with different conditions if you can combine conditions into one. (Although compiler will probably do the similar optimization here as well.)

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+
+    /**
+     * Helper function for filtering invalid Relation members. Each Relation member have either an
+     * inner or outer role, and the inner role must not contain the natural=rock tag.
+     *
+     * @param member
+     *            - RelationMember
+     * @return true if member is valid
+     */
+    private boolean isInnerOrOuterMember(final RelationMember member)
```
mertemin: Method is doing more than checking if given member is inner or outer. Member is checked if it is a rock or not. I'd suggest either updating method name or taking rock piece out of the method and moving it into `filter` call.

mertemin: On a second thought, conditions used for filtering on lines 209-211 could be combined in this method.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+            // Get Relation members that are an Area, have either an inner or outer member,
+            // and have yet to be examined
+            final Set<RelationMember> relationMembers = relation.members().stream()
+                    .filter(member -> member.getEntity() instanceof Area)
+                    .filter(member -> !this.isFlagged(member.getEntity().getIdentifier()))
+                    .filter(this::isInnerOrOuterMember).collect(Collectors.toSet());
+
+            if (!relationMembers.isEmpty())
+            {
+                final List<String> flagInstructions = new ArrayList<>();
```
mertemin: `CheckFlag` has a method `addInstruction` that could be used instead of collecting a list of instructions explicitly.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
```diff
@@ -0,0 +1,310 @@
...
+     *
+     * @param relationMembers
+     *            - Set of valid RelationMembers
+     * @param invalidRelationMembers
+     *            - New Set of RelationMembers that fail surface area test
+     * @param instructions
+     *            - Flag Instructions for RelationMembers that fail surface area test
+     */
+    private void getInvalidRelationMembers(final Set<RelationMember> relationMembers,
+            final Set<RelationMember> invalidRelationMembers, final List<String> instructions)
```
mertemin: This method has side effects. It is updating `invalidRelationMembers` and `instructions` that are passed in as parameters.

src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    private final WaterbodyAndIslandSizeCheck check = new WaterbodyAndIslandSizeCheck(
+            ConfigurationResolver.emptyConfiguration());
+
+    @Test
+    public void largeIsletTest()
+    {
+        this.verifier.actual(this.setup.getLargeIsletAtlas(), this.check);
+        this.verifier.verifyNotEmpty();
```
mertemin: I'd suggest verifying the flag or at least the feature flagged here.

src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeTest.java
```diff
@@ -0,0 +1,94 @@
...
+public class WaterbodyAndIslandSizeTest
+{
+    @Rule
+    public WaterbodyAndIslandSizeTestRule setup = new WaterbodyAndIslandSizeTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    private final WaterbodyAndIslandSizeCheck check = new WaterbodyAndIslandSizeCheck(
+            ConfigurationResolver.emptyConfiguration());
```
mertemin: This check is not stateless (because it is flagging identifiers). So the order of tests could change the result.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
danielduhh: Yeah, good idea

src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterbodyAndIslandSizeCheck.java
danielduhh: Nice! Didn't know that

PR #46, opened 2018-05-23
Invalid Access Tag Check

Bentleysb:
This check flags Edges and Lines where the `access` tag is invalid due to an improper use or lack of supporting tags.  

To be flagged all of the following must be true:

* A `highway` tag is present and above a minimum priority
* An `access` tag is present with a value of `no`
* There are no supporting tags for `access=no`, such as `public_transport=yes`
* The Edge or Line is not in an Area or Relation with a `landuse=military` or `military=*` tag

This PR includes flag logic, configuration, unit tests, and documentation.

**OSM Example:**
Way [id:28957025](https://www.openstreetmap.org/way/28957025) has the `access=no` tag without any supporting tags, such as `public_transport=yes`.


src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
```diff
@@ -0,0 +1,164 @@
...
+     * @param object
+     *            {@link LineItem} to check
+     * @return {@code true} if input {@link LineItem} is in an {@link Area} or {@link Relation} with
+     *         tag {@code landuse=MILITARY} or tag key {@code military}
+     */
+    private boolean isInMilitaryArea(final LineItem object)
+    {
+        for (final Area area : object.getAtlas()
+                .areas(area -> Validators.isOfType(area, LandUseTag.class, LandUseTag.MILITARY)
+                        || Validators.hasValuesFor(area, MilitaryTag.class)))
```
mertemin: This method is iterating over all the areas. Please use a method like `areasIntersecting` that would limit the search.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
```diff
@@ -0,0 +1,164 @@
...
+            if (object.intersects(areaPolygon)
+                    || areaPolygon.fullyGeometricallyEncloses(object.asPolyLine()))
+            {
+                return true;
+            }
+        }
+        for (final Relation relation : object.getAtlas().relations(
+                relation -> (Validators.isOfType(relation, LandUseTag.class, LandUseTag.MILITARY)
+                        || Validators.hasValuesFor(relation, MilitaryTag.class))
+                        && relation.isMultiPolygon()))
```
mertemin: Again this method is iterating over all the relations.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
```diff
@@ -0,0 +1,164 @@
...
+import org.openstreetmap.atlas.tags.LandUseTag;
+import org.openstreetmap.atlas.tags.MilitaryTag;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+
+/**
+ * This check flags {@link Edge}s and {@link Line}s that include an access tag with a value of no,
+ * and does not have any supporting tags. Supporting tags declare what is or is not included in
+ * {@code access=no}. For example a supporting tag of {@code public_transport=yes} would mean only
+ * public transport vehicles are allowed.
```
mertemin: Please document the use of `tags.filter` configuration. Otherwise, this check's implementation doesn't have a direct use of the tag example mentioned above. It'd be confusing.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheckTestRule.java
```diff
@@ -0,0 +1,372 @@
...
+            nodes = { @Node(coordinates = @Loc(value = TEST_1)),
+                    @Node(coordinates = @Loc(value = TEST_2)),
+                    @Node(coordinates = @Loc(value = TEST_3)),
+                    @Node(coordinates = @Loc(value = TEST_4)),
+                    @Node(coordinates = @Loc(value = TEST_5)),
+                    @Node(coordinates = @Loc(value = TEST_6)),
+                    @Node(coordinates = @Loc(value = TEST_7)),
+                    @Node(coordinates = @Loc(value = TEST_8)),
+                    @Node(coordinates = @Loc(value = TEST_9)),
+                    @Node(coordinates = @Loc(value = TEST_10)) },
```
mertemin: You don't need to create a node for each shape point in an edge. You only need nodes for edge's end points.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheckTestRule.java
```diff
@@ -0,0 +1,372 @@
...
+    private static final String TEST_5 = "0.0,0.4";
+    private static final String TEST_6 = "-0.1,0.5";
+    private static final String TEST_7 = "0.0,0.6";
+    private static final String TEST_8 = "0.0,0.7";
+    private static final String TEST_9 = "0.1,0.8";
+    private static final String TEST_10 = "0.0,0.9";
+    private static final String TEST_11 = "0.2,-0.1";
+    private static final String TEST_12 = "-0.2,-0.1";
+    private static final String TEST_13 = "-0.2,1.0";
+    private static final String TEST_14 = "0.2,1.0";
```
mertemin: Although it is OK to use these values, I'd suggest using locations that are closer to real life lat/long values.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
Bentleysb: The issue with using `areasIntersecting` is that it requires a polygon, and this check is working with `LineItem`s. As far as I am aware, `LineItem`s can be converted to polygons by either using their bounding box, or by connecting there first and last nodes. If this was done, false negatives could result if a road that would be flagged was near a military tagged area.    
`areasCovering` could be used, but would require iteration over all the `Location`s in the polyline of a `LineItem`. That is the best option I can come up with. I would love to hear if there is an even better alternative that you can think of.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
Bentleysb: This, unfortunately, does not seem to have as easy a fix as with areas. There does not appear to be an equivalent to `areasCovering` for relations. The only thing I could find was `relationsWithEntitiesIntersecting`. Using this would create false positive flags for anything that was completely contained in a relation, if I am not mistaken. I do not see an alternative to iterating over all relations, but would love to consider anything you might be able to come up with. I only need to test multipolygon relations for the intersection or containment of the current check object.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
mertemin: `LineItem` has a method called `bounds`, which will give you a `Polygon`. You can retrieve areas intersecting that and filter them down. e.g. `atlas.areasIntersecting(line.bounds(), area -> (line.intersects(areaPolygon) || areaPolygon.fullyGeometricallyEncloses(line.asPolyLine())) && (Validators.isOfType(area, LandUseTag.class, LandUseTag.MILITARY) || Validators.hasValuesFor(area, MilitaryTag.class)))`.

You should alsoe cache the result of `object.asPolyLine()`, because that method would be returning a new `PolyLine` object per call.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidAccessTagCheck.java
mertemin: See my other comment please. Retrieve relations intersecting line bounds and then filter them down.

Bentleysb: Thanks for the guidance. Your solution for areas is great. Unfortunately, it does not work in the case of relations. Using `relationsWithEntitiesIntersecting` will create false positive flags for items with bounds that are completely contained by the relation. 

Example:
https://www.openstreetmap.org/way/251110595 contained by
https://www.openstreetmap.org/relation/7437523

mertemin: Iterating over each and every relation should be avoided.

You can further filter relations returned from `relationsWithEntitiesIntersecting` method. That method accepts a second optional parameter: `Predicate<Relation>`. Something like:
```
object.getAtlas().relationsWithEntitiesIntersecting(object.bounds(), relation -> (Validators.isOfType(relation, LandUseTag.class, LandUseTag.MILITARY) || Validators.hasValuesFor(relation, MilitaryTag.class) && relation.isMultiPolygon()))
```
should give multi polygon relations with intersecting entities. Then `MultiPolygon` could be created from relation and checked for intersection.

Bentleysb: I agree that this should be avoided, but after all this I do not see an alternative. I thank you very much for all your suggestions, unfortunately, the use of `relationsWithEntitiesIntersecting` will create false positive flags. The LineItems in question do not intersect any of the members of the military relations that contain them, even when the bounds of the LineItems are used. Thus, no relations are passed to the predicate for further filtering.  

To completely substitute this loop, there would need to be a method that checks for containment of an item in a relation, not intersection. 

As another example:
Unit test `accessNoInHighwayEdgesLanduseMilitaryRelation` simulates this scenario.

mertemin: Sure, but that sounds like a problem in atlas - possibly missing an API that could check for intersection against the relation altogether. Feel free to create an issue on atlas project, document it here and then move forward. False positives are acceptable. However, iterating over all relations should be avoided given that the number of relations in an Atlas could be huge and such iteration will be repeated potentially for each Edge in the same Atlas.

mertemin: Alternatively, you can edit this check to go through military areas rather than `Edge`s and `Line`s. That way you can iterate military areas first and mark `Edge`s and `Line`s.

Bentleysb: Missing atlas relation method issue: [Relation Covering Method #137](https://github.com/osmlab/atlas/issues/137)

jklamer:
@Bentleysb sorry for the long merge delays, can you fix the conflicting files and we'll go ahead and merge

PR #48, opened 2018-06-04
Intersecting Building Check Enhancments

Bentleysb:
The enhancements made to the Intersecting Building check rework the way an overlap is defined. Previously an overlap was defined by having one area completely surrounding or matching another. Now overlap is defined as areas of the exact same dimensions. Intersections where one area surrounds another are now defined as 'contains'.
The configurable overlap.lower.limit was removed to accommodate this change. It is now a static 1.0, or complete area match.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheck.java
@@ -14,6 +14,7 @@ import org.openstreetmap.atlas.geography.clipping.Clip;
 import org.openstreetmap.atlas.geography.clipping.Clip.ClipType;
 import org.openstreetmap.atlas.tags.BuildingTag;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Surface;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -59,7 +60,7 @@ public class IntersectingBuildingsCheck extends BaseCheck<String>
     private static final String UNIQUE_IDENTIFIER_FORMAT = "%s,%s";
 
     // Minimum intersection to be contained
-    private static final double overlapLowerLimit = 1.0;
+    private static final double OVERLAP_LOWER_LIMIT = 1.0;
 
     // Overlap below this limit is not considered to be intersecting
     private final double intersectionLowerLimit;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheck.java
```diff
@@ -159,8 +152,28 @@ public boolean validCheckForObject(final AtlasObject object)
             // Flag based on intersection type
             if (resultType == IntersectionType.OVERLAP)
             {
-                flag.addObject(otherBuilding, this.getLocalizedInstruction(0,
-                        object.getOsmIdentifier(), otherBuilding.getOsmIdentifier()));
+                // Get object as a Surface
+                final Surface objectAsSurface = ((Area) object).asPolygon().surface();
+                // If object is larger than otherBuilding, the instruction states object contains
+                // otherBuilding
+                if (objectAsSurface.isLargerThan(otherBuilding.asPolygon().surface()))
```
mertemin: Every call to `otherBuilding.asPolygon().surface()` method will create a polygon and calculate surface area again. Please cache the value to avoid redundant calculations.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheck.java
```diff
@@ -240,7 +251,7 @@ private IntersectionType findIntersectionType(final Polygon polygon, final Polyg
         final long baselineArea = Math.min(polygon.surface().asDm7Squared(),
                 otherPolygon.surface().asDm7Squared());
         final double proportion = (double) intersectionArea / baselineArea;
-        if (proportion >= this.overlapLowerLimit)
+        if (proportion >= this.OVERLAP_LOWER_LIMIT)
```
mertemin: `this.OVERLAP_LOWER_LIMIT` -> `OVERLAP_LOWER_LIMIT`

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheckTest.java
```diff
@@ -103,17 +103,16 @@ public void testSmallIntersectionBuildingsAtlasWithIncreasedIntersectionLowerLim
     }
 
     @Test
-    public void testSmallIntersectionBuildingsAtlasWithSmallOverlapLowerLimit()
```
mertemin: Why is this test case being removed?

Bentleysb: This test is no longer part of what the check looks for. The check now looks for containment instead of overlap in the case this test presents. `testContainsBuildingAtlas()` replaces this test.

mertemin: That is fine, but I'd suggest then cleaning up `IntersectingBuildingsTestCaseRule` by removing `smallIntersectionBuildingsAtlas` method and the corresponding variable.

PR #49, opened 2018-06-04
Fixing issue that was causing exception in MapRouletteClient when Def…

mgcuthbert:
…aultPriority was not set in the configuration of a challenge.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteClient.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteClient.java
@@ -211,7 +211,7 @@ public class MapRouletteClient implements Serializable
             {
                 // retrieve the already created project and challenge
                 final Project project = this.projects.get(key.getFirst());
-                if (project == null)
+                if (project == null || project.getId() == -1)
                 {
                     logger.warn(
                             "Failed to upload batch to MapRoulette, Project {} has not been created.",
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteClient.java
```diff
@@ -221,7 +221,7 @@ private void upload(final Tuple<String, String> key)
                 {
                     final Challenge challenge = this.challenges.get(project.getId())
                             .get(key.getSecond());
-                    if (challenge == null)
+                    if (challenge == null || challenge.getId() == -1)
```
mertemin: I am guessing `Challenge.KEY_DEFAULT_PRIORITY_VALUE` could be used instead of `-1` here.

src/main/java/org/openstreetmap/atlas/checks/maproulette/serializer/ChallengeDeserializer.java
```diff
@@ -45,8 +45,8 @@ public Challenge deserialize(final JsonElement json, final Type typeOfT,
         ChallengePriority priority;
         try
         {
-            priority = ChallengePriority.valueOf(
-                    this.getStringValue(challengeObject, Challenge.KEY_DEFAULT_PRIORITY, ""));
+            priority = ChallengePriority.valueOf(this.getStringValue(challengeObject,
+                    Challenge.KEY_DEFAULT_PRIORITY, Challenge.KEY_DEFAULT_PRIORITY_VALUE + ""));
```
mertemin: I'd suggest using `String.valueOf` rather than concatenation with empty string. Check [this](https://stackoverflow.com/a/7752401/853005) out.

src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteClient.java
mgcuthbert: No this actually is not a value related to default priority. This is the returned ID from challenge creation. If during challenge creation the MapRoulette server returns an error or anything the ID of the challenge will be -1

PR #54, opened 2018-06-15
Fix integer division bug

jklamer:
Computing the batch size based on number of buckets has a bug where running more checks than the dividend will create an int division result of 0 which then causes multiple output files with only one geojson. Now the min batch size is 100, the integer division is simplified, and the max batch size remains constant.


src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessor.java
```diff
@@ -157,9 +157,9 @@ protected int computeBatchSize()
         }
         if (this.featureBuckets.isEmpty())
         {
-            return BATCH_DIVIDEND * BATCH_MULTIPLIER;
+            return MAX_BATCH_SUM;
         }
-        return BATCH_DIVIDEND / this.featureBuckets.size() * BATCH_MULTIPLIER;
+        return Math.max(MAX_BATCH_SUM / this.featureBuckets.size(), MIN_BATCH_SIZE);
```
mertemin: Previous if block can potentially be combined here for clarity. When I first looked, I thought `MAX_BATCH_SUM / this.featureBuckets.size()` could throw division by zero since `this.featureBuckets.size()` can return 0. But previous if block is taking take care of it.

PR #55, opened 2018-06-18
Invalid Lanes Tag Check

Bentleysb:
This check flags Edges where there is a `lanes` tag with an abnormal value. 

Abnormal values are defined as anything not in the configurable `lanes.filter`.  
The default values are set as `1,1.5,2,3,4,5,6,7,8,9,10`. 

Toll booth plazas are ignored, as they often have an abnormally high number of lanes. 

**OSM Examples:**
1. Way [id:156825478](https://www.openstreetmap.org/way/156825478) is flagged correctly by this check as a highway with no lanes.
2. Way [id:247311654](https://www.openstreetmap.org/way/247311654) is connected to a toll booth and has a high but valid number of lanes. It is not flagged.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -61,14 +61,14 @@
       "difficulty": "EASY"
     }
   },
-  "DuplicatePointCheck": {
-    "challenge": {
-      "description": "Tasks contain node locations where duplicate points of interest are found.",
-      "blurb": "Duplicate Points",
-      "instruction": "Open your favorite editor and remove one or more of the duplicate points of interest until only one remains.",
-      "difficulty": "EASY"
-    }
-  },
+   "DuplicatePointCheck": {
+     "challenge": {
+       "description": "Tasks contain node locations where duplicate points of interest are found.",
+       "blurb": "Duplicate Points",
+       "instruction": "Open your favorite editor and remove one or more of the duplicate points of interest until only one remains.",
+       "difficulty": "EASY"
+     }
+   },
   "DuplicateWaysCheck": {
     "challenge": {
       "description": "Tasks contain Ways which have been partially or completely duplicated.",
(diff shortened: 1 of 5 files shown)


config/configuration.json
```diff
@@ -107,6 +107,16 @@
       "tags":"highway"
     }
   },
+  "InvalidLanesTagCheck": {
+    "lanes.filter": "Lanes->1,1.5,2,3,4,5,6,7,8,9,10",
```
mgcuthbert: What does it means if a the lane tag has a value of 1.5?

mgcuthbert: Oh just found the answer here: https://wiki.openstreetmap.org/wiki/Key:lanes Which seems to indicate a narrow road. Interesting.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidLanesTagCheck.java
```diff
@@ -0,0 +1,168 @@
...
+        // Queue of edges to be processed
+        final ArrayDeque<Edge> toProcess = new ArrayDeque<>();
+        Edge polledEdge;
+
+        // Add original edge
+        connectedEdges.add((Edge) object);
+        toProcess.add((Edge) object);
+
+        // Get all connected edges with lanes tag values not in the lanesFilter
+        while (!toProcess.isEmpty())
```
mgcuthbert: You might want to also want to have a size limit on the connected edges that checked for here. The if statement below does limit the likelihood of this getting out of control, but under a very specific set of circumstances you could get into a situation where this loops for quite a while.

jklamer:
@Bentleysb sorry again, looks like there is merge conflicts in the configuration.json, probably because I merged something else. Can you resolve?

PR #57, opened 2018-06-20
Sign Post Check Enhancement

Bentleysb:
This enhancement changes default configurable values, removes the `motorway_junction` test from on ramps, reworks the `findFirstRampEdge` method, and enables this check.  

The new default values create less errors from overlapping values.

The `motorway_junction` test was removed from on ramps because they do not require a node with that tag: [OSM wiki: motorway_junction](https://wiki.openstreetmap.org/wiki/Tag:highway=motorway_junction)

`findFirstRampEdge` no longer uses heading to determine connection. It now uses a recursive search for arterial roads based on the configurable filters and minimum arterial classification. 

This check was disabled via the configuration. The configuration.json in this PR re-enables it.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
@@ -10,7 +10,6 @@ import org.apache.directory.api.util.Strings;
 import org.openstreetmap.atlas.checks.atlas.predicates.TypePredicates;
 import org.openstreetmap.atlas.checks.base.BaseCheck;
 import org.openstreetmap.atlas.checks.flag.CheckFlag;
-import org.openstreetmap.atlas.geography.Heading;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
 import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Node;
@@ -206,9 +205,7 @@ public class SignPostCheck extends BaseCheck<String>
 
         // Go back and collect edges
         final List<Edge> endEdges = new ArrayList<>();
-        // Check that we are not over the max and the edge haas a heading
-        final Optional<Heading> initialHeading = finalEdge.asPolyLine().initialHeading();
-        if (count < maxEdgeCount && initialHeading.isPresent())
+        if (count < maxEdgeCount)
         {
             // Collect in edges and make sure the list is not empty
             final Set<Edge> inEdges = finalEdge.inEdges();
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
@@ -201,17 +201,15 @@ public class SignPostCheck extends BaseCheck<String>
      */
     private List<Edge> findFirstRampEdge(final Edge finalEdge, final long maxEdgeCount)
     {
-        final int count = 0;
-
         // Go back and collect edges
         final List<Edge> endEdges = new ArrayList<>();
-        if (count < maxEdgeCount)
+        if (maxEdgeCount > 0)
         {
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
```diff
@@ -195,53 +199,48 @@ public boolean validCheckForObject(final AtlasObject object)
...
+                        endEdges.add(finalEdge);
+                        return endEdges;
+                    }
+                    // Recurse through inEdges if this edge is not the other side of a bidirectional
+                    // way
+                    else if (nextEdge.highwayTag().isIdenticalClassification(finalEdge.highwayTag())
+                            && finalEdge.getMasterEdgeIdentifier() != nextEdge
+                                    .getMasterEdgeIdentifier())
+                    {
+                        endEdges.addAll(findFirstRampEdge(nextEdge, maxEdgeCount - 1));
```
jklamer: So the idea here is that from the highway edge you walk backwards along all incoming link edges, getting a maximum of maxEdgeCount deep but allowing for multiple branches of that depth?

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheckTestRule.java
```diff
@@ -396,6 +396,93 @@
...
+            // nodes
+            nodes = { @Node(coordinates = @Loc(value = HIGHWAY_1)),
+                    @Node(coordinates = @Loc(value = HIGHWAY_3), tags = {
+                            "highway=motorway_junction" }),
+                    @Node(coordinates = @Loc(value = HIGHWAY_5)),
+                    @Node(coordinates = @Loc(value = LINK_1)),
+                    @Node(coordinates = @Loc(value = LINK_4)) },
+            // edges
+            edges = {
+                    @Edge(coordinates = { @Loc(value = HIGHWAY_1), @Loc(value = HIGHWAY_2),
```
jklamer: Not necessarily a comment about this line, but can you add a test that includes two way edges and make sure everything is as intended? Not sure of your exposure with the test atlas annotation, but it is pretty raw. To have a properly formatted reverse edge you need to create another ```@Edge``` with the locations reversed. And I also believe that it needs to have a flipped ID compared to the main edge. For example  the masterEdge ```@Edge( id = x ...``` and the reverse  ```@Edge( id = -x ...```.

Another thing to look into is 
```
@TestAtlas (loadFromJosmOsmResource = "relativePathFromSamePackagePathInTestResources")
private testAtlas;
```
This allows you to open JOSM and draw the problem in a layer and save that layer as an osm xml file to the Test resources package path, then use that data as a test atlas. It will handled edges/reverse edges provided you tag them correctly.

Bentleysb: I was attempting to test that in `motorwayMotorwayLinkTwoWayBranchMissingDestinationAtlas`, but did not quite have it all right. I have updated that test to use ids and have more 2 way edges.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
Bentleysb: That is correct. This was done to handle branching on ramps and have a hard escape for the recursive loop.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SignPostCheck.java
```diff
@@ -152,24 +162,18 @@ public boolean validCheckForObject(final AtlasObject object)
                 .forEach(inEdge ->
                 {
                     // Find the source of in-ramp
-                    final Edge rampEdge = findFirstRampEdge(inEdge);
+                    final List<Edge> rampEdgeList = findFirstRampEdge(inEdge, maxEdgeCountForRamp);
```
jklamer: nit: can you add ```this.maxEdgeCountForRamp```

PR #59, opened 2018-07-14
Build updates: Sonar, Jacoco, Gradle 4.8.1

matthieun:
Update to the build scripts and sonar/jacoco integration.


build.gradle
```diff
@@ -43,57 +36,6 @@ repositories
     maven { url "http://repository.cloudera.com/cloudera/cloudera-repos/" }
 }
 
-idea {
-    project {
-        languageLevel = '1.8'
-    }
-}
-
-spotless {
```
jklamer: Why are we getting rid of spotless? This is very useful

gradle/quality.gradle
```diff
@@ -0,0 +1,70 @@
...
+    reports.html.destination = file("${reporting.baseDir}/${name}")
+}
+
+configurations
+{
+    integrationTestCompile.extendsFrom testCompile
+    integrationTestRuntime.extendsFrom testRuntime
+}
+
+spotless {
```
jklamer: Oh I see it now. ''''gradle spolessApply''' from the workspace still be how to run this?

gradlew
```diff
@@ -1,25 +1,43 @@
...
+SAVED="`pwd`"
+cd "`dirname \"$PRG\"`/" >/dev/null
+APP_HOME="`pwd -P`"
+cd "$SAVED" >/dev/null
 
 APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
+# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+DEFAULT_JVM_OPTS=""
```
jklamer: What set up will we need to do with this? Will it work without editing this?

build.gradle
matthieun: I tried re-organizing the build to have a smaller default build file. Now all the configuration for spotless, checkstyle, jacoco, sonar and tests go into `gradle/quality.gradle` and is imported by the main build file.

gradle/quality.gradle
matthieun: Yes, the same commands work!

gradlew
matthieun: `gradlew` is an auto-generated script by the gradle command. So I would not touch it :)

build.gradle
matthieun: See [this diff](https://github.com/osmlab/atlas-checks/pull/59/files#diff-c197962302397baf3a4cc36463dce5eaR19)!

matthieun:
All the tests succeeded so in that case there does not seem to be any issue here. There was an issue for Atlas that immediately failed some tests, so it was very visible.

PR #60, opened 2018-07-19
Street Name Integers Only Check

Bentleysb:
This check flags edges that have only integers in their name. 

An example of an improper name tag in this case is something like: `name=42`  
Items like `name=1st Ave` are allowed, and not flagged.

Multiple variations of the name tag can be checked by adding their tag key to the configurable `name_keys.filter`. Default values are:
* name
* name:left
* name:right

**OSM Example:**
1. The way [id:395169730](https://www.openstreetmap.org/way/395169730) is a road and has the name 1543 (an integer).

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
@@ -44,7 +44,7 @@ public class StreetNameIntegersOnlyCheck extends BaseCheck
     public StreetNameIntegersOnlyCheck(final Configuration configuration)
     {
         super(configuration);
-        this.nameKeys = (List<String>) configurationValue(configuration, "name_keys.filter",
+        this.nameKeys = (List<String>) configurationValue(configuration, "name.keys.filter",
                 NAME_KEYS_DEFAULT);
     }
 
@@ -77,18 +77,21 @@ public class StreetNameIntegersOnlyCheck extends BaseCheck
         // Try to convert name to an integer. If it converts without failure it should be flagged.
         for (final String nameKey : this.nameKeys)
         {
-            try
+            final Optional<String> nameValue = object.getTag(nameKey);
+            if (nameValue.isPresent())
             {
-                Integer.parseInt(
-                        object.getOsmTags().getOrDefault(nameKey, "a").replaceAll(" ", ""));
+                try
+                {
+                    Integer.parseInt(nameValue.get().replaceAll(" ", ""));
+                }
+                catch (final NumberFormatException e)
+                {
+                    continue;
+                }
+                this.markAsFlagged(object.getOsmIdentifier());
+                return Optional.of(this.createFlag(object,
+                        this.getLocalizedInstruction(0, object.getOsmIdentifier())));
             }
-            catch (final NumberFormatException e)
-            {
-                continue;
-            }
-            this.markAsFlagged(object.getOsmIdentifier());
-            return Optional.of(this.createFlag(object,
-                    this.getLocalizedInstruction(0, object.getOsmIdentifier())));
         }
 
         return Optional.empty();
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
@@ -62,8 +62,8 @@ public class StreetNameIntegersOnlyCheck extends BaseCheck
     public boolean validCheckForObject(final AtlasObject object)
     {
         final Map<String, String> osmTags = object.getOsmTags();
-        return HighwayTag.isCarNavigableHighway(object) && object instanceof Edge
-                && ((Edge) object).isMasterEdge()
+        return object instanceof Edge && ((Edge) object).isMasterEdge()
+                && HighwayTag.isCarNavigableHighway(object)
                 && this.nameKeys.stream().anyMatch(osmTags::containsKey);
     }
 
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
@@ -58,18 +58,10 @@ public class StreetNameIntegersOnlyCheck extends BaseCheck
     @Override
     public boolean validCheckForObject(final AtlasObject object)
     {
-        if (HighwayTag.isCarNavigableHighway(object) && !this.isFlagged(object.getOsmIdentifier())
-                && object instanceof Edge)
-        {
-            for (final String nameKey : this.nameKeys)
-            {
-                if (object.getOsmTags().containsKey(nameKey))
-                {
-                    return true;
-                }
-            }
-        }
-        return false;
+        return HighwayTag.isCarNavigableHighway(object)
+                && !this.isFlagged(object.getOsmIdentifier()) && object instanceof Edge
+                && this.nameKeys.stream()
+                        .anyMatch(nameKey -> object.getOsmTags().containsKey(nameKey));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,105 @@
...
+     * This function will validate if the supplied atlas object is valid for the check.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return HighwayTag.isCarNavigableHighway(object)
```
jklamer: Any reason this is limited to car navigable highways?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,105 @@
...
+    {
+        // Try to convert name to an integer. If it converts without failure it should be flagged.
+        for (final String nameKey : this.nameKeys)
+        {
+            final Optional<String> nameValue = object.getTag(nameKey);
+            if (nameValue.isPresent())
+            {
+                try
+                {
+                    Integer.parseInt(nameValue.get().replaceAll(" ", ""));
```
jklamer: Can you test and make sure this will replace tab and other possible unicode whitespace characters? https://www.fileformat.info/info/unicode/category/Zs/list.htm 

If not can you update it to?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return HighwayTag.isCarNavigableHighway(object)
+                && !this.isFlagged(object.getOsmIdentifier()) && object instanceof Edge
```
jklamer: Because you're only flagging one object in the ```flag``` function you don't need to ```markAsFlag``` or check ```isFlagged```. This does a good job of stopping an edge and its reversed from both getting flagged but you can add ``` object instanceof Edge && ((Edge) object).isMasterEdge()``` here to make that clearer and avoid holding the OSM Identifiers. This will also mean the geometry being flagged is more likely to match exactly with the osm geometry.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheckTest.java
```diff
@@ -0,0 +1,54 @@
...
+public class StreetNameIntegersOnlyCheckTest
+{
+    @Rule
+    public StreetNameIntegersOnlyCheckTestRule setup = new StreetNameIntegersOnlyCheckTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    @Test
+    public void motorwayWithIntegerNameTagTest()
```
jklamer: can your write a quick test that disables or adds more tags in config?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
chrushr: We were only looking at street names, and the non-navigable highways are too trivial to be considered as streets. Plus, including them did not change the testing results.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return HighwayTag.isCarNavigableHighway(object) && object instanceof Edge
+                && ((Edge) object).isMasterEdge() && this.nameKeys.stream()
+                        .anyMatch(nameKey -> object.getOsmTags().containsKey(nameKey));
```
jklamer: ```getOsmTags``` creates a new Map at every call. Please make it once at the beginning.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,105 @@
...
+     * This function will validate if the supplied atlas object is valid for the check.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        final Map<String, String> osmTags = object.getOsmTags();
```
jklamer: 👍

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,105 @@
...
+    {
+        // Try to convert name to an integer. If it converts without failure it should be flagged.
+        for (final String nameKey : this.nameKeys)
+        {
+            final Optional<String> nameValue = object.getTag(nameKey);
+            if (nameValue.isPresent())
+            {
+                try
+                {
+                    Integer.parseInt(nameValue.get().replaceAll("\\h", ""));
```
jklamer: I think something you may want to look at is guava CharMatcher. They have a pre-made char matcher called ```BREAKING_WHITESPACE``` that has a ```replaceFrom``` method that might just do everything you need here.

Bentleysb: That is perfect! I had no idea this existed. Super cool.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheckTestRule.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Loc;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Node;
+
+/**
+ * Tests for {@link StreetNameIntegersOnlyCheck}
+ *
+ * @author bbreithaupt
+ */
+
+public class StreetNameIntegersOnlyCheckTestRule extends CoreTestRule
```
jklamer: Can you please add an example that should be flagged with some unsavory breaking whitespace characters separating numbers?

Bentleysb: Should have had that, but it looks like I typoed the change into `motorwayWithNonIntegerNameTagAtlas` instead of `motorwayWithIntegerNameTagAtlas`. That will be fixed in a sec.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameIntegersOnlyCheck.java
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        final Map<String, String> osmTags = object.getOsmTags();
+        return HighwayTag.isCarNavigableHighway(object) && object instanceof Edge
```
AlexHsieh22: Consider putting ```object instances Edge``` as the first condition followed by the isMasterEdge check.

PR #66, opened 2018-08-08
Mixed Case Name Check

Bentleysb:
This check flags objects with name tags that improperly use mixed cases.

Proper case use is defined by set standards and configurable exceptions. 

The standards are as follows:

* Words must start with a capital unless:
    * The first letter is preceded by a number
    * All the words in the name are lower case
* All other letters must be lower case unless: 
    * They follow an apostrophe and, they are not the last letter of the word
    * The entire word is uppercase, except the last letter if it follows or is followed by an apostrophe

The standards are broken by the following configurable exceptions (with default values):

* Articles that are capitalised only if they are the first word:
    * a, an, the
* Prepositions that do not need to start with a capital:
    * and, from, to, of, by, upon, on, off, at, as, into, like, near, onto, per, till, up, via, with, for, in
* Name affixes that may be followed by a capital:
    * Mc, Mac, Mck, Mhic, Mic
* Mixed case units of measurement that are valid after a number:
    * kV
    
These configurables allow this check to be adapted to test different languages.  
OSM uses the `name` tag for the name in a locations primary language, and `name:[ISOcode]` for other languages.
This check uses two configurable to control what languages are checked.

The first is a list of ISO codes for countries that should have there `name` tag checked. It has default values of:

* AIA, ATG, AUS, BHS, BRB, BLZ, BMU, BWA, VGB, CMR, CAN, CYM, DMA, FJI, GMB, GHA, GIB, GRD, GUY, IRL, JAM, KEN, LSO, MWI, MLT, MUS, MSR, NAM, NZL, NGA, PNG, SYC, SLE, SGP, SLB, ZAF, SWZ, TZA, TON, TTO, TCA, UGA, GBR, USA, VUT, ZMB, ZWE

The second is a list of `name:[ISOcode]` tags to check. Default values are:

* name:en

A final configurable is a list of characters that names are split by to for words. Its default values are: 

* SPACE, \-, /, &, @, –

**OSM Example:**
1. Way [id:4780932622](https://www.openstreetmap.org/node/4780932622) has the name `NZ Convenience store`. It is flagged because the S in store should be capitalized. 
2. Way [id:463905525](https://www.openstreetmap.org/way/463905525) has the name:en of `ATM drive through`. This is flagged because drive and through should be capitalized. ATM is fine because it is an all upper case word.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -141,7 +141,7 @@
     "lower_case":{
       "articles":["and", "to", "of"],
       "prepositions":["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as", "into",
-        "like", "near", "onto", "per", "till", "up", "via", "with", "for"]
+        "like", "near", "onto", "per", "till", "up", "via", "with", "for", "in"]
     },
     "words.split.characters":" -/()&@–",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -138,8 +138,11 @@
       "SGP", "SLB", "ZAF", "SWZ", "TZA", "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT",
       "ZMB", "ZWE"],
     "name.language.keys":["name:en"],
-    "lower_case_words":["and", "to", "of", "the"],
-    "characters.capital.prefix":"-/(&",
+    "lower_case":{
+      "prepositions":["and", "to", "of"],
+      "articles":["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as", "into", "like", "near", "onto", "per", "till", "up", "via", "with"]
+    },
+    "words.split.characters":" -/(&@–",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
     "challenge": {
       "description": "Tasks containing objects with mixed case names.",
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -145,6 +145,7 @@
     },
     "words.split.characters":" -/&@–",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
+    "units.mixed_case":["kV"],
     "challenge": {
       "description": "Tasks containing objects with mixed case names.",
       "blurb": "Mixed Case Name",
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -139,9 +139,10 @@
       "ZMB", "ZWE"],
     "name.language.keys":["name:en"],
     "lower_case":{
-      "articles":["and", "to", "of"],
-      "prepositions":["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as", "into",
-        "like", "near", "onto", "per", "till", "up", "via", "with", "for", "in"]
+      "articles":["a", "an", "the"],
+      "prepositions":["and", "from",
+        "to", "of", "by", "upon", "on", "off", "at", "as", "into", "like", "near", "onto",
+        "per", "till", "up", "via", "with", "for", "in"]
     },
     "words.split.characters":" -/&@–",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -132,7 +132,7 @@
     }
   },
   "MixedCaseNameCheck": {
-    "check_name.countries":["AIA", "ATG", "AUT", "BHS", "BRB", "BLZ", "BMU", "BWA", "VGB",
+    "check_name.countries":["AIA", "ATG", "AUS", "BHS", "BRB", "BLZ", "BMU", "BWA", "VGB",
       "CMR", "CAN", "CYM", "DMA", "FJI", "GMB", "GHA", "GIB", "GRD", "GUY", "IRL", "JAM",
       "KEN", "LSO", "MWI", "MLT", "MUS", "MSR", "NAM", "NZL", "NGA", "PNG", "SYC", "SLE",
       "SGP", "SLB", "ZAF", "SWZ", "TZA", "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT",
@@ -141,7 +141,7 @@
     "lower_case":{
       "prepositions":["and", "to", "of"],
       "articles":["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as", "into",
-        "like", "near", "onto", "per", "till", "up", "via", "with"]
+        "like", "near", "onto", "per", "till", "up", "via", "with", "for"]
     },
     "words.split.characters":" -/()&@–",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -138,7 +138,7 @@
       "SGP", "SLB", "ZAF", "SWZ", "TZA", "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT",
       "ZMB", "ZWE"],
     "name.language.keys":["name:en"],
-    "lower_case_words":["and", "to", "of", "de", "del", "los", "la", "y"],
+    "lower_case_words":["and", "to", "of", "the"],
     "characters.capital.prefix":"-/(&",
     "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
     "challenge": {
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -137,15 +137,15 @@
       "KEN", "LSO", "MWI", "MLT", "MUS", "MSR", "NAM", "NZL", "NGA", "PNG", "SYC", "SLE",
       "SGP", "SLB", "ZAF", "SWZ", "TZA", "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT",
       "ZMB", "ZWE"],
-    "name.language.keys":["name:en"],
-    "lower_case":{
-      "articles":["a", "an", "the"],
-      "prepositions":["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as", "into",
-        "like", "near", "onto", "per", "till", "up", "via", "with", "for", "in"]
+    "name":{
+      "language.keys":["name:en"],
+      "affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
+      "articles": ["a", "an", "the"],
+      "prepositions": ["and", "from", "to", "of", "by", "upon", "on", "off", "at", "as",
+        "into", "like", "near", "onto", "per", "till", "up", "via", "with", "for", "in"],
+      "units":["kv"]
     },
-    "words.split.characters":" -/&@–",
-    "name_affixes":["Mc", "Mac", "Mck","Mhic", "Mic"],
-    "units.mixed_case":["kV"],
+    "regex.split":" -/&@–",
     "challenge": {
       "description": "Tasks containing objects with mixed case names.",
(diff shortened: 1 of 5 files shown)


config/configuration.json
```diff
@@ -131,6 +131,28 @@
       "difficulty": "MEDIUM"
     }
   },
+  "MixedCaseNameCheck": {
+    "check_name.countries":["AIA", "ATG", "AUS", "BHS", "BRB", "BLZ", "BMU", "BWA", "VGB",
+      "CMR", "CAN", "CYM", "DMA", "FJI", "GMB", "GHA", "GIB", "GRD", "GUY", "IRL", "JAM",
+      "KEN", "LSO", "MWI", "MLT", "MUS", "MSR", "NAM", "NZL", "NGA", "PNG", "SYC", "SLE",
+      "SGP", "SLB", "ZAF", "SWZ", "TZA", "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT",
+      "ZMB", "ZWE"],
+    "name":{
```
mgcuthbert: I like the amount of configurability you added into this check. Very nice!

docs/checks/mixedCaseNameCheck.md
```diff
@@ -0,0 +1,75 @@
+# Mixed Case Name Check
+
+This check flags objects with name tags that improperly use mixed cases.
+
+Proper case use is defined by set standards and configurable exceptions. 
+
+The standards are as follows:
+
+* Words must start with a capital unless:
```
mgcuthbert: Adding examples of these use cases would be helpful.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+     * This function will validate if the supplied atlas object is valid for the check.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // Valid objects are items that were OSM nodes or ways
```
mgcuthbert: I would also add something at the end like "(Equivalent to Atlas nodes, points, edges, lines and areas)"

config/configuration.json
```diff
@@ -131,6 +131,28 @@
       "difficulty": "MEDIUM"
     }
   },
+  "MixedCaseNameCheck": {
+    "check_name.countries":["AIA", "ATG", "AUS", "BHS", "BRB", "BLZ", "BMU", "BWA", "VGB",
```
mgcuthbert: Please use the standardized, "countries.whitelist". Then the country list will be handled automatically by the framework and you don't have to add all the extra code.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+ * @author bbreithaupt
+ */
+public class MixedCaseNameCheck extends BaseCheck
+{
+
+    private static final long serialVersionUID = 7109483897229499466L;
+
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            "{0} {1,number,#} has (an) invalid mixed-case value(s) for the following tag(s): {2}.");
+    private static final List<String> CHECK_NAME_COUNTRIES_DEFAULT = Arrays.asList("AIA", "ATG",
```
mgcuthbert: This should be unnecessary, all country based filtering is handled automatically in the base check.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+            "to", "of", "by", "upon", "on", "off", "at", "as", "into", "like", "near", "onto",
+            "per", "till", "up", "via", "with", "for", "in");
+    private static final List<String> LOWER_CASE_ARTICLES_DEFAULT = Arrays.asList("a", "an", "the");
+    private static final String SPLIT_CHARACTERS_DEFAULT = " -/&@–";
+    private static final List<String> NAME_AFFIXES_DEFAULT = Arrays.asList("Mc", "Mac", "Mck",
+            "Mhic", "Mic");
+    private static final List<String> MIXED_CASE_UNITS_DEFAULT = Arrays.asList("kV");
+
+    // A list of countries where the name tag should be checked
+    private final List<String> checkNameCountries;
```
mgcuthbert: Don't need this variable, for the same reason mentioned above. BaseCheck will handle country filtering for you.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // Valid objects are items that were OSM nodes or ways
+        return !(object instanceof Relation) && !this.isFlagged(object.getOsmIdentifier())
+                && ((object.getTags().containsKey(ISOCountryTag.KEY)
+                        // Must have an ISO code that is in checkNameCountries...
+                        && this.checkNameCountries
```
mgcuthbert: This is unnecessary, country whitelists and blacklists are handled by the BaseCheck.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+     * @return an optional {@link CheckFlag} object that
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final List<String> mixedCaseNameTags = new ArrayList<>();
+        final Map<String, String> osmTags = object.getOsmTags();
+
+        // Check ISO against list of countries for testing name tag
+        if (this.checkNameCountries.contains(object.tag(ISOCountryTag.KEY).toUpperCase())
```
mgcuthbert: Again same comments as above. Not necessary, as it is all handled by the BaseCheck.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,263 @@
...
+     * Tests each word in a string for proper use of case in a name.
+     *
+     * @param value
+     *            String to check
+     * @return true when there is improper case in any of the words
+     */
+    private boolean isMixedCase(final String value)
+    {
+        // Check if it is all lower case
+        if (Pattern.compile("\\p{Lu}").matcher(value).find())
```
mgcuthbert: So with this, and again on line 181. Regular expressions have the potential of hurting performance quite badly. Have you done any performance testing to check what kind of impact regular expression matching could have on running this particular check?

mgcuthbert:
My biggest concern with this check is that you have multiple places where you use regular expression matching. A couple times in for loops. I think it would be really important to get some performance numbers, as I have a suspicion that for larger data sets this could slow down things quite dramatically.

Bentleysb:
I have not run any performance tests. I will work on getting some numbers and get back to you.

config/configuration.json
mgcuthbert: I think this needs to be more clearly defined in the documentation. Why wouldn't you check the name tag and name:[language] tags regardless of whether the ISOCountryTag exists or not.

Bentleysb: The idea is that you only want to check names that are in a language that the check is configured to handle. By default the check is configured for english. So, you only want to check the `name` tag when the country is english speaking, but `name:en` can be checked in any country because it will always be in english. I will see about clarifying that in the docs.

Bentleysb:
test this please

Bentleysb:
Performance tests have been run from both memory usage and runtime. 

The memory usage was measured overall and for java.util.regex specifically.  In varied testing of atlas size and number of flags generated, the total usage ranged from 1.8GB - 4.3GB and the regex usage ranged from 1.4MB - 192MB. There was no correlation between overall and regex memory usage. There was also no correlation with flags generated. There was correlation between the atlas size and regex memory usage. 

Runtime was benchmarked using overall times for the gradle run task as run on one check at a time. The atlas used was of Yangon, Myanmar as its flag count for Mixed Case name was quite high. Mixed Case Names flag count was 1719, whereas the mean for all checks was 253.
For all checks there was a range in time from 28.5-58.7 seconds with a mean of 31.6sec. Mixed Case Name had a runtime of 31.9sec, giving it a z-score of .05 standard deviations. There was an outlier in the set. After its removal, the set mean comes down to 30.6sec and Mixed Case Name's z-score becomes 0.65 standard deviations. 

Hopefully that gives a decent overview of the results. I am happy to share more of the data if you would like. I think both performance metrics are reasonable for the complexity of this check. As always, I am happy to discuss further.

config/configuration.json
Bentleysb: I tried to make this a little clearer in https://github.com/osmlab/atlas-checks/pull/66/commits/6d34f4dd832360d43ae113cb12bab1c0c2d9e518. Let me know if you still think it needs more clarification.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,277 @@
...
+            "TON", "TTO", "TCA", "UGA", "GBR", "USA", "VUT", "ZMB", "ZWE");
+    private static final List<String> LANGUAGE_NAME_TAGS_DEFAULT = Arrays.asList("name:en");
+    private static final List<String> LOWER_CASE_PREPOSITIONS_DEFAULT = Arrays.asList("and", "from",
+            "to", "of", "by", "upon", "on", "off", "at", "as", "into", "like", "near", "onto",
+            "per", "till", "up", "via", "with", "for", "in");
+    private static final List<String> LOWER_CASE_ARTICLES_DEFAULT = Arrays.asList("a", "an", "the");
+    private static final String SPLIT_CHARACTERS_DEFAULT = " -/&@–";
+    private static final List<String> NAME_AFFIXES_DEFAULT = Arrays.asList("Mc", "Mac", "Mck",
+            "Mhic", "Mic");
+    private static final List<String> MIXED_CASE_UNITS_DEFAULT = Arrays.asList("kV");
```
jklamer: Are there a lot of kV name tags?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
```diff
@@ -0,0 +1,277 @@
...
+            for (final String word : wordArray)
+            {
+                // Check if the word is intentionally mixed case
+                if (!isMixedCaseUnit(word))
+                {
+                    final Matcher firstLetterMatcher = this.anyLetterPattern.matcher(word);
+                    // If the word is not in the list of prepositions, and the
+                    // word is not both in the article list and not the first word: check that
+                    // the first letter is a capital
+                    if ((!this.lowerCasePrepositions.contains(word)
```
jklamer: don't be afraid to break this giant block up a bit for readability.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/MixedCaseNameCheck.java
Bentleysb: In some countries they made up 5-10% of the flags, from a random sample.

PR #68, opened 2018-08-28
One Member Relation Enhancement

Bentleysb:
This enhancement is to fix an issue where flags with empty geometries were being created. This was occurring when a relation with another relation as its only member was flagged. 

Now in these situations the geometry of the nested relations is recursively gathered, to show what the relation represents, and the situation is noted in the instruction.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
@@ -86,10 +86,12 @@ public class OneMemberRelationCheck extends BaseCheck
      */
     private Set<AtlasObject> getRelationMembers(final Relation relation)
     {
+        // Gather a set of the relation's members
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/relations/OneMemberRelationCheck.java
```diff
@@ -55,20 +59,54 @@ public boolean validCheckForObject(final AtlasObject object)
...
     }
+
+    /**
+     * Gathers all the members of a relation, replacing child relations with their members.
+     *
+     * @param relation
+     *            {@link Relation} to get the members of
+     * @return A {@link Set} of the relations members as {@link AtlasObject}s
+     */
+    private Set<AtlasObject> getRelationMembers(final Relation relation)
```
MikeGost: This looks like pretty valuable functionality - essentially it's flattening the relation tree structure. Can you add an atlas issue requesting this functionality? This is definitely a re-usable API.

PR #69, opened 2018-09-07
Invalid Mini Roundabout Check

nachtm:
This check flags mini-roundabouts that are likely to be poor representations of the ground truth. More specifically, it flags all Atlas objects which meet the following conditions: 
 - Is a `Node` object
 - Has the tag `highway=MINI_ROUNDABOUT`
 - Has less than `valence.minimum` number of car-navigable connecting edges AND
     - Has exactly 2 connecting edges, and those edges share the same OSM ID, OR
     - The node has neither `direction=CLOCKWISE` nor `direction=ANTICLOCKWISE` values

After performing sensitivity analysis, the best value for `valence.minimum` was determined to be 6, 
but this number is configurable.

**OSM Examples:**
The node [1092811451](https://www.openstreetmap.org/node/1092811451) is incorrectly flagged as a 
mini-roundabout and should probably be flagged as a `highway=TURNING_LOOP` or 
`highway=TURNING_CIRCLE` instead. 

The node [2367787523](https://www.openstreetmap.org/node/2367787523) has only 5 connecting car-navigable
edges. Satellite imagery confirms that this is not a mini-roundabout and should be reviewed by an editor.

**Dependencies**
This PR depends on the `DirectionTag` from osmlab/atlas#206 and should not be included in `atlas-checks` until `dependencies.gradle` has been updated to depend on a version of `atlas` that includes the `DirectionTag`.


MikeGost:
Should be safe to upgrade `dependencies.gradle` to rely on atlas 5.1.11, this will fix the build.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Flags any Nodes tagged as highway=MINI_ROUNDABOUT that do not have very many incoming/outgoing
+ * Edges as potentially tagged incorrectly.
+ *
+ * @author nachtm
+ */
+public class InvalidMiniRoundaboutCheck extends BaseCheck<Long>
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,149 @@
...
+    private static final long DEFAULT_VALENCE = 6;
+    private static final String MINIMUM_VALENCE_KEY = "valence.minimum";
+    private static final String OTHER_EDGES_INSTRUCTION = "This Mini-Roundabout Node ({0,number,#})"
```
MikeGost: Nit: It might make sense to make the instructions in terms of Ways instead of Edges. The average user won't understand how an OSM Way maps to a car-navigable edge.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,149 @@
...
+            + " has {1, number,#} connecting car-navigable Edges. Consider changing this.";
+    private static final String TWO_EDGES_INSTRUCTION = "This Mini-Roundabout Node ({0,number,#}) "
+            + "has 2 connecting car-navigable Edges. Consider changing this to highway=TURNING_LOOP or "
+            + "highway=TURNING_CIRCLE.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(TWO_EDGES_INSTRUCTION,
+            OTHER_EDGES_INSTRUCTION);
+    private static final DirectionTag[] VALID_DIRECTIONS = { DirectionTag.CLOCKWISE,
+            DirectionTag.ANTICLOCKWISE };
+    private final long valenceMinimum;
```
MikeGost: Nit: minimumValence?

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,149 @@
...
+        else if (!Validators.isOfType(node, DirectionTag.class, VALID_DIRECTIONS)
+                && valence < valenceMinimum && valence > 0)
+        {
+            result = Optional.of(this.flagNode(node, carNavigableEdges,
+                    this.getLocalizedInstruction(1, node.getOsmIdentifier(), valence)));
+        }
+        // Otherwise, this mini-roundabout is probably legitimate. Don't flag it.
+        else
+        {
+            result = Optional.empty();
```
MikeGost: Nit: You could potentially initialize this to empty to begin with and only update in case of flagging action.

src/test/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheckTestRule.java
```diff
@@ -0,0 +1,175 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas;
+
+/**
+ * Loads the data for the {@link InvalidMiniRoundaboutCheck} unit tests.
+ *
+ * @author nachtm
+ */
+public class InvalidMiniRoundaboutCheckTestRule extends CoreTestRule
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,143 @@
...
+ * Edges as potentially tagged incorrectly.
+ *
+ * @author nachtm
+ */
+public class InvalidMiniRoundaboutCheck extends BaseCheck<Long>
+{
+    private static final long DEFAULT_VALENCE = 6;
+    private static final String MINIMUM_VALENCE_KEY = "valence.minimum";
+    private static final String OTHER_EDGES_INSTRUCTION = "This Mini-Roundabout Node ({0,number,#})"
+            + " has {1, number,#} connecting car-navigable Ways. Consider changing this.";
```
jklamer: Can you include what they should consider changing it to?

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,143 @@
...
+        {
+            result = Optional.of(this.flagNode(node, carNavigableEdges,
+                    this.getLocalizedInstruction(0, node.getOsmIdentifier())));
+        }
+        // Otherwise, if there is not a direction tag, and the valence is less than valenceMinimum,
+        // we want to flag it.
+        else if (!Validators.isOfType(node, DirectionTag.class, VALID_DIRECTIONS)
+                && valence < minimumValence && valence > 0)
+        {
+            result = Optional.of(this.flagNode(node, carNavigableEdges,
```
jklamer: So I believe this will flag if one of three ways coming out of the round-about is a one way, but not if they are all 2 way?
Is this intended? If not, I would prefer filtering for master edges in ```this.getCarNavigableEdges(node);``` , and then the valence is about the topology and view of what you see. 

This would simplify  ```this.isTurnaround(carNavigableEdges)``` as well to ```return carNavigableEdges.size() == 1```

nachtm: Yes, you are correct and this is the intended behavior. Often, nodes with a total valence of 5 are not actually mini-roundabouts (see [OSM ID 2367787523](https://www.openstreetmap.org/node/2367787523)). I can certainly add a comment to clarify this somewhere if you think it's unclear.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
nachtm: I rephrased this one a little bit. Let me know what you think. The tricky thing is that using minimum valence is kind of a smell test for mini-roundabouts:  something is wrong, but there isn't enough information to diagnose exactly what the problem is, and there are many possibilities.

If you think it's still unclear, I can gather a list of potential suggestions and add them into this instruction.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
jklamer: no this is good. This was more confirming that it was as intended and thought about.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
jklamer: A quick list of alternatives would be great.

nachtm: How is this? I don't think this is an extensive list, but at least it's a place to start from for a user.

src/main/java/org/openstreetmap/atlas/checks/validation/points/InvalidMiniRoundaboutCheck.java
```diff
@@ -0,0 +1,156 @@
...
+        // If this Node is a turnaround, we always want to flag it.
+        if (this.isTurnaround(carNavigableEdges))
+        {
+            result = Optional.of(this.flagNode(node, carNavigableEdges,
+                    this.getLocalizedInstruction(0, node.getOsmIdentifier())));
+        }
+        // Otherwise, if there is not a direction tag, and the valence is less than valenceMinimum,
+        // we want to flag it.
+        else if (!Validators.isOfType(node, DirectionTag.class, VALID_DIRECTIONS)
+                && valence < minimumValence && valence > 0)
```
jklamer: nit: this.minimumValence

PR #72, opened 2018-09-17
Flag File Diff Commands

Bentleysb:
This adds command style tools for diffing geojson and log flag files. It also adds the `AtlasChecksCommand` class to run these tools from. This is built to replicate the structure of Atlas commands/sub-commands. 

The diffing commands take either a file or directory as input, and an optional output directory path. They will then report the number of unique flags, and the flags added, changed and subtracted from the source file(s) to the target file(s). If the output folder is specified the altered flags will be written to files in that directory. 

The added and subtracted numbers are calculated using the flag identifiers. If the flag with a specific id appears in either the source and not the target or visa versa, they are added to the appropriate output. The changes number is calculated using the atlas identifiers of the features in each flag. If a flag appears in both source and target it has its features' atlas ids cross referenced. If there is any mismatch the flag is added to the changes output. 

Example Command Line Output:
```
 Total Items: 7
   Additions: 1
     Changes: 2
Subtractions: 3
```


jklamer:
Cool functionality!

PR #74, opened 2018-10-10
Unresgister after complete

jklamer:
Unregister all processors on the EventBus after complete is called because no more events are allowed to get posted on the service after complete is called and the lingering processors allow for the possibility of memory leak. 

Testing:
Unsure how to test without exposing the processor container which I do not want to do for the design of the class. Passes all unit tests.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/event/EventService.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/event/EventService.java
@@ -1,5 +1,7 @@
 package org.openstreetmap.atlas.checks.event;
 
+import java.util.Collection;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/checks/event/EventService.java
```diff
@@ -25,6 +27,9 @@
     // Event bus to dispatch events
     private final EventBus eventBus;
 
+    // Container of processors on the EventService
+    private final Collection<Processor<?>> processors = new HashSet<>();
```
MikeGost: Collection --> Map?

Issue #87, opened 2018-10-24
CheckFlagEvent HighwayTag Bug


Bentleysb:
`CheckFlagEvent.featureDecorator()` throughs an `IllegalArgumentException` when flag objects have a highway tag value that is not in Atlas's HighwayTag Enum set. Normally this would not occur because highway tags are filtered out in atlas creation by the edge.json filter. However, when non-edges are flagged and have highway tags this can occur.

PR #88, opened 2018-10-24
CheckFlagEvent HighwayTag Bug

Bentleysb:
### Description:

This PR is to fix https://github.com/osmlab/atlas-checks/issues/87. A try catch statement is used to ignore bad highway tag values. 

### Potential Impact:

Stop the exception from stopping the program.

### Unit Test Approach:

A bad highway tag was added to the Area in the test atlas to test for this bug. 

### Test Results:

None

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessorTestRule.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessorTestRule.java
@@ -13,7 +13,7 @@ import org.openstreetmap.atlas.utilities.testing.TestAtlas.Node;
 /**
  * {@link CheckFlagGeoJsonProcessorTest} test data
  *
- * @author brian_l_davis
+ * @author brian_l_davis, bbreithaupt
  */
 public class CheckFlagGeoJsonProcessorTestRule extends CoreTestRule
 {
@@ -44,7 +44,8 @@ public class CheckFlagGeoJsonProcessorTestRule extends CoreTestRule
 
                     @Area(coordinates = { @Loc(value = TEST_5), @Loc(value = TEST_2),
                             @Loc(value = TEST_4), @Loc(value = TEST_1),
-                            @Loc(value = TEST_6) }, tags = { "building=yes" }) })
+                            @Loc(value = TEST_6) }, tags = { "building=yes",
+                                    "highway=notanenumvalue" }) })
     private Atlas atlas;
 
     public CheckFlagEvent getCheckFlagEvent()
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessorTestRule.java
```diff
@@ -44,7 +44,8 @@
 
                     @Area(coordinates = { @Loc(value = TEST_5), @Loc(value = TEST_2),
                             @Loc(value = TEST_4), @Loc(value = TEST_1),
-                            @Loc(value = TEST_6) }, tags = { "building=yes" }) })
+                            @Loc(value = TEST_6) }, tags = { "building=yes",
+                                    "highway=notanenumvalue" }) })
```
jklamer: is the function above called on areas?

Bentleysb: It is called on the jsonified tags of all check flag objects. So yes, but not as an Area object.

jklamer:
@Bentleysb make sure the branch is up to date so it can be merged. May be trivial.

src/test/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessorTestRule.java
hallahan: This should catch the solved bug, but why is it in a Area (a polygon)? Shouldn't you add it to:

```
edges = {

                    @Edge(coordinates = { @Loc(value = TEST_1), @Loc(value = TEST_2),
                            @Loc(value = TEST_3) }, tags = { "highway=primary" }) },
```

Bentleysb: It doesn't really matter too much which one the bad tag is added to. However, I wanted to keep the test realistic. This issue was detected when running the `AreasWithHighwayTagCheck`, and it was because the tag was not in an edge that it threw the exception. Normally, something with a highway tag is only ingested into an atlas if it has a highway tag that is in the Atlas enum set. This is due to the way the edge filter works. When something does get through as a non-edge with a highway tag, its inclusion in the enum is not guaranteed.

hallahan: Ah, thank you for the clarification. Makes sense!

hallahan:
LGTM! @jklamer can you? I don't have authority on this repo.

PR #92, opened 2018-10-30
Update InvalidLanesTagCheck Filter

Bentleysb:
### Description:

This PR changes the 'L' in 'Lanes' to be lowercase for the `lanesFilter`. This was acceptable before the recent updates the TaggableFilter class in atlas, but no longer gets automatically converted to all lower case. Without this fix, this check is currently flagging every car navigable edge with a lanes tag. 

### Potential Impact:

The check will once again work as intended with default values. 

### Unit Test Approach:

Nothing changed. 

### Test Results:

Ran this on multiple countries before and after. Flag counts dropped back to normal after the fix.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -139,7 +139,7 @@
     }
   },
   "InvalidLanesTagCheck": {
-    "lanes.filter": "Lanes->1,1.5,2,3,4,5,6,7,8,9,10",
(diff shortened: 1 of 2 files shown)


hallahan:
Can we merge this?

hallahan:
Thanks!

PR #95, opened 2018-11-06
checkFilter flag for IntegritySparkJob 

nachtm:
### Description:

This PR adds the ability to pass in a `-checkFilter` flag into `IntegritySparkJob`. This lets us quickly run one or two checks. For example, now I can run 
```
./gradlew run -Pchecks.local.checkFilter=PoolSizeCheck
```
 to quickly get the output for just `PoolSizeCheck`, or 
```
./gradlew run -Pchecks.local.checkFilter=PoolSizeCheck,AddressPointMatchCheck
```
to get the output for multiple checks. 

The code changes are as follows:
* ~addition of an overloaded `loadChecksForCountry(final String country, final Predicate<Class> filter)` function to `CheckResourceLoader`.~
* ~addition of a `loadEnabledChecks(final Predicate<Class> additionalFilter)` function, which loads only the default-enabled checks that also pass `additionalFilter`.~
 * addition of a `CheckResourceLoader.checks.whitelist` and `CheckResourceLoader.checks.blacklist` property
 * added a new switch `CHECK_FILTER` which takes in a list of comma-delimited check names. When the checks are run for each country, we now only run a check if:
     * the check is enabled for that country, **and** 
         * its name is included in the `CHECK_FILTER` flag **or** 
         * the `CHECK_FILTER` flag is not present

### Potential Impact:

This only adds functionality, so nothing downstream should be impacted. 

### Unit Test Approach:

No unit tests exist for IntegrityCheckSparkJob. I did manually test a lot of different configurations for this argument, and added some unit tests to CheckResourceLoader to test the new methods.

### Test Results:

All permutations of this command behaved as expected, and all unit tests are passing.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
@@ -91,7 +91,7 @@ public class IntegrityCheckSparkJob extends SparkJob
     private static final Switch<Boolean> PBF_SAVE_INTERMEDIATE_ATLAS = new Switch<>("savePbfAtlas",
             "Saves intermediate atlas files created when processing OSM protobuf data.",
             Boolean::valueOf, Optionality.OPTIONAL, "false");
-    private static final Switch<Set<String>> CHECKS_TO_RUN = new Switch<>("checksToRun",
+    private static final Switch<Set<String>> CHECK_FILTER = new Switch<>("checkFilter",
             "Comma-separated list of checks to run", value -> StringList
                     .split(value, CommonConstants.COMMA).stream().collect(Collectors.toSet()),
             Optionality.OPTIONAL);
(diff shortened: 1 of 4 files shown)


jklamer:
I am not 100% sold on this usage pattern. 

+'s: 
- Super easy to use (big plus)

-'s:
- I would prefer all decisions about checks and where they run be handled through the configuration.

What do you think about adding a feature to the configuration such that there is a checks.whitelist / blacklist functionality in the same pattern that we do with checks and countries? Just on the configuration wide level? 

You could then create a ```StandardConfiguration``` with the command line white/black list data and use that to create a ```MergedConfiguration``` with what normally goes into the CheckResourceLoader, and the CheckResource loader uses that information whether it is in the configuration resource or override from the command line?

hallahan:
I like it!

nachtm:
Thanks for the reviews everyone! In particular, @jklamer's advice made this much cleaner in my opinion. 

Now, `CheckResourceLoader` has a `checks.blacklist` and `checks.whitelist` property, which take in arrays of string names. Checks whose names match a member of one of these properties will be blacklisted or whitelisted, respectively. A check must be enabled **and** in/not in the whitelist/blacklist (if it exists) in order to be run.

The `checkFilter` switch now creates a `StandardConfiguration` which gets merged with the regular configuration loaded from a file and populates the `CheckResourceLoader.checks.whitelist` property, which behaves as described above.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -217,8 +222,22 @@ public void start(final CommandMap commandMap)
...
+        @SuppressWarnings("unchecked")
+        final Optional<List<String>> checkFilter = (Optional<List<String>>) commandMap
+                .getOption(CHECK_FILTER);
+
+        final Configuration checksConfiguration = new MergedConfiguration(Stream
+                .concat(Stream.of(
+                        ConfigurationResolver.loadConfiguration(commandMap, CONFIGURATION_FILES,
+                                CONFIGURATION_JSON)),
+                        Stream.of(checkFilter
+                                .<Configuration> map(whitelist -> new StandardConfiguration(
```
jklamer: Whats with the casting here??

nachtm: Otherwise, `checkFilter.map(whitelist -> new StandardConfiguration(...))` returns an `Optional<StandardConfiguration>`. So then the `orElse` would need to take a `StandardConfiguration`, and I didn't see a concise way to create an empty `StandardConfiguration`. 

There might be a better way around this, but this was the cleanest way I saw to keep everything concise.

PR #97, opened 2018-11-13
Update atlas version

jklamer:
### Description:

Update to latest atlas and remove legacy osmpbf loader.

### Potential Impact:

Possible loader changes under the hood for those using AtlasDataSource. The result should be the same.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
@@ -18,7 +18,7 @@ import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.AtlasResourceLoader;
 import org.openstreetmap.atlas.geography.atlas.multi.MultiAtlas;
 import org.openstreetmap.atlas.geography.atlas.pbf.AtlasLoadingOption;
-import org.openstreetmap.atlas.geography.atlas.pbf.OsmPbfLoader;
+import org.openstreetmap.atlas.geography.atlas.raw.creation.RawAtlasGenerator;
 import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
```diff
@@ -193,6 +193,6 @@ private Atlas loadPbf(final Resource input, final String country)
         final CountryBoundaryMap map = CountryBoundaryMap
                 .fromBoundaryMap(Collections.singletonMap(country, this.polygon));
         final AtlasLoadingOption option = AtlasLoadingOption.createOptionWithAllEnabled(map);
-        return new OsmPbfLoader(input, this.polygon, option).read();
+        return new RawAtlasGenerator(input, option, this.polygon).build();
```
MikeGost: Sorry, chiming in late - is this intended to be country-sliced and way-sectioned? In that case, you may want to add the following code: 

```
atlas = new RawAtlasGenerator(input, option, this.polygon).build();
if (option.isCountrySlicing())
{
       atlas = new RawAtlasCountrySlicer(
       option.getCountryBoundaryMap().getLoadedCountries(),
       option.getCountryBoundaryMap()).slice(atlas);
}
if (option.isWaySectioning())
{
     atlas = new WaySectionProcessor(atlas, option).run();
}
```
Otherwise, you're only creating a raw atlas (which has points, lines, relations) and nothing else (not sliced, sectioned, doesn't have edges, areas or nodes). cc. @jklamer

MikeGost:
I have concerns about how the upgrade was done - please take a look at the inline comment.

src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
lucaspcram: Ah this is my bad. @jklamer had me review this and I said it looked OK. I should have thought more carefully about what was going on.

PR #98, opened 2018-11-14
Self Intersecting Poly Line Check Enhancement

sayas01:
### Description:

This PR adds a minimum highway type to existing SelfIntersectingPolyLineCheck. The enhancement adds a `minimum.highway.filter` with a default value of `highway=service` to the config so that roads with low priority are not flagged. Valid edges for the check will now be checked for `building tags` or highway tags with equal or higher priority than the config value. 

### Potential Impact:

NA

### Unit Test Approach:

Added unit tests to test the `minimum.highway.type` configurable.

### Test Results:

NA

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -256,7 +256,7 @@
     }
   },
   "SelfIntersectingPolylineCheck": {
-    "tags.filter":"highway->*&highway->!construction&highway->!footway&highway->!path|building->*",
+    "minimum.highway.type":"service",
     "challenge": {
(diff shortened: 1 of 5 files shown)


chrushr:
@jklamer Total flags reduced by approximately 34% in RUS (those eliminated were not invalid flags to begin with, but they could be ignored due to low priority)

chrushr:
@jklamer Around 30%. It depends though -- for countries with massive amount of rural roads the flags can be reduced by a lot, whereas for denser urban places the number of flags might not change a lot.

PR #99, opened 2018-11-14
Generator update

jklamer:
### Description:

Update the atlas generator version

Describe other (non-unit) test results here.


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.openstreetmap.atlas
-version=4.0.8-SNAPSHOT
+version=4.0.9-SNAPSHOT
```
MikeGost: I thought this was updated after release is done?

jklamer: 4.0.8 is out
was missing newest generator tho.

PR #101, opened 2018-11-17
Tippecanoe Vector Tile Support

hallahan:
### Description:

This PR provides the ability to output atlas-checks into line-delimited GeoJSON that can be processed by Mapbox's tippecanoe vector tile converter. We have also provided a wrapper CLI that will take that line delimited GeoJSON output and drive tippecanoe to do the conversion for you.

Before we merge this PR, we need to do an atlas release with the following PR merged:

https://github.com/osmlab/atlas/pull/275

Then we need to do a release and update the `dependencies.gradle` for atlas-checks accordingly.

See instructions for more details:

https://github.com/hallahan/atlas-checks/blob/tippecanoe/src/main/java/org/openstreetmap/atlas/checks/vectortiles/README.md

### Potential Impact:

This PR will not impact any other parts of atlas-checks. It just adds the ability to easily make vector tiles with atlas check data.

### Unit Test Approach:

https://github.com/osmlab/atlas-checks/blob/79741af237d3d7556ebcd3295cd839a73cdff06a/src/test/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessorTest.java

https://github.com/osmlab/atlas-checks/blob/79741af237d3d7556ebcd3295cd839a73cdff06a/src/test/java/org/openstreetmap/atlas/checks/flag/CheckFlagTest.java#L102

### Test Results:

I've run `IntegrityCheckSparkJob` with the tippecanoe styled line-delimited GeoJSON. Then, I have converted that output into vector tiles with `TippecanoeCheckConverter`.

Also, unit tests are passing.


build.gradle
```diff
@@ -72,7 +72,8 @@ configurations
 dependencies
 {
     compile packages.commons
-    compile packages.atlas
+    compile 'org.apache.maven:maven-artifact:3.5.4'
+    compile files('libs/atlas.jar')
```
hallahan: This is just temporary, we need to do a new atlas release for this PR to work.

hallahan: Also note that Travis is failing because of this.

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java
```diff
@@ -217,4 +219,35 @@ public String toString()
...
+    public String asLineDelimitedGeoJsonFeatures()
+    {
+        return asLineDelimitedGeoJsonFeatures((jsonObject) ->
+        {
+        });
+    }
+
+    public String asLineDelimitedGeoJsonFeatures(final Consumer<JsonObject> jsonMutator)
+    {
+        final JsonObject flagGeoJsonFeature = flag.asGeoJsonFeature();
```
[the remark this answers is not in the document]
lucaspcram: It's strange. If you look at `config/checkstyle/checkstyle.xml`, you can see it's configured to complain when missing `this` on a field.
```
.
.
.
<module name="RequireThis">
             <property name="checkMethods" value="false"/>
</module>
.
.
.
```
But for whatever reason it doesn't actually end up complaining if you leave off `this`.

To clarify the issue I introduced earlier: my save-action was incorrectly adding `this` to method calls as well, which you can see is explicitly disabled in the checkstyle config. I fixed all those extra `this` on the methods since `this` on methods is redundant anyway.

I think there's an argument to be made for using `this` on fields, but I don't really think it's worth aggravating over.

Just be extra careful when adding local variables that you aren't shadowing a member variable :)

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -65,7 +66,8 @@
     {
         FLAGS,
         GEOJSON,
-        METRICS
+        METRICS,
+        TIPPECANOE
```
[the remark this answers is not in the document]
hallahan: Would you prefer LDGEOJSON ? Note that this output is focused around also providing the tippecanoe object with minzoom and layer properties. I'm ok with either. Or any other suggestion?

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java
```diff
@@ -217,4 +219,35 @@ public String toString()
...
+        final JsonObject flagGeoJsonProperties = flagGeoJsonFeature.get("properties")
+                .getAsJsonObject();
+        flagGeoJsonProperties.addProperty("flag:check", getCheckName());
```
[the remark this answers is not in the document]
hallahan: If we did `getSimpleName`, that would not be getting the name of the check's class, it would be getting `CheckFlagEvent`. That actual name is passed around through various constructors in the code base, but ultimately it is set in BaseCheck.java:

```java
private final String name = this.getClass().getSimpleName();
```

So, that is the actual property, but it's passed around several times. That can be cleaned up, but it's unrelated to the objective of this PR.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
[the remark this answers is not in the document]
jklamer: I think you're right. The output is useful on its own but its tailored for tippacanoue and the alternatives are very wordy and not clear.

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+    @Subscribe
+    public void process(final ShutdownEvent event)
+    {
+        try
+        {
+            super.write();
+        }
+        catch (final Exception e)
+        {
+            logger.warn("CheckFlag file write is failed.", e);
```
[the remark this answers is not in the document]
hallahan: fixed

hallahan:
Ready for review! @jklamer @yiqingj

hallahan:
We want to merge this:

https://github.com/osmlab/atlas/pull/296

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java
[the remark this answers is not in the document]
jklamer: Got messed up. Thought this was "getClassName". This is good 👍

src/main/java/org/openstreetmap/atlas/checks/flag/FlagVisualizer.java
```diff
@@ -1,49 +0,0 @@
-package org.openstreetmap.atlas.checks.flag;
```
jklamer: Why delete?

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -428,4 +437,63 @@ public String toString()
...
+
+        // Get the first bounds.
+        if (iterator.hasNext())
+        {
+            bounds = iterator.next().bounds();
+        }
+        // If we don't have it, let's instead return null island.
+        else
+        {
+            return GeoJsonUtils.geometry(GeoJsonUtils.POINT, GeoJsonUtils.coordinate(0.0, 0.0));
```
jklamer: Are you worried about a ton of output here?

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -428,4 +437,63 @@ public String toString()
...
+        }
+
+        // Otherwise, let's get the rest of the bounds and expand the bounds we have.
+        while (iterator.hasNext())
+        {
+            final Rectangle nextBounds = iterator.next().bounds();
+            bounds = bounds.combine(nextBounds);
+        }
+
+        // We want the bbox to be at least ten meters wide and high. This is for straight lines and
```
jklamer: 👍

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java
```diff
@@ -62,6 +65,20 @@ public String getCountry()
...
+     * @param flagIdentifier
+     *            We always will want to know the id of the flag assocaited with this flag object.
+     * @return GeoJSON representation of the flagged object.
+     */
+    public abstract JsonObject asGeoJsonFeature(String flagIdentifier);
+
+    /**
+     * @return The bounds of the object.
+     */
+    public abstract Rectangle bounds();
```
jklamer: can you make Flagged object implement Located too?

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
```diff
@@ -24,40 +27,35 @@
...
-            else
-            {
-                this.polyLine = new PolyLine(((AtlasItem) object).getRawGeometry());
-            }
+            // An Area's geometry doesn't include the end (same as start) location. To close the
+            // area boundary, we need to add the end location manually.
+            final List<Location> geometry = Iterables
+                    .asList(((AtlasItem) atlasItem).getRawGeometry());
+            geometry.add(geometry.get(0));
+            this.polyLine = new PolyLine(geometry);
```
jklamer: ```this.polyLine= ((Area) atlasItem).asPolygon().closedLoop()```

src/main/java/org/openstreetmap/atlas/checks/flag/FlagVisualizer.java
hallahan: It isn't being used anywhere. I cleared this with @mgcuthbert , just minor housekeeping.

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
hallahan: I haven't seen this condition met any any outputs thus far. This is a "just in case" situation.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
hallahan: Good call, this wasn't my code, just taking it out of an if statement. That should be cleaner.

hallahan:
Ready for merge.

hallahan:
Ok, @yiqingj ready for merge!

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
hallahan: Oh, I see how I can do it. I do think we should have polygons be polygons, but I can satisfy your suggestion by having `.closedLoop()` feed into constructing a `PolyLine`.

hallahan: Ok, done. I added a note that it's unclear why we only have FlaggedPolyline, and no FlaggedPolygon, since this is strange...

PR #112, opened 2018-12-17
MapRoulette Upload Command

nachtm:
### Description:

This is a new command to facilitate batch uploading of .log files to MapRoulette. More specifically, running `MapRouletteUploadCommand` with the following parameters
```
-logfiles=/path/to/a/folder/of/log/files/
-config=/path/to/a/configuration/file.json
-maproulette=server:port:projectName:apiKey
```
will take that `logfile` folder and create a project titled `projectName` at the given instance of MapRoulette. Inside of that project, there will be one challenge per check present in the log files inside the `logfile` folder.

### Potential Impact:

None -- this is a new command. 

### Unit Test Approach:

Tried uploading log files from checks that I had previously run, including those without challenge parameters in `configuration.json`, and points showed up as markers on MapRoulette.

### Test Results:

Files were uploaded successfully. Checks without challenge parameters were uploaded with empty descriptions. 

This is related to, but not dependent on, #111.


src/main/java/org/openstreetmap/atlas/checks/commands/MapRouletteUploadCommand.java
```diff
@@ -0,0 +1,154 @@
...
+            if (FilenameUtils.getExtension(logFile.getName()).equals(LOG_EXTENSION))
+            {
+                try (BufferedReader reader = new BufferedReader(new FileReader(logFile.getPath())))
+                {
+                    reader.lines().forEach(line ->
+                    {
+                        final Task task = gson.fromJson(line, Task.class);
+                        try
+                        {
+                            this.addTask(this.getChallenge(task.getChallengeName(), instructions),
```
jklamer: this function doesn't upload anything to maproulette correct? And if not, are you worried about there performance of trying to upload too many tasks at once if there are a large numbers of files/lines

nachtm: I think my comment disappeared somehow so apologies if you're getting double-notified.

The actual upload happens a little further down. I could potentially do an upload after each file -- assuming that MR treats a new upload to the same challenge as an append rather than a replace. If MR handles new uploads by replacing, then I think we have to read in the entire folder before uploading, since results from a particular check could be strewn over multiple files. I can look into that soon and get back to you

nachtm: @jklamer What do you think about this? Just tested on a couple of files and I can confirm that the end result is correct.

Uploading after each file wipes local memory clean so loading a large number of files should be manageable, and I think each file is generally small enough in size that we won't need to worry about splitting there

PR #114, opened 2019-01-02
Update dependency

chrushr:
### Description:

Increment atlas version from 5.3.0 to 5.3.1

Updated a function name changed in atlas https://github.com/osmlab/atlas/pull/312

### Unit Test Approach:

local build was successful

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
@@ -69,7 +69,7 @@ public class FlaggedPoint extends FlaggedObject
         final JsonObject properties;
         if (locationItem != null)
         {
-            feature = locationItem.asGeoJsonFeature();
+            feature = locationItem.asGeoJsonGeometry();
             properties = feature.getAsJsonObject("properties");
         }
         else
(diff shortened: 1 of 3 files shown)


chrushr:
This is currently blocked due to the function name change in https://github.com/osmlab/atlas/pull/312

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -69,7 +69,7 @@ public JsonObject asGeoJsonFeature(final String flagIdentifier)
         final JsonObject properties;
         if (locationItem != null)
         {
-            feature = locationItem.asGeoJsonFeature();
+            feature = locationItem.asGeoJsonGeometry();
```
yiqingj: is this a breaking change?

danielduhh: Yeah, on the Atlas side (5.3.1)

yiqingj: ahh I hate breaking changes... @matthieun  what's the reason to rename this method?

yiqingj:
OK talked to Matthieu offline, the conclusion is: if we have to make breaking changes,  we should label the PR, and clearly mention breaking changes in release note, and jump minor version number.  Also @deprecated should be used if possible.

PR #115, opened 2019-01-09
Add documentation for country-specific overrides

nachtm:
### Description:

`CheckResourceLoader` has the ability to override parameters on a country-specific basis, but there is no documentation for this use case. This PR adds documentation in the configuration README so that users have a place to reference without having to dig through source code.

### Potential Impact:

Docs will be a little clearer.

### Unit Test Approach:

None -- no code changes.

### Test Results:

N/A

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,6 +99,35 @@ You can also check for multiple values in a tag by separating the values by a co
 
 `highway->primary,secondary,tertiary`
 
+#### override
+
+This key allows you to specify certain values for your configuration for particular countries. Each check is allowed to have an `override` object.
+The properties of this object must have keys that are equal to ISO3 country codes.
+The values inside this object are themselves objects, which mimic the structure and properties of the default configuration for this check.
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -99,6 +115,46 @@ You can also check for multiple values in a tag by separating the values by a co
...
+      "one": {
+        "value": 1.5,
+        "word": "orange"
+      }
+    }
+  }
+}
+```
+For the fictional country with ISO code `JKL`, `CheckOne` will be run with `value` equal to 2.0 and `word` equal to banana, while in `ABC`, `DEF`, and `GHI`, `value` will be 1.5 and `word` will be orange.
+In the rest of the world, the default values of 1.0 and apple will be used.
```
jklamer: Also, I believe something to note: If JKL were to be a part of group one. It's specific override would take precedent over the group override.

jklamer:
Once again, great stuff.

PR #116, opened 2019-01-10
Relations Support for CheckFlag

sayas01:
### Description:

Currently, Atlas Objects that are of type Relation, are not handled by the [CheckFlag](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java) class in Atlas-Checks. CheckFlag considers only [AtlasItems](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasItem.java). The current approach to flag relation is by flagging all members of the relation and not the relation itself. As a result, it is difficult to visualize the Relation and no information about the Relation (Relation ID, Relation OSM tags etc) other than the member information is obtained by flagging its members. This project creates a new standard for modeling Relations in geojson so that they can be properly visualized.

This PR involves:
1. Updates to classes CheckFlagEvent and CheckFlag.
2. New class FlaggedRelation added that extends FlaggedObjects.
3. Updates to [FlaggedObject](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java) class to use the same property names in geojson as in [GeoJsonUtil](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java) class in Atlas to have a common standard between the two.

The feature property and geometry for FlaggedRelation are being populated by leveraging the 'asGeoJsonFeature' method in the Relation class. For Multipolygon relations, the geometry is populated from its inner and outer rings and for other types of relations, the geometry is  bounding box as a Polygon geometry.(Refer [asGeoJsonFeature)](https://github.com/osmlab/atlas/blob/bd28979affbf3f7051b0f2d77dd3a33fd73eef07/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java#L89)

### Potential Impact:

Ability to add Relations directly to CheckFlag instead of the current approach of just adding its members

### Unit Test Approach:

Unit tests have been added to test the enhancement.

### Test Results:

All tests passing


sayas01:
Closing to re run travis

sayas01:
@yiqingj It is a valid concern! Incase of large `boundary` type relations, the geometry is populated by the [`asGeojsonGeometry`](https://github.com/osmlab/atlas/blob/06c75037680f2e17cb3c8b6ae58679d0d6f57a08/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java#L87) method in Relation class by considering only the bounds as you mentioned. If we also want to update the geometry of very large multipolygon relations like the boundary relations, we should decide on what kinds of multipolygons(in terms of flattened member count) we need to handle this way; that is, how big the multipolygon reations have to be to treat it like the boundary relations. I would be happy to have this conversation.

sayas01:
@jklamer This is ready for review.

PR #119, opened 2019-01-25
ShadowDetectionCheck Enhancement: use altitude extractor

Bentleysb:
### Description:

This updates ShadowDetectionCheck to use the AltitudeExtractor via the MinHeightTag and HeightTag classes. This allows it to process values from these tags that include units (e.g. `3 m`, `6'3"`). This fixes the related false positive cases mentioned in https://github.com/osmlab/atlas-checks/pull/103.

### Potential Impact:

None

### Unit Test Approach:

Added unit test for tags with meters and feet/inches.

### Test Results:

Ran before and after on SGP and DNK. SGP had no changes and DNK had a loss of 8 cases. This is consistent with what was expected from testing done for https://github.com/osmlab/atlas-checks/pull/103. All lost flags were items that were flagged due to tags with units.


src/main/java/org/openstreetmap/atlas/checks/validation/areas/ShadowDetectionCheck.java
```diff
@@ -250,60 +251,33 @@ private boolean neighborsHeightContains(final AtlasObject part, final AtlasObjec
...
-            {
-                partMaxHeight = Double.parseDouble(partTags.get(BuildingLevelsTag.KEY))
-                        * LEVEL_TO_METERS_CONVERSION;
-            }
-            else
-            {
-                // Default to 0 height above the minimum
-                partMaxHeight = partMinHeight;
-            }
+            final double partMaxHeight = HeightTag.get(part).map(Altitude::asMeters)
```
jklamer: not specifically this line but does this check if the part max height is below the min height?

Bentleysb: It does. At the end of this method there is an attempt to create a range from the 2 height values. If the min is above the max an error is thrown, and the method returns false. This would cause the broken part to appear floating to the check, and it would be flagged.

jklamer: dope

PR #122, opened 2019-02-12
Add createFlag with multiple instructions

nachtm:
### Description:

This adds an overridden `createFlag` that takes in a list of instructions. All other `createFlag` functions only take in a single instruction, forcing users to manually concatenate multiple instructions together instead of using the `addInstructions` method already built-in to `CheckFlag`.

One note -- if I can change `CheckFlag`'s constructor to take in an `Iterable<String>` instead of a `List<String>`, then this function could take an iterable, which would be more general and easier to use. 

### Potential Impact:

Only a new method to use.

### Unit Test Approach:

There are no unit tests for `BaseCheck` at the moment -- adding them would probably be worthwhile. In the meantime, I tested this by modifying a couple of checks to use this constructor and running them on small datasets.

### Test Results:

All results look valid.


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -340,6 +340,12 @@ protected CheckFlag createFlag(final Set<AtlasObject> objects, final String inst
                 Collections.singletonList(instruction), points);
     }
 
+    protected CheckFlag createFlag(final Set<AtlasObject> objects, final List<String> instructions,
```
jklamer: minor ask, can you also implement an overloaded version without the need for locations which mirrors the other CheckFlag constructor.

nachtm: Did you one better and added that, plus the equivalent solo AtlasObject methods!

jklamer: huge

PR #128, opened 2019-02-21
remove tippecanoe from default outputs formats

danielduhh:
### Description:

Remove tippecanoe (line-delmited geojson for vector tiles) output format from default format in IntegrityCheckSparkJob.

### Potential Impact:

You now must explicitly specify tippecanoe in the IntegrityCheckSparkJob `outputFormats` to generate that data

### Unit Test Approach:

No unit tests, but tested running the spark job on my local machine and it operates as expected get generate that specific geojson

### Test Results:

Results passed


jklamer:
Whats the thinking on this? Too much output?

danielduhh:
Yeah - somewhat unnecessary duplicated output. Tippecanoe contains everything in the flag output, plus additional features (like bounding boxes)

danielduhh:
I also ran into `java.lang.OutOfMemoryError: GC overhead limit exceeded`  & `java.lang.OutOfMemoryError: Java heap space` issues when running both flag & tippecanoe outputs in large countries @jklamer

PR #130, opened 2019-02-26
Highway To Ferry Tag Check

sayas01:
### Description:

This PR is to add a new Integrity Check - HighwayToFerryTagCheck. The purpose of this check is to flag all Edges with _route=FERRY_ and _highway=PATH_ (or higher). According to [OSM wiki](https://wiki.openstreetmap.org/wiki/Key:ferry), when _route=FERRY_ exists, the ferry Tag can take values of highway classification and so, it is recommended to have a ferry Tag with the highway Tag value instead of a highway Tag for ferry routes.

### Potential Impact:

Flags Edges that are Ferry routes if there is no corresponding Ferry tag with a highway classification or if it has both Ferry and Highway tags.

### Unit Test Approach:

Various unit tests have been added to test the functionality of the check.

### Test Results:

All tests passing successfully.


src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayToFerryTagCheck.java
```diff
@@ -0,0 +1,127 @@
...
+public class HighwayToFerryTagCheck extends BaseCheck<Long>
+{
+    private static final String MINIMUM_HIGHWAY_TYPE_DEFAULT = HighwayTag.PATH.toString();
+    private static final String FERRY_TAG_IF_SAME_AS_HIGHWAY_INSTRUCTION = "This way {0,number,#} has a Ferry and a Highway tag for a ferry route. "
+            + "Please verify and remove the highway tag.";
+    private static final String FERRY_TAG_IF_ABSENT_INSTRUCTION = "This way {0,number,#} has a Highway tag for a ferry route instead of a Ferry tag. "
+            + "Please verify and add a Ferry tag with the Highway tag value and remove the highway tag.";
+    private static final String FERRY_TAG_IF_DIFFERENT_FROM_HIGHWAY_INSTRUCTION = "This way {0,number,#} has a Ferry and a Highway tag for a ferry route. "
+            + "Please verify and update the Ferry tag with the Highway tag value and remove the highway tag.";
+    private static final Predicate<Taggable> HAS_FERRY_TAG = object -> Validators
```
jklamer: This will have a problem running on spark because the Predicate interface isn't serializable. You can cast this to ```(Serializeable & Predicate<Taggable>)``` or just create a static function that will return this predicate. I would suggest the static function because its slightly cleaner.

sayas01: I have updated the code by removing the predicate and using Validators directly in the condition check(line [77](https://github.com/osmlab/atlas-checks/blob/96d453987317cd271f93c8d2a9e5fe2026a61ff4/src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayToFerryTagCheck.java#L77)).

PR #137, opened 2019-03-19
Small refactor of MapRouletteCommand to add DisplayName parameter

nachtm:
### Description:

Currently, subclasses of MapRouletteCommand are able to connect to MapRoulette through a four-part command-line switch of the form `maproulette=SERVER:PORT:PROJECT_NAME:API_KEY`. However, the endpoint in MapRoulette that creates a project also takes in a `displayName` parameter, which gives the project a more human-friendly name (think `Test Upload` instead of `test_upload`).

This allows users to pass in a project display name separately, like `projectDisplayName=DISPLAY_NAME`.

In order to make this happen, I made changes to the following classes:
 - `Project` now has a new `displayName` field, which is automatically serialized by its `toJson` function. There is also now a corresponding constructor and getter for `displayName`. 
 - I added a class `ProjectConfiguration`. Semantically, the distinction between a `ProjectConfiguration` and a `Project` is that a `ProjectConfiguration` defines and initializes `Project`s. So, we can pass around what a project should look like in the future without it actually being initialized. Previously this was not necessary, since the only field we cared about was name. 
 - I updated `MapRouletteConfiguration` to be compatible with `ProjectConfiguration`. 
 - Some private functions in `MapRouletteClient` now take `ProjectConfiguration`s instead of just the project name. I also added a package-private constructor that takes in a `TaskLoader` so I could write tests with a mocked-up `TaskLoader` instead of a full-fledged `MapRouletteConnection`. 
 - Obviously, MapRouletteCommand has been updated to parse and handle the new argument format properly

### Potential Impact:

I worked to maintain backwards compatibility with the old usage of `MapRouletteCommand`s, so there should be no impact, except that MapRouletteCommands can now specify DisplayName!

### Unit Test Approach:

Tried submitting projects with and without -projectDisplayName to MapRoulette.

### Test Results:

Compare the new usage:

Parameters:
image

MapRoulette edit project page:
image

To the old usage:

Parameters:
image

MapRoulette edit project page:
image

Also compare their displays in the manage page:
image

This can be merged after:
 - [x] I write some unit tests to make sonar happy

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ProjectConfiguration.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ProjectConfiguration.java
@@ -27,9 +27,13 @@ public class ProjectConfiguration
 
     /**
      * Defines a project and all of its fields.
-     * @param name The name of the project
-     * @param description The description of the project
-     * @param displayName The name displayed on Map Roulette for the project
+     * 
+     * @param name
+     *            The name of the project
+     * @param description
+     *            The description of the project
+     * @param displayName
+     *            The name displayed on Map Roulette for the project
      */
     public ProjectConfiguration(final String name, final String description,
             final String displayName)
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ProjectConfiguration.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ProjectConfiguration.java
@@ -1,12 +1,14 @@
 package org.openstreetmap.atlas.checks.maproulette.data;
 
+import java.io.Serializable;
+
 /**
  * Helper class to decouple MapRouletteConfiguration from configuring projects. This allows easier
  * updating of code that handles projects without touching general Map Roulette configuration code.
  *
  * @author nachtm
  */
-public class ProjectConfiguration
+public class ProjectConfiguration implements Serializable
 {
     private String name;
     private String description;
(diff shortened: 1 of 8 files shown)


src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ProjectConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Helper class to decouple MapRouletteConfiguration from configuring projects. This allows easier
+ * updating of code that handles projects without touching general Map Roulette configuration code.
+ *
+ * @author nachtm
+ */
+public class ProjectConfiguration implements Serializable
+{
+    private String name;
```
mgcuthbert: I would make all these variables final, as it seems that they are only set in the constructor and only accessed through getters.

src/test/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteClientTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.openstreetmap.atlas.checks.maproulette.data.Project;
+import org.openstreetmap.atlas.checks.maproulette.data.ProjectConfiguration;
+import org.openstreetmap.atlas.checks.maproulette.data.Task;
+
+/**
+ * Unit tests for MapRouletteClient
+ *
+ * @author nachtm
+ */
+public class MapRouletteClientTest
```
mgcuthbert: Very nice set of tests for this new code. Thanks.

PR #138, opened 2019-03-20
MapRouletteUploadCommand: Handle gzipped files

nachtm:
### Description:

Previously, log files that were .gzipped were skipped by MapRouletteUploadCommand. Users that wanted to upload zipped files to MapRoulette would have to manually unzip them, which could be tedious. 

This PR gives MapRouletteUploadCommand the capability to handle `.log.gz` files on its own.

### Potential Impact:

None -- just new capabilities.

### Unit Test Approach:

Tried uploading gzipped files and non-zipped files.

### Test Results:

Given the following directory,
image
 
running with the following params
image

successfully shows up in MapRoulette, with all tasks present:
image


src/main/java/org/openstreetmap/atlas/checks/commands/MapRouletteUploadCommand.java
```diff
@@ -81,14 +88,68 @@ protected void execute(final CommandMap commandMap,
...
                 }
-            }
+            });
         });
     }
 
+    /**
+     * An enum containing the different types of input files that we can handle.
+     */
+    private enum HandledFileType
```
danielduhh: Something like `OutputFileType` would be more clear here

src/main/java/org/openstreetmap/atlas/checks/commands/MapRouletteUploadCommand.java
```diff
@@ -81,14 +88,68 @@ protected void execute(final CommandMap commandMap,
...
+    {
+        LOG,
+        ZIPPED_LOG
```
danielduhh: `COMPRESSED_LOG`

src/main/java/org/openstreetmap/atlas/checks/commands/MapRouletteUploadCommand.java
```diff
@@ -81,14 +88,68 @@ protected void execute(final CommandMap commandMap,
...
+
+    /**
+     * Determine whether or not this file is something we can handle, and classify it accordingly.
+     * 
+     * @param logFile
+     *            any file
+     * @return if this file is something this command can handle, the appropriate HandledFileType
+     *         enum value; otherwise, an empty optional.
+     */
+    private Optional<HandledFileType> maybeGetHandledType(final File logFile)
```
danielduhh: We should avoid words like `maybe` in function names 😆 . How about `getOutputFileType`?

nachtm:
Thanks for the suggestions @danielduhh! As far as testing goes, I'm aware and thinking about the best way to test the bits I added. From what I can see, testing what's currently exposed would require uploading to Map Roulette. I'm open to suggestions on how to do this properly

nachtm:
I've figured out a decent solution, but it relies on [TestMapRouletteConnection](https://github.com/osmlab/atlas-checks/pull/137/files#diff-a419b3c17abc6fe22df9eec29bf679a6) from #137. So, I'm going to close this until that gets pulled in and I can use it here.

nachtm:
Since #137 got merged, I'm re-opening this!

danielduhh:
@nachtm Is there anyway you can avoid committing the binary file ?

nachtm:
@danielduhh Should be better now

PR #143, opened 2019-04-01
Flag Diff Commands Overhaul

Bentleysb:
### Description:

This is a major overhaul to the flag diff commands. 

First and foremost, this brings the diffs inline with recent breaking change in the json keys used geojson representations of Atlas objects. These keys are now no longer redefined in the diff command code, and instead pull directly from constants defined in Atlas. 
These breaks were not caught previously due to the fact that the diffs used pre-generated files as the source of unit test data. To prevent this happening again, the unit test now generate data dynamically from test atlases. All the pre-generated flag files have been removed. 

Second, the diff commands no longer use flag ids. These ids are not consistent enough from run to run to be reliably used for diffing. Instead atlas identifiers are used in all parts of the diffing process. This means there is no longer a use for the change output, so it has been removed. 

Third, the AtlasChecksGeoJSONDiffSubCommand has been enhanced to properly diff multiple checks at the same time. Previously when this was attempted all checks were diffed as one unit, which could result in the overriding of value and erroneous results. The LogDiffSubCommand already has this capability.

Fourth, the console printout was changed to list diffs per check. This quickly gives the user a more fine grained idea of the changes. It will now look like:
```
Total Items: 6

Additions: 3
Check1: 1
Check2: 2

Subtractions: 1
Check2: 1
```

Finally, the handling of the 2D hash maps that hold flag data was refactored. Previously the maps were initialized and passed into void methods to be populated. This practice is not preferable, as it makes the data manipulation harder to track. These methods have been altered to instead return hash maps that are then merged together.

### Potential Impact:

No downstream impacts.

### Unit Test Approach:

Unit test now dynamically create test data from test atlases. Thus, all unit test have been altered, but provide the same coverage.
In addition, the geojson diff tests have been altered to test for correct handling of multiple checks, atlas ids, and non-atlas points. Anything related to the change output has been removed from the unit tests. 

### Test Results:

Both diffs were tested on flag files generated from oms data that was about 3 months apart. A sampling of the results showed that the diffs are working as expected.


Bentleysb:
@danielduhh 
I have update the `source` and `target` arguments to be `reference` and `input`. 

I do not have any metrics on large country runtime. There is definitely lots of room for performance enhancements, though.

src/main/java/org/openstreetmap/atlas/checks/commands/JSONFlagDiffSubCommand.java
```diff
@@ -119,57 +80,68 @@ public String getDescription()
...
 
     @Override
     public void usage(final PrintStream writer)
     {
         writer.print(
-                "-source=path/to/first/flag/file,path/to/second/flag/file : file of flags to compare changes from\n");
+                "-reference=path/to/first/flag/file,path/to/second/flag/file : file of flags to use as a baseline\n");
         writer.print(
-                "-target=path/to/first/flag/file,path/to/second/flag/file : file of flags to compare changes to\n");
+                "-target=path/to/first/flag/file,path/to/second/flag/file : file of flags to compare changes from the baseline\n");
```
danielduhh: this is `input` now

src/main/java/org/openstreetmap/atlas/checks/commands/JSONFlagDiffSubCommand.java
```diff
@@ -27,87 +28,47 @@
...
 public abstract class JSONFlagDiffSubCommand implements FlexibleSubCommand
 {
-    private static final Command.Switch<File> SOURCE_FILE_PARAMETER = new Command.Switch<>("source",
-            "A file or directory of files containing atlas-checks flags to compare changes from.",
+    private static final Command.Switch<File> REFERENCE_FILE_PARAMETER = new Command.Switch<>(
+            "reference",
+            "A file or directory of files containing atlas-checks flags to use as a baseline for comparison.",
             File::new, Command.Optionality.REQUIRED);
 
-    private static final Command.Switch<File> TARGET_FILE_PARAMETER = new Command.Switch<>("target",
```
danielduhh: nit: couple extra lines here

src/main/java/org/openstreetmap/atlas/checks/commands/JSONFlagDiffSubCommand.java
```diff
@@ -119,57 +80,68 @@ public String getDescription()
...
+    @SuppressWarnings("squid:S106")
     public int execute(final CommandMap command)
     {
         // Get files and parse to maps
-        getFilesOfType((File) command.get(SOURCE_FILE_PARAMETER))
-                .forEach(path -> this.mapFeatures(path, this.source));
-        getFilesOfType((File) command.get(TARGET_FILE_PARAMETER))
-                .forEach(path -> this.mapFeatures(path, this.target));
+        this.getFilesOfType((File) command.get(REFERENCE_FILE_PARAMETER))
+                .forEach(path -> this.source = this.mergeMaps(this.mapFeatures(path), this.source));
```
danielduhh: `this.source -> this.reference`

src/main/java/org/openstreetmap/atlas/checks/commands/JSONFlagDiffSubCommand.java
```diff
@@ -119,57 +80,68 @@ public String getDescription()
...
+        this.getFilesOfType((File) command.get(INPUT_FILE_PARAMETER))
+                .forEach(path -> this.target = this.mergeMaps(this.mapFeatures(path), this.target));
```
danielduhh: `this.target` -> `this.input`

danielduhh:
@Bentleysb do you mind fixing the code smell reported by sonar? -> https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas-checks&open=AWrmRUA3VDQP4Uufm0mq&pullRequest=143&resolved=false

PR #144, opened 2019-04-09
Fixing GeoJSON creation

jklamer:
### Description:

Currently, each flagged object should be able to return a properly formed GeoJSON feature with flag ID information. This got broken with the GeoJSON updates where we are now adding properties find to a geometry object which is not to spec. This constructs the feature for each flagged object in the proper way. 

### Potential Impact:

Flag parsers should work the same way as they did before the breaking update. 

### Unit Test Approach:

All unit tests the same.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
@@ -12,6 +12,7 @@ import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Line;
 import org.openstreetmap.atlas.geography.atlas.items.Node;
 import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.geography.geojson.GeoJsonUtils;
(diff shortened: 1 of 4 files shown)


jklamer:
@Bentleysb can you verify that with properly formed GeoJSON the utilities you have written before the change should work / my description matches what you've seen?

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPolyline.java
```diff
@@ -56,13 +57,11 @@ public FlaggedPolyline(final AtlasItem atlasItem)
...
     {
-        final JsonObject feature = this.atlasItem.asGeoJsonGeometry();
+        final JsonObject geoJsonGeometry = this.atlasItem.asGeoJsonGeometry();
         final JsonObject jsonProperties = this.atlasItem.getGeoJsonProperties();
-
         jsonProperties.addProperty("flag:id", flagIdentifier);
         jsonProperties.addProperty("flag:type", FlaggedPolyline.class.getSimpleName());
-        feature.add("properties", jsonProperties);
-        return feature;
+        return GeoJsonUtils.feature(geoJsonGeometry, jsonProperties);
```
danielduhh: Nice - this is cleaner !

PR #152, opened 2019-04-23
MapRouletteUploadCommand Country and Check Filters

Bentleysb:
### Description:

This adds two filter arguments to the MapRouletteUploadCommand. One is a list of ISO3 country codes, the other is a list of names of atlas checks. If either or both filters are provided, only flags that have values that are in the filters will be uploaded. 

### Potential Impact:

No impact, just optional arguments.

### Unit Test Approach:

A unit test for each filter was added. Test rules and log file generation were altered to accommodate this. 

### Test Results:

None

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
@@ -77,6 +77,11 @@ public class MapRouletteUploadCommand extends MapRouletteCommand
         final Gson gson = new GsonBuilder().registerTypeAdapter(Task.class,
                 new TaskDeserializer(configuration.getProjectName())).create();
         final Configuration instructions = this.loadConfiguration(commandMap);
+        // Get the countries filter
+        final List<String> countries = (List<String>) commandMap.get(COUNTRIES);
+        // Get the checks filter
+        final List<String> checks = (List<String>) commandMap.get(CHECKS);
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
```diff
@@ -61,6 +77,11 @@ protected void execute(final CommandMap commandMap,
         final Gson gson = new GsonBuilder().registerTypeAdapter(Task.class,
                 new TaskDeserializer(configuration.getProjectName())).create();
         final Configuration instructions = this.loadConfiguration(commandMap);
+        // Get the countries filter
+        final List<String> countries = (List<String>) commandMap.get(COUNTRIES);
```
jklamer: I believe you can use getOptional which will mean you don't have to worry about nulls, just for safety's sake.

src/test/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommandTest.java
```diff
@@ -22,62 +23,106 @@
...
-                FOLDER.child("zipped.log.gz").toString()).withCompression(true);
-        zippedProcessor.process(setup.getAnotherBasicFlag());
-        zippedProcessor.process(new ShutdownEvent());
+            this.filesCreated = true;
+        }
     }
 
-    @After
-    public void cleanup()
+    @AfterClass
```
jklamer: 👍

danielduhh:
I should be able to test this once https://github.com/osmlab/atlas-checks/pull/156 gets merged...

PR #153, opened 2019-04-24
Flag Statistics Sub-Command

Bentleysb:
### Description:

This adds a command for counting flags in log files on mass. 
Given a directory of ISO country folders containing log files, it will get the number of flags per check per country. The result is a table of the collected data. The data is written to a csv file(s) in the provided output folder. 
Additionally, this command can take in two directories as inputs and calculate the difference between the flag counts. 
This command is built on top of the [Atlas Shell Tools](https://github.com/osmlab/atlas/tree/dev/atlas-shell-tools) system from Atlas. 

There are 3 output types: full, totals, and counts
By default the full output is used, but any of them can be enabled or disabled using the `--output-type` argument.

Example full output:

|Check|ABC|XYZ|Total|
|---|---|---|---|
|Check1|6|8|14|
|Check2|2||2|
|Check3||2|2|
|Total|8|10|18|

Example full difference output:

|Check|ABC|XYZ|Total|
|---|---|---|---|
|Check1|0|4|4|
|Check2|-2||-2|
|Check3||0|0|
|Total|-2|4|2|

Example totals output with reference and difference columns:

|Check|Reference|Input|Difference|
|---|---|---|---|
|Check1|10|14|4|
|Check2|4|2|-2|
|Check3|2|2|0|

Example counts output with reference and difference columns:

|Country|Check|Reference|Input|Difference|
|---|---|---|---|---|
|ABC|Check1|6|6|0|
|ABC|Check2|4|2|-2|
|ABC|Check3||||
|XYZ|Check1|4|8|4|
|XYZ|Check2||||
|XYZ|Check3|2|2|0|

### Potential Impact:

None

### Unit Test Approach:
Unit test check that both the count and diff functions work as expected.

### Test Results:

Tested on atlas checks runs of over 200 countries. A sampling of returned counts and diffs were as expected.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
@@ -42,7 +42,7 @@ import com.google.gson.JsonObject;
 public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
 {
     private static final String INPUT_OPTION = "input";
-    private static final String DIFFERENCE_OPTION = "difference";
+    private static final String REFERENCE_OPTION = "reference";
     private static final String OUTPUT_OPTION = "output";
     private static final String GENERATOR = "generator";
 
@@ -71,16 +71,16 @@ public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
         final Map<String, Map<String, Counter>> inputCounts = this.getCountryCheckCounts(
                 this.optionAndArgumentDelegate.getOptionArgument(INPUT_OPTION).get());
 
-        final Optional<String> differenceFilePath = this.optionAndArgumentDelegate
-                .getOptionArgument(DIFFERENCE_OPTION);
+        final Optional<String> referenceFilePath = this.optionAndArgumentDelegate
+                .getOptionArgument(REFERENCE_OPTION);
         try
         {
             // If a second input is supplied..
-            if (differenceFilePath.isPresent())
+            if (referenceFilePath.isPresent())
             {
                 // Read the second input
                 final Map<String, Map<String, Counter>> targetCounts = this
-                        .getCountryCheckCounts(differenceFilePath.get());
+                        .getCountryCheckCounts(referenceFilePath.get());
                 // Get the difference between the outputs and output it
                 this.writeOutput(getDifference(inputCounts, targetCounts),
                         this.optionAndArgumentDelegate.getOptionArgument(OUTPUT_OPTION));
@@ -104,7 +104,7 @@ public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
     @Override
     public String getCommandName()
     {
-        return "flag-stats";
+        return "flag-statistics";
     }
 
     @Override
@@ -128,9 +128,9 @@ public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
         this.registerOptionWithRequiredArgument(INPUT_OPTION, 'i',
                 "A directory of folders containing atlas-checks log files.",
                 OptionOptionality.REQUIRED, INPUT_OPTION);
-        this.registerOptionWithRequiredArgument(DIFFERENCE_OPTION, 'd',
+        this.registerOptionWithRequiredArgument(REFERENCE_OPTION, 'r',
                 "A second set of log files to diff against.", OptionOptionality.OPTIONAL,
-                DIFFERENCE_OPTION);
+                REFERENCE_OPTION);
         this.registerOptionWithRequiredArgument(OUTPUT_OPTION, 'o', "A csv to output results to.",
                 OptionOptionality.OPTIONAL, OUTPUT_OPTION);
         super.registerOptionsAndArguments();
@@ -192,7 +192,8 @@ public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
                     catch (final IOException exception)
                     {
                         this.outputDelegate.printlnWarnMessage(
-                                "File read failed with exception" + exception.getMessage());
+                                String.format("Exception thrown while reading file %s: %s",
+                                        file.getName(), exception.getMessage()));
                     }
                 });
 
--- a/src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
@@ -59,6 +59,7 @@ public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
     }
 
     @Override
+    @SuppressWarnings("squid:S3655")
     public int execute()
     {
         final HashMap<String, HashMap<String, Counter>> inputCounts = this.getCountryCheckCounts(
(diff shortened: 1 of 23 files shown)


src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,330 @@
...
+ * This command takes a folder of directories of atlas-checks log files and counts the number of
+ * flags per country per check. It optionally takes a second folder and returns the difference
+ * between the inputs.
+ *
+ * @author bbreithaupt
+ */
+public class FlagStatisticsSubCommand extends AbstractAtlasShellToolsCommand
+{
+    private static final String INPUT_OPTION = "input";
+    private static final String DIFFERENCE_OPTION = "difference";
```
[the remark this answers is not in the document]
Bentleysb: Reference does not make a lot of sense, as this is the target of the diff, not the source.

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,330 @@
...
+import org.openstreetmap.atlas.utilities.command.parsing.OptionOptionality;
+import org.openstreetmap.atlas.utilities.scalars.Counter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+/**
+ * This command takes a folder of directories of atlas-checks log files and counts the number of
```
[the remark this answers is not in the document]
Bentleysb: It does not have to be that flag folder, though that is the expected input most of the time.

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,330 @@
...
+        // Check all files in the folder and all sub-folders
+        new File(path).listFilesRecursively().stream()
+                // Filter the files to only include log files, either gzipped or uncompressed
+                .filter(file -> FilenameUtils
+                        .getExtension(file.isGzipped() ? FilenameUtils.getBaseName(file.getName())
+                                : file.getName())
+                        .equalsIgnoreCase("log"))
+                .forEach(file ->
+                {
+                    logger.info("Reading: {}", file.getName());
```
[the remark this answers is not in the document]
Bentleysb: Not necessarily, but it makes it so you arn't looking at a blank screen for 3 minutes.

Bentleysb:
@danielduhh, that run time sounds about right for a diff of 2 full atlas-checks runs. It just takes a while to read in the thousands of flags.

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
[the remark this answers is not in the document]
danielduhh: The source is `input`, and the target (or reference) is what you're comparing this to, right? I think reference makes sense...

Bentleysb:
@danielduhh, I have added the two requested outputs. They can be enabled using the `--output-types` argument. There are example in the PR description. Additionally, the original output is now generated in three files, if a difference is requested. One file gives the input counts, one the reference counts, and the third the difference counts. To accommodate this the output parameter now takes a folder instead of a file. 
I also changed to using parallelStreams and ConcurrentHashMaps for reading the log files. This has decreased the overall runtime by about 200%.

danielduhh:
Nice job optimizing the run time.
What's the difference between the `total` and `counts` output? In your example, it looks like one has the country field, one does not. If that's the case, a single output with `country,check,source,reference,diff` should work

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+        // mathematical difference of the companion target counter (defaulting to 0 if absent) and
+        // the source counter
+        source.forEach((country, checkCounts) ->
+        {
+            difference.putIfAbsent(country, new HashMap<>());
+            checkCounts
+                    .forEach((check, count) -> difference.get(country).put(check,
+                            new Counter(target.getOrDefault(country, new HashMap<>())
+                                    .getOrDefault(check, new Counter()).getValue()
+                                    - count.getValue())));
```
danielduhh: I think this is backwards. If the `target` (or baseline) is 20, and the `source` (or input) is 10, the result should be -10.

Bentleysb: I think this gets back to terminology again. The input is the source/baseline, the reference is the target. So input/source/baseline should be older than reference/target.

Bentleysb:
@danielduhh, counts is countrified, totals is not (so it is the total of all countries for each check).

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
danielduhh: Your example shows -2 as the difference between 4 (input) & 2 (reference); However, the new run (input) has 2 additional flags from the comparison, so it should be +2, right?
```csv
Country,Check,Input,Reference,Difference
ABC,Check2,4,2,-2
```

Bentleysb: The new run is reference.

Bentleysb:
Based on offline discussion, I have updated the reference input to be treated as the older data set.

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+                        differenceCounts);
+                // Add the reference and difference metrics to the counts output
+                countsOutput = generateCountsOutputStage2(countsOutput, referenceCounts,
+                        differenceCounts);
+            }
+
+            // Write the totals output if requested
+            if (outputTypes.contains("totals"))
+            {
+                this.writeCSV(outputFolder + "/totals.csv", totalsOutput);
```
danielduhh: totals.csv -> checkSummary.csv. Also, can you break this out into a constant

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+                        .getCountryCheckCounts(referenceFilePath.get());
+
+                // Get the difference
+                final Map<String, Map<String, Counter>> differenceCounts = this
+                        .getDifference(referenceCounts, inputCounts);
+
+                // Write outputs for the reference and difference if requested
+                if (outputTypes.contains("full"))
+                {
+                    this.writeCSV(outputFolder + "/reference.csv",
```
danielduhh: I don't think this output is necessary. This will be a duplicate of the "input.csv" output if we continuously output each `output-types` from run to run

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+                            generateFullOutput(referenceCounts));
+                    this.writeCSV(outputFolder + "/difference.csv",
```
danielduhh: difference.csv -> runSummaryDifference.csv

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+
+            // Write the totals output if requested
+            if (outputTypes.contains("totals"))
+            {
+                this.writeCSV(outputFolder + "/totals.csv", totalsOutput);
+            }
+            // Write the counts output if requested
+            if (outputTypes.contains("counts"))
+            {
+                this.writeCSV(outputFolder + "/counts.csv", countsOutput);
```
danielduhh: Can you think of something more informative that `counts.csv`?

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+     *            total counts
+     * @param referenceCountryCheckCounts
+     *            a 2D {@link Map} of flag {@link Counter}s per check {@link String} per country
+     *            {@link String}
+     * @param diffCountryCheckCounts
+     *            a 2D {@link Map} of flag {@link Counter}s per check {@link String} per country
+     *            {@link String}
+     * @return a 2D {@link List} of {@link String}s
+     */
+    private List<List<String>> generateTotalsOutputStage2(final List<List<String>> stage1Output,
```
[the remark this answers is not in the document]
danielduhh: This will require updating the name of the "stage1" function as well...

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+     *            country, check, and associated count
+     * @param referenceCountryCheckCounts
+     *            a 2D {@link Map} of flag {@link Counter}s per check {@link String} per country
+     *            {@link String}
+     * @param diffCountryCheckCounts
+     *            a 2D {@link Map} of flag {@link Counter}s per check {@link String} per country
+     *            {@link String}
+     * @return a 2D {@link List} of {@link String}s
+     */
+    private List<List<String>> generateCountsOutputStage2(final List<List<String>> stage1Output,
```
danielduhh: Same comment about stage1/stage2

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -0,0 +1,619 @@
...
+     */
+    private List<List<String>> generateTotalsOutputStage2(final List<List<String>> stage1Output,
+            final Map<String, Map<String, Counter>> referenceCountryCheckCounts,
+            final Map<String, Map<String, Counter>> diffCountryCheckCounts)
+    {
+        // Copy the output lines list
+        final List<List<String>> outputLines = new ArrayList<>(stage1Output);
+
+        // Add new headers
+        outputLines.get(0).add(1, "Reference");
```
danielduhh: How do you feel about indicating these fields are sum's in the headers. Maybe  `Sum(Reference)` and `Sum(Difference)` ?

src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -111,32 +116,30 @@ public int execute()
...
-                        differenceCounts);
+                countsOutput = addReferenceAndDifferenceToCountsOutput(countsOutput,
+                        referenceCounts, differenceCounts);
             }
 
             // Write the totals output if requested
             if (outputTypes.contains("totals"))
             {
-                this.writeCSV(outputFolder + "/totals.csv", totalsOutput);
+                this.writeCSV(outputFolder + "/checkSummary.csv", totalsOutput);
```
danielduhh: Should we change the output type name as well? Maybe we can put all of the supported `output-types` in a enum set

Bentleysb: Done

PR #156, opened 2019-04-30
Map Roulette Configuration Full URL Parsing Bug

Bentleysb:
### Description:

This fixes a bug with the MapRouletteConfiguration and MapRouletteConnection. Previously these assumed that all URIs used, but did not include, the `http` scheme (e.g. maproulette.server.mine:42 vs http://maproulette.server.mine:42). This updates MapRouletteConfiguration and its `parse` method to look for and store schemes, or continue to use `http` as a default. MapRouletteConnection has been updated to look for the set scheme and use that in building a URI.
Additionally, MapRouletteConfiguration now throws a CoreException when it fails to parse a configuration string, instead of logging a debug message and trowing a NPE. This is preferable, as debug messages are not written to stdout by default, and this is important information for the user. 

### Potential Impact:

This is generally non-breaking, due to the addition of a backwards compatible constructor and the optional use of the scheme field from MapRouletteConfiguration. However,  it could break at runtime if a server URI using something other than http is entered (e.g. https), and is passed to something other than MapRouletteConnection.

### Unit Test Approach:

Added a unit test that includes a https server URL.
Update a unit test to reflect the change from a null return type to a CoreException. 

### Test Results:

Tested uploading a challenge and changing a project name on https://maproulette.org. This was completely successful when using port 443.

### DIFF
--- a/docs/standalone.md
+++ b/docs/standalone.md
@@ -62,10 +62,10 @@ of desired formats, like so:
 The standalone application for Atlas Checks can be published directly to MapRoulette using your personal MapRoulette
 API Key. This can be easily done by including the following parameter during your gradle run.
 
-`gradle run -Pchecks.local.maproulette=maproulette.org:80:HOME_12345:APIKey`
+`gradle run -Pchecks.local.maproulette=http://maproulette.org:80:HOME_12345:APIKey`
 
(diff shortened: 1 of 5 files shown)


danielduhh:
Hey @Bentleysb, can you test this with https://maproulette.org & update the Test Results description with your results? Thanks!

PR #175, opened 2019-06-26
Flag Statistics: All Outputs By Default

Bentleysb:
### Description:

This changes the default output types for the FlagStatisticsSubCommand to be all output types. 

### Potential Impact:
Anything that used the default will see an increase in output files. 

### Unit Test Approach:

No new unit tests.

### Test Results:

Tested with and without the output-types argument. The results were as expected.


src/main/java/org/openstreetmap/atlas/checks/commands/FlagStatisticsSubCommand.java
```diff
@@ -94,9 +94,14 @@ public int execute()
         final String outputFolder = this.optionAndArgumentDelegate.getOptionArgument(OUTPUT_OPTION)
                 .get();
         // Get the output types
-        final List<String> outputTypes = Arrays.asList(StringUtils
-                .split(this.optionAndArgumentDelegate.getOptionArgument(OUTPUT_TYPES_OPTION)
-                        .orElse(OutputTypes.RUN_SUMMARY.toString()).toUpperCase(), ','));
+        final List<String> outputTypes = Arrays
```
danielduhh: 👍

PR #176, opened 2019-06-28
New check GeneralizedCoastlineCheck

seancoulter:
### Description:

Adding a new check GeneralizedCoastlineCheck, which flags coastlines that have a given percentage of line segments that are longer than a configured value. Coastlines are defined to be LineItems that have the tag "natural=coastline". They may also be LineItems whose parent relation has this tag.

### Potential Impact:

This is a new check with no dependencies as of yet. I don't anticipate any downstream impact. However, since the suggested "fix" to this check involves adding more nodes to the coastline, there may be related effects.

### Unit Test Approach:

Unit tests using relations, lines, nodes.

### Test Results:

All unit tests pass. Testing also done on Argentina, Iceland pbfs with satisfactory results.


danielduhh:
@seancoulter Could you fix this code smell ?
https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas-checks&open=AWuf1Pw8dgEyLZvl98fS&pullRequest=176&resolved=false&types=CODE_SMELL

PR #178, opened 2019-07-08
Use OsmWayWalker for more consistent flags

nachtm:
### Description:
Currently, a number of checks flag a single edge and then mark the entire OSM way as flagged using `markAsFlagged`. This leads to potential inconsistency of flagged edges between runs, which causes issues for diffing tools. 

In order to address this issue, we can use the handy `OsmWayWalker` to flag all of the edges in a way. In particular, this PR adds the `OsmWayWalker` to the following checks:
 -  RoundaboutClosedLoopCheck
 - SignPostCheck
 - SingleSegmentMotorwayCheck
 - HighwayToFerryTagCheck
 - InvalidAccessTagCheck
 - InvalidLanesTagCheck
 - MixedCaseNameCheck
 - UnusualLayerTagsCheck

### Potential Impact:

Change in number of features impacted (but no change in flag count). Also, some offsetting diffs -- I ran on a small atlas and got the following diff using the geojson-diff command:

```
Total Items: 7739

Additions: 187
SinkIslandCheck: 3
RoundaboutClosedLoopCheck: 1
SignPostCheck: 71
InvalidLanesTagCheck: 7
InvalidAccessTagCheck: 6
UnusualLayerTagsCheck: 69
MixedCaseNameCheck: 30

Subtractions: 187
SinkIslandCheck: 3
RoundaboutClosedLoopCheck: 1
SignPostCheck: 71
InvalidLanesTagCheck: 7
InvalidAccessTagCheck: 6
UnusualLayerTagsCheck: 69
MixedCaseNameCheck: 30
```

### Unit Test Approach:

Unit tests passing. Ran a diff between a baseline version and this, getting the results above.

I had to modify the unit tests for SignPost check because a lot of the test cases had no `id` specified, so they were defaulting to `id=1`, `id=2`, etc. Then the logic for getting OSM id gave `osmid=0`, `osmid=0`, etc. So, OsmWayWalker was expanding to the whole network since the whole network had the same OSM id. Since the SignPostCheck unit tests count the number of flagged objects in each flag, this was throwing off the counts.

To address this, I added distinct ids with the properly suffixed `000000` to the end of each element. After that, the unit tests passed.
 
### Test Results:

Total change of +0 in flag counts. I also went through the flags quickly and verified that they were logical updates.

One thing worth pointing out is that SinkIsland requires a slightly different treatment, since sometimes we only want to flag part of a way. I have a potential solution but it requires further testing/thought/analysis to make sure that it makes sense.


nachtm:
@danielduhh Here is the diff from a 5-country run:
```
Total Items: 74039

Additions: 1805
SinkIslandCheck: 12
RoundaboutClosedLoopCheck: 21
SignPostCheck: 542
InvalidLanesTagCheck: 17
InvalidAccessTagCheck: 6
UnusualLayerTagsCheck: 187
MixedCaseNameCheck: 1020

Subtractions: 1802
SinkIslandCheck: 9
RoundaboutClosedLoopCheck: 21
SignPostCheck: 542
InvalidLanesTagCheck: 17
InvalidAccessTagCheck: 6
UnusualLayerTagsCheck: 187
MixedCaseNameCheck: 1020
```
The only diff is a slight change in SinkIslandCheck, which is not modified by this update. SinkIsland needs to be addressed in a slightly different fashion and probably requires some extra validation to make sure that it's handled properly. 

The others are impacted but have no actual change in flag count numbers. So, we shouldn't expect any change in flag count numbers due to this PR.

PR #179, opened 2019-07-08
SinkIslandCheck Enhancements to Exclude Certain Edges 

sayas01:
### Description:

This PR adds the following enhancements to SinkIslandCheck:
1) Exclude all service roads that are within any airport polygons. Current version has a validation that excludes edges that have aeroway tags with values runway or taxiway. Since ways with aeroway tags are not ingested as edges, this validation has been removed and a method that checks if the road(service highways) is within airport polygon has been added.
2) Previously only the end node of the edge was checked for _amenity=parking_entrance_ which resulted in some false positives. Additional validation to check for _amenity=parking_entrance_ tag at the start node of the edge has been added which should fix these cases.
3) Service roads that ends in buildings have been excluded from the check.
4) Previously, navigability of the edge was checked only in terms of highway tags. Additional validation in terms of _Access_ tag and _Motorvehicle_ and _Motorcar_ tags have been added to determine the car navigability of the edge.

### Potential Impact:

From the analysis, it was seen that there is both addition and subtraction of flags with these enhancements. Subtractions have been verified to be the cases that should be removed by these enhancements like services roads within airports. Additions have been verified to be mainly due to the updated car navigability condition. Most of the additions are for edges that are not marked _access=no_ but are connected to _access=no_ edges, which are valid sink islands.

### Unit Test Approach:

Unit tests have been added for the new updates. Additional validation has been done by testing the updates on a sample of countries and verifying it with the previous version.

### Test Results:

All tests passing and the additions and subtractions have been verified to be valid.


sayas01:
@danielduhh Following is the diff between pre and post enhancement on 4 countries:
```
Total Items: 2135

Additions: 936
SinkIslandCheck: 936

Subtractions: 215
SinkIslandCheck: 215
```
The new flags have been added due to the navigability conditions that have been added in the enhancement. All the additions and subtractions have been verified to be valid with 0% FP. Below is an example of flags that have been added due to the navigability condition:

PR #182, opened 2019-07-19
MalformedRoundaboutCheck Use ComplexRoundabout

Bentleysb:
### Description:

With the addition of the CompleRoundabout class to Atlas, much of the logic in MalformedRoundaboutCheck is redundant. This removes the redundant logic and replaces it with calls to ComplexRoundabout.

### Potential Impact:
There should not be any downstream impact. 

### Unit Test Approach:

Added one unit test for the left driving countries configurable.

### Test Results:

Tested the check on 5 countries. The flags were the same before and after the changes.


danielduhh:
@Bentleysb

danielduhh:
@Bentleysb ping

Bentleysb:
@danielduhh, I would have to rerun the diff, but I can. However, as I put in the description the runs had no differences.

PR #183, opened 2019-07-30
Add PedestrianAreaOverlappingEdgeCheck

sayas01:
### Description:

The purpose of this check is to flag pedestrian areas that are not properly snapped to its overlapping edges. Pedestrian areas are defined as Ways with highway=PEDESTRIAN and area=YES tags. Overlapping roads are Edges with any highway=** except highway=path, highway=steps, highway=footway, and highway=pedestrian tag with the same elevation (same LevelTag and LayerTag) as the pedestrian area. If the overlapping edge is not properly snapped to the pedestrian area, then, the area, the overlapping edge and all its connected edges that are within the area are marked as flagged.

### Potential Impact:

None

### Unit Test Approach:

Unit tests have been added to test the functionality.

### Test Results:

All tests passing.


src/main/java/org/openstreetmap/atlas/checks/validation/areas/PedestrianAreaOverlappingEdgeCheck.java
```diff
@@ -0,0 +1,245 @@
...
+    {
+        return
+        // We do not want any pedestrian edges that are also pedestrian areas.
+        // Currently, pedestrian ways are ingested as both edges and as areas.
+        edge.getOsmIdentifier() != area.getOsmIdentifier() && edge.isMasterEdge()
+                && !HighwayTag.isPedestrianNavigableHighway(edge)
+                // Valid edge should be a car navigable highway
+                && (HighwayTag.isCarNavigableHighway(edge)
+                        // or a cycleway with no foot access
+                        || (Validators.isOfType(edge, HighwayTag.class, HighwayTag.CYCLEWAY)
```
danielduhh: Can you see this list of acceptable intersecting tags changing in the future? If so, would this make sense as a configurable ?

sayas01: For this check, valid edges are non pedestrian navigable ways and I believe we have covered all non pedestrian navigable ways with these validations. If at all there could be changes, it would be in the car navigable highway validation.

src/main/java/org/openstreetmap/atlas/checks/validation/areas/PedestrianAreaOverlappingEdgeCheck.java
```diff
@@ -0,0 +1,245 @@
...
+                            && areaPolygon.fullyGeometricallyEncloses(connectedEdge.asPolyLine()))
+                            .forEach(overlappingEdges::add);
+                    // Add the intersecting edge as well to the set
+                    overlappingEdges.add(edge);
+                }
+            });
+        }
+        if (!overlappingEdges.isEmpty())
+        {
+            this.markAsFlagged(object.getOsmIdentifier());
```
danielduhh: Did you want to mark Atlas ids instead?

sayas01: I am marking the osm identifier of the area and atlas identifiers of the edges that it overlap.

danielduhh: I see! Sorry I couldn't locate where that was happening with edges

src/main/java/org/openstreetmap/atlas/checks/validation/areas/PedestrianAreaOverlappingEdgeCheck.java
danielduhh: Okay, if we dont see
`highway=cycleway|service=alley|parking_aisle|driveway|emergency_access|drive_through` ever changing, then I'm okay with it as is

PR #184, opened 2019-07-30
Checks database upload Command

danielduhh:
### Description:

This adds a new command to upload integrity check flags into a PostgreSQL database, which gives users the ability to write powerful SQL & PostGIS queries on the output of various check runs. Storing check flag information in a database can help answer questions about relationships between flags in various checks. For example.... "How many checks does this osm id appear in?" or "How many times is this atlas feature duplicated in this check?"

The command takes two switches:
 - `flag_path` - directory to checks log files
 - `database_ur`l -database connection string: host[:port]/database. See the [jdbc docs](https://jdbc.postgresql.org/documentation/head/connect.html) for optional parameters like user, password, currentSchema, etc.

```bash
org.openstreetmap.atlas.checks.database.FlagDatabaseSubCommand
--flag_path=/Users/user/Downloads/flag/IDN
--database_url=127.0.0.1/flags_test?user=postgres&currentSchema=public
```
Where...

`host` - 127.0.0.1
`port` (optional) - 5432
`database` - flags_test
`username` - postgres
`schema `- public

The database schema is simple. There's two tables `flag` & `feature`

Screen Shot 2019-07-29 at 4 52 41 PM
Screen Shot 2019-07-29 at 4 52 43 PM

I also, added a very simple `CheckFlagDeserializer` that allows us to easily load integrity check outputs and convert them into CheckFlags.

These are the results of some runtime tests I have on my local machine with the following java memory parameters: `-Xms10240m -Xmx10240m`

| countries | file size | flags        | features   | batch size | runtime      |
|-----------|-----------|--------------|------------|------------|--------------|
| 1         | 7.2MB     | 14,715       | 59,929     | 1000       | 15.5 seconds |
| 1         | 80MB      | 440,393      | 1,053,726  | 1000       | 2.2 minutes  |
| 1         | 171MB     | 333,207      | 723,938    | 1000       | 2.6 minutes  |
| 210       | 1.55GB    | 5,654,892,12 | 12,122,284 | 1000       | 58.9 minutes |

TODO

- [x] Add tags as hstore

### Potential Impact:

This is a new command, so there is no downstream impact.

### Unit Test Approach:

- [x] TODO CheckFlagDeserializer unit tests
- [x] TODO DatabaseConnection unit tests
- [x] TODO FlagDatabaseSubCommand unit tests

Created unit tests for CheckFlagDeserializer, DatabaseConnection & FlagDatabaseSubCommand.

I also tested that the number of flags generated by the checks equals the number of flag records inserted into the database.

```sql
select count(*) from flag;
 count 
-------
   208
(1 row)
```

### Test Results:

Number of check flags uploaded as expected for various countries.


src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return this.connection;
+    }
+
+    private URI createConnectionURI(final String connectionString)
+    {
+        return URI.create(String.format("postgresql://%s", connectionString));
+    }
+
+    private void createDatabaseSchema()
```
mgcuthbert: I would not mix connection handling with database schema creation. This should be separated out.

src/main/java/org/openstreetmap/atlas/checks/database/FlagDatabaseSubCommand.java
```diff
@@ -0,0 +1,376 @@
...
+        {
+            // If this file is something we handle, read and upload the tasks contained within
+            final Optional<OutputFileType> optionalHandledFileType = getOptionalOutputType(file);
+            optionalHandledFileType.ifPresent(outputFileType ->
+            {
+
+                try (BufferedReader reader = this.getReader(file, outputFileType))
+                {
+
+                    final PreparedStatement flagSqlStatement = databaseConnection.prepareStatement(
```
mgcuthbert: You should put your prepared statement in the try enclosure above so that they are automatically closed.

src/main/resources/org/openstreetmap/atlas/checks/database/schema.sql
```diff
@@ -0,0 +1,33 @@
+DROP SCHEMA IF EXISTS {schema} CASCADE;
+CREATE SCHEMA IF NOT EXISTS {schema};
+CREATE EXTENSION IF NOT EXISTS postgis;
+CREATE EXTENSION IF NOT EXISTS hstore;
+
+DROP TABLE IF EXISTS flag CASCADE;
```
mgcuthbert: You sure you want to drop the table if it exists? Maybe only create the table if it doesn't exist.

src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java
```diff
@@ -0,0 +1,90 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.datasource.init.ScriptUtils;
+
+/**
+ * Connect and create schema for CheckFlag database
+ *
+ * @author danielbaah
+ */
+public class DatabaseConnection
```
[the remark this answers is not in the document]
danielduhh: Good point - I can look into hikaricp. My original thought was to make this class specifically for the upload command but now am realizing the access modifier is public.

src/main/resources/org/openstreetmap/atlas/checks/database/schema.sql
danielduhh: I see this going two ways:
1) Create a new schema for each run. With this, the flags and features contained in their tables will all be associated with a single run
2) Add various runs to the same schema. In this case, you can differentiate each run by the `date_created` timestamp

danielduhh:
@nachtm @Bentleysb Do you mind looking at the FileUtility class I added to remove duplicated logic in both MapRouletteUploadCommand and this one

src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java
danielduhh: I moved this to the FlagDatabaseSub command class

src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java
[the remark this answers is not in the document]
danielduhh: I didn't end up bringing in hikari, but I simplified the class and am no longer saving the connection as a class variable. It also implements AutoCloseable

Bentleysb:
@danielduhh, the file utility changes look good to me.

danielduhh:
Hey @mgcuthbert, I addressed the comments in your initial PR review. Let me know if there's anything outstanding

PR #186, opened 2019-08-07
[DNM] [WIP] Sharded Integrity checks

jklamer:
### Description:
Currently running the Integrity Checks jobs on large countries requires server level hardware because the Atlases and loaded and processed at once. This process preserves the Edge network available to the checks as well as the full picture of the data, but it has serious ramifications when it comes to running on local hardware, developing new checks, and the ability to parallelize the work. 

This PR adds a new job that produces checks in parallel and then deduplicates and processes them by country. This should allow for much smaller memory footprint needed to run jobs locally, and much better parallelizations when running on a cluster. 

This job is designed with the as many of the same inputs as the old job, allowing a simple switching of command class when wanting to use the new job for any old run configurations. 
Ways in which this job is different however:
- Checks are run on Dynamic atlas of certain shards instead of multi-atlas of all the shards
- Checks are deduplicated with the assumption that all Flags with the same ID produced from the same check are the same
- The output is not committed by a final copy to the final location, it is simply printed there. Meaning results will be in the output even on a job failure. 
- Needs a newer atlas with the sharding file in the file structure. 

### Potential Impact:

- Less accurate large network checks dependent on Dynamic atlas policy 
- Assumption made for deduplication could fail 

### Unit Test Approach:

New classes with business logic has unit tests for basic functionalities. Ensuring they preform as they are made to do. 

### Test Results:

[WIP] Steps
1. Ensure no regression with original job by comparing outputs for same inputs.
2. Test I/O and options of job. 
3. Confirm memory footprint characteristics of sharded job.
4. Confirm fidelity of output up to 95% with non sharded job.


danielduhh:
Replaced by #259

PR #187, opened 2019-08-13
Road name spelling consistency check (new check)

seancoulter:
### Description:

This is a new check that's designed to flag road segments whose Name tags have slightly different spellings. "Slightly different spelling" is understood to mean a edit distance of 1 between two strings. An edit distance is defined to be the number of character insertions, deletions, and substitutions needed for two strings to exactly match. The goal here is to bring attention to areas of a road with spelling mixups so that editors can create a uniform spelling for the entire road.

This check should flag road segments whose Name tags contain:
--Accented character inconsistencies
--Possessive apostrophe misplacement/absence
--Other character additions/deletions

### Potential Impact:

None. This is a new check with no dependencies.

### Unit Test Approach:

Unit tests involving roads with multiple, single, and no spelling inconsistencies among segments.

### Test Results:

Unit tests pass


docs/checks/RoadNameSpellingConsistencyCheck.md
```diff
@@ -0,0 +1,23 @@
...
+3. Edge [id:28509866](https://www.openstreetmap.org/way/28509866) is spelled Godwin Court, and a service road just to the east is called Goodwin Court.
+
+### Code Review
+
+The check starts by ensuring a given AtlasObject is a named Master Edge that has not been flagged and is of a high enough HighwayTag priority. The check then collects all Edges that have a common connection to the starting Edge and are also within a configurable search area. The collected Edges are filtered for those that have names that are at an edit distance of 1 from the starting Edge's name. However, there are exceptions in this process. If two Edges' names contain different numbers or space-delimited characters, they are not considered to have inconsistent spelling with one another and are not flagged.
+Any Edges whose names are within 1 edit distance of each other are flagged.
+
+For the source code for this check, please refer to [RoadNameSpellingConsistencyCheck](https://github.com/seancoulter/atlas-checks/blob/RoadNameSpellingConsistencyCheck/src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameSpellingConsistencyCheck.java)
```
ranek: Please use a relative path, or link to where this will be in the official osmlab repo.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameSpellingConsistencyCheckWalker.java
```diff
@@ -0,0 +1,277 @@
...
+     * only take in Strings that are the same length, or whose lengths are off by a single
+     * character.
+     *
+     * @param incomingEdgeName
+     *            the name of the next edge in the search area
+     * @param startingEdgeName
+     *            the name of the edge from which the search started
+     * @return true if the edit distance between two edge's names is 1; false otherwise
+     */
+    private static boolean editDistanceIsOne(final String incomingEdgeName,
```
ranek: This method, and the subsequent `editDistanceIsOneHelper()`, have confusing names. I would consider renaming this method to make it more clear that it checks for identifier substrings, and then maybe rename `editDistanceIsOneHelper()` to `editDistanceIsOne()` since it's the method that actually checks the edit distance as normally understood.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameSpellingConsistencyCheckWalker.java
```diff
@@ -0,0 +1,277 @@
...
+    {
+        return incomingEdge ->
+        {
+            if (!incomingEdge.getName().isPresent())
+            {
+                return false;
+            }
+            final String startEdgeName = startEdge.getName().get();
+            final String incomingEdgeName = incomingEdge.getName().get();
+            return checkBeforeEditDistance(incomingEdgeName, startEdgeName);
```
ranek: As a future improvement, I would be curious whether there are any cases where first doing Unicode normalization on the strings would change the result. For example, if there are cases where an accented character is decomposed in the name of one Way, and composed on another, such that the edit distance looks like two, when two equivalent strings would only have an edit distance of one from each other.

seancoulter: Interesting. I could see something like this being used if we wanted to focus on spelling inconsistencies related to character numbers between the two strings as opposed to accent mixup.

yiqingj: not completely relevant but as another future improvement if we want to handle more complex cases, we can also consider leveraging some libraries like Lucene. e.g. https://lucene.apache.org/core/7_3_1/suggest/org/apache/lucene/search/spell/LevenshteinDistance.html

config/configuration.json
```diff
@@ -268,6 +268,15 @@
       "difficulty": "HARD"
     }
   },
+  "RoadNameSpellingConsistencyCheck": {
+    "distance.search.maximum": "500.0",
```
yiqingj: this needs to be fixed.

seancoulter: Thanks. Good catch

PR #188, opened 2019-08-21
Atlas version bump

shawat94:
Bumping atlas version 5.7.1 -> 5.7.5

**Testing Approach:**
Built locally

**Test Results:**
Built successfully


ranek:
Looks good. Is there a reason to go to 5.7.4 instead of 5.7.5?

shawat94:
That was an oversight on my part, I referred to the original PR comment for the Atlas version rather than checking and testing the newer version. I've just built with 5.7.5 and updated the PR.

PR #193, opened 2019-09-26
[GeneralizedCoastlineCheck] Sharp angle enhancement

seancoulter:
### Description:

Edits are made to GeneralizedCoastlineCheck so that generalized coastlines with sharp angles are given separate instructions for editing. The goal here is to encourage focused editing on areas of a generalized coastline that most need it. Sharp angles in a generalized coastline often reflect poor digitization, and should be targeted for fixing. The enhancement this PR brings addresses that.

### Potential Impact:

No change in the number of flags generated by GeneralizedCoastlineCheck. Instructions and markers in a flag for a coastline are subject to change if that coastline has sharp angles.

### Unit Test Approach:

Unit tests

### Test Results:

Unit tests pass


seancoulter:
DNM while we look into making sharp angle logic switchable

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/GeneralizedCoastlineCheck.java
```diff
@@ -90,7 +92,8 @@ public boolean validCheckForObject(final AtlasObject object)
             // Contains midpoints and sharp angle locations
             final List<Location> pointsForFlagging = this.getPointsForFlagging((LineItem) object);
             // If there were sharp angles
-            if (pointsForFlagging.addAll(this.getSharpAngleLocations((LineItem) object)))
+            if (pointsForFlagging.addAll(this.getSharpAngleLocations((LineItem) object))
+                    && this.sharpAngleThreshold.asDegrees() > DO_NOT_EVALUATE_DEFAULT_CAP)
```
Bentleysb: Just a thought, I am not sure if this makes sense for this check or not. You could use `DO_NOT_EVALUATE_DEFAULT_CAP` as the default value for sharpAngleThreshold. That way if `angle.minimum.threshold` is not given in the config this functions is turned off, instead of needing to know that -1 needs to be given to switch it off. The disadvantage would be this feature is turned off by default.

seancoulter: Good idea. I like this better since there might've been some confusion around valid configuration values. Made the changes in 3608a46

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/GeneralizedCoastlineCheck.java
```diff
@@ -38,8 +38,8 @@
             SHARP_ANGLE_INSTRUCTIONS);
     private static final double MINIMUM_DISTANCE_BETWEEN_NODES = 100;
     private static final double MINIMUM_NODE_PAIR_THRESHOLD_PERCENTAGE = 30.0;
-    private static final double SHARP_ANGLE_THRESHOLD_DEFAULT = 83.0;
-    private static final double DO_NOT_EVALUATE_DEFAULT_CAP = -1;
+    private static final double SHARP_ANGLE_THRESHOLD_DEFAULT = Integer.MAX_VALUE;
+    private static final double DO_NOT_EVALUATE_DEFAULT_CAP = Integer.MAX_VALUE;
```
Bentleysb: Do you think we need to define 2 static variables with the same value? I can see the descriptive value, but it seems a bit strange still.

seancoulter: Yes, it was more for the descriptive value, but I made changes in c7549a8 which hopefully gives us the best of both worlds

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/GeneralizedCoastlineCheck.java
```diff
@@ -15,27 +17,35 @@
...
-            "This coastline is generalized, as {0}% of node pairs are {1} or more meters apart. To fix, increase the number of nodes along this coastline. The midpoint of generalized segments is dotted for convenience.");
+    private static final String BASIC_INSTRUCTIONS = "This coastline is generalized, as {0}% of node pairs are {1} or more apart. To fix, add more nodes to this coastline. The midpoints of generalized segments are dotted for convenience.";
+    private static final String SHARP_ANGLE_INSTRUCTIONS = "This coastline is generalized, as {0}% of node pairs are {1} or more apart. There are also sharp angles exceeding {2} degrees. To fix, add more nodes to smooth angles and break up long segments of coastline. Suggested areas to add nodes are dotted.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(BASIC_INSTRUCTIONS,
+            SHARP_ANGLE_INSTRUCTIONS);
     private static final double MINIMUM_DISTANCE_BETWEEN_NODES = 100;
     private static final double MINIMUM_NODE_PAIR_THRESHOLD_PERCENTAGE = 30.0;
+    private static final double SHARP_ANGLE_THRESHOLD_DEFAULT = Integer.MAX_VALUE;
```
danielduhh: Why Max value here?

seancoulter: It's supposed to be a value that shouldn't ever be used in the configuration. It could technically be anything over 359 degrees but I think this makes it clearest

PR #201, opened 2019-10-08
[FloatingEdgeCheck] No longer flag floating edges in airports

seancoulter:
### Description:

This PR changes FloatingEdgeCheck so that it no longer flags Edges that are within/overlap an airport area. Edge networks found in airports often do not follow the same navigation standards as other road networks and so they shouldn't be subject to the same Check logic.

### Potential Impact:

FloatingEdgeCheck should not flag floating edges that are in airports now.

### Unit Test Approach:

Unit tests with overlap on an airport taxiway, and containment within an airport

### Test Results:

Unit tests pass

### DIFF
--- a/docs/checks/floatingEdgeCheck.md
+++ b/docs/checks/floatingEdgeCheck.md
@@ -2,7 +2,7 @@
 
 #### Description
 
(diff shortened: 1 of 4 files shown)


danielduhh:
Do you have any idea of roughly how many flags we should expect to drop?

PR #202, opened 2019-10-11
RoundaboutConnectorCheck

Bentleysb:
### Description:

This check flags roads that connect to a roundabout at too sharp an angle. 

This is usually indicative of one way roads that have been drawn backwards, two way roads that should be one way, or other similar errors that cause routing/navigation issues. 

The check has two configurable values that set the threshold angles for flagging one and two way roads. Please see the readme for more on these values. 

**OSM Examples:**
1. The way [id:296416348](https://www.openstreetmap.org/way/296416348) is digitized backwards.
2. The way [id:606815435](https://www.openstreetmap.org/way/606815435) should be made one way.
3. The way [id:256474193](https://www.openstreetmap.org/way/256474193) is one way in the correct direction, but connects at a sharp angle due to poor digitization.

### Potential Impact:

New check, new flags

### Unit Test Approach:

Unit tests cover a variety of one and to way scenarios for both left and right hand drive countries. 

### Test Results:

This check was tested in 7 countries. 170 out of 449 flags were sampled. 13 of the sampled flags were found to be false positives. That gives this check a 7% false positive rate. The 13 false positives cases were generally caused by poor roundabout digitization or rare unconventional roundabout design.


docs/available_checks.md
```diff
@@ -18,6 +18,7 @@ This document is a list of tables with a description and link to documentation f
 | [FloatingEdgeCheck](checks/floatingEdgeCheck.md) | The purpose of this check is to identify Edges that are not accessible or navigable from the rest of the Edge network due to lack of connectivity or access restrictions. |
 | [MalformedRoundaboutCheck](checks/malformedRoundaboutCheck.md) | The purpose of this check is to identify roundabouts mapped in the opposite direction of traffic. The check takes into consideration countries with both left-side and right-side traffic. There are three types of map errors that will be flagged by this check: 1. Wrong-way-roundabouts, 2. Multi-directional roundabouts, and 3. Roundabouts with poor geometry. |
 | [~~RoundaboutClosedLoopCheck~~ (Deprecated)](checks/roundaboutClosedLoopCheck.md) | The purpose of this check is to identify Roundabout Edges that are bi-directional or have an end Node with less than 2 connections.  **This check has been deprecated and is no longer active.** |
+| [RoundaboutConnectorCheck](checks/roundaboutConnectorCheck.md) | The purpose of this check is to identify roads that connect to a roundabout at too sharp an angle |
```
danielduhh: 👍 !

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @author bbreithaupt
+ */
+public class RoundaboutConnectorCheck extends BaseCheck<Long>
+{
+    private static final long serialVersionUID = -5311314357995383430L;
+
+    private static final Double ONE_WAY_THRESHOLD_DEFAULT = 100.0;
+    private static final Double TWO_WAY_THRESHOLD_DEFAULT = 130.0;
+    private static final String ONE_WAY_INSTRUCTION = "This way, id:{0,number,#}, is connected to a roundabout at too sharp and angle.";
```
danielduhh: Is this the backwards digitization case? If so, it would be worth specifying that in the instructions

config/configuration.json
```diff
@@ -297,6 +297,19 @@
...
     }
   },
+  "RoundaboutConnectorCheck": {
+    "threshold": {
+      "one_way": 100.0,
+      "two_way": 130.0
+    },
+    "challenge": {
+      "description": "Roads connected to roundabouts at too sharp an angle.",
+      "blurb": "Sharp intersections between roundabouts and connecting roads are often digitization/navigation errors. Adjust these intersections to create smoother transitions in an out of roundabouts.",
```
sayas01: typo: "Adjust these intersections to create smoother transitions in ~an~ **and** out of roundabouts."

docs/available_checks.md
sayas01: 🎉

docs/checks/roundaboutConnectorCheck.md
```diff
@@ -0,0 +1,54 @@
+# Roundabout Connector Check
+
+This check flags roads that connect to a roundabout at too sharp an angle. 
+
+This is usually indicative of one way roads that have been drawn backwards, or two way roads that should be one way.   
+Other map errors found by this check are poorly digitized connector roads (poorly drawn but don't brake routing),
```
sayas01: typo: (poorly drawn but don't ~brake~ break routing)

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @author bbreithaupt
+ */
+public class RoundaboutConnectorCheck extends BaseCheck<Long>
+{
+    private static final long serialVersionUID = -5311314357995383430L;
+
+    private static final Double ONE_WAY_THRESHOLD_DEFAULT = 100.0;
+    private static final Double TWO_WAY_THRESHOLD_DEFAULT = 130.0;
+    private static final String ONE_WAY_INSTRUCTION = "This way, id:{0,number,#}, is connected to a roundabout at too sharp and angle. It may be digitized backwards";
```
sayas01: typo: is connected to a roundabout at too sharp ~and~ an angle. 
Same for the below instruction.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge && !this.isFlagged(object.getOsmIdentifier())
+                && !JunctionTag.isRoundabout(object)
+                && ((Edge) object).highwayTag().isMoreImportantThan(HighwayTag.SERVICE);
```
sayas01: Is there a possibility for the minimum highway type to change? Should we make this a configurable?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Get the first connected roundabout {@link Edge} from the in and out {@link Edge}s
+     *
+     * @param edge
+     *            the {@link Edge} to look for connected roundabouts from
+     * @return an {@link Optional} of a roundabout {@link Edge}
+     */
+    private Optional<Edge> getRoundaboutEdge(final Edge edge)
+    {
+        return Stream.concat(edge.inEdges().stream(), edge.outEdges().stream())
```
sayas01: Why don't we just use edge.connectedEdges() here instead of concatenating the in and out edges?

Bentleysb: Connected Edges can contain edges that route into the end node or out of the start node. These are not navigable connected, and thus we don't want to consider them.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
Bentleysb: Yeah, I guess we may want to limit this at some point. I initially did it like this because I was thinking we would not want to go below service, but we may want to restrict this to higher highway values for more targeted flags. I will change this to be configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutConnectorCheck.java
sayas01: Makes sense. Thanks!

PR #207, opened 2019-10-20
[Flag Database] Fix performance issue

danielduhh:
### Description:

This makes a couple structural updates to the FlagDatabaseSubCommand in order to fix a performance issue introduced in (#194). Adding a subquery to the feature `INSERT` statement is an expensive operation given there's over 13 million flagged features in the world.

We can no longer batch the flag records because we need to store and pass the primary key to the feature table. Given that, I put a counter in place to batch 1000 feature records at a time.

### Potential Impact:

I changed the name and method signature of the public sub command function from batchFlagStatement to executeFlagStatement. This forced me to update the unit tests as well

### Unit Test Approach:

Updated mock unit tests to verify the right jdbc function was getting called.

### Test Results:

Tested the performance on various checkflag sized outputs.

| size   | flags   | features | runtime (before) | runtime (after) |
|--------|---------|----------|------------------|-----------------|
| 6.3MB  | 24,147  | 54,142   | 2.5 min          | 23 seconds      |
| 13MB   | 73,513  | 131,230  | 29.707 min       | 45.550 seconds  |
| 52.3MB | 150,310 | 230,863  |    TODO              | 2.115 minutes   |

java args: `-Xms10240m -Xmx10240m`

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/database/FlagDatabaseSubCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/database/FlagDatabaseSubCommand.java
@@ -10,6 +10,7 @@ import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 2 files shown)


danielduhh:
Thank's yall!

PR #209, opened 2019-10-23
Add InvalidPiersCheck and Update Atlas version

sayas01:
### Description:

**Note: The Atlas dependency has been updated to 5.8.3 to use two new methods in the LayerTag and LevelTag class.**

This PR adds a new check to identify piers (OSM ways with man_made=Pier Tag) that: (1) have a linear geometry but _DO NOT_ have an area=Yes Tag, or (2) have a polygonal geometry but _DO NOT_ have an area=Yes Tag.  A valid pier must either have 
- a highway Tag or
- overlaps a highway at the same layer and/or same level  as the pier or
- is connected to a ferry route or
- is connected to or overlaps a building/amenity=Ferry_Terminal at the same layer and/or same level as the pier.  In addition, if a pier is polygonal, then it will also be considered a valid check candidate if it has building and/or amenity=Ferry_Terminal Tags. 

If a pier matched any of the above conditions, it is flagged to be fixed.
Example: This [pier](https://www.openstreetmap.org/way/117567537) has no area=yes tag has a connected ferry and overlapping highway.

### Potential Impact:

New check to identify invalid piers. Since the Atlas dependency version has been bumped to 5.8.3, all the updates to Atlas in 5.8.3 will be available to use.

### Unit Test Approach:

Added unit test and tested locally on a couple of countries.

### Test Results:

All tests passing and analysis of the outputs of the local run gave 0% FP.


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InvalidPiersCheck.java
```diff
@@ -0,0 +1,350 @@
...
+        final boolean intersectsBuilding;
+        final Atlas atlas = originalEdge.getAtlas();
+        // Checks if the pier is connected to a ferry route
+        intersectsFerryRoute = edgesOfOSMWay.stream()
+                .anyMatch(eachEdge -> eachEdge.connectedEdges().stream().filter(
+                        connectedEdge -> this.areOnSameLevelOrLayer(connectedEdge, originalEdge))
+                        .anyMatch(RouteTag::isFerry));
+        if (isPolygonal)
+        {
+            intersectsBuilding = edgesOfOSMWay.stream().anyMatch(eachEdge -> Iterables
```
Bentleysb: Couldn't we do this once for `polygon` instead of performing the check on each in Edge in `edgesOfOSMWay`.

dependencies.gradle
```diff
@@ -1,7 +1,7 @@
 project.ext.versions = [
     checkstyle: '8.18',
     jacoco: '0.8.3',
-    atlas: '5.8.0',
+    atlas: '5.8.3',
```
Bentleysb: Do we want to do this in this PR, or mark this as DNM and wait for a dependency bump?

sayas01: I added it as atlas 5.8.3 is a certified version; Since check isn't high priority, I can wait for the next release as well. @danielduhh have we ever updated dependencies like this or we normally wait for dependency bump ?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InvalidPiersCheck.java
sayas01: I had used polygon first like ```Iterables.stream(atlas.areasIntersecting(polygon,intersectingArea ->
                    this.areOnSameLevelOrLayer(intersectingArea, originalEdge)
                    && (IS_BUILDING.test(intersectingArea)
                    || IS_FERRY_TERMINAL.test(intersectingArea)))).iterator().hasNext()``` which gave me two FPs in a country. Adjacent areas where pulled in using the polygon. So I went for each edges.

Bentleysb: Interesting, I wounder if that has to do with polygonal vs liner intersection. It does seem like this is an expensive way to perform this calculation, as opposed to running the query for a single geometry. Do you think it would work to use linear intersection by querying/testing on a PolyLine made from all the way Edges?

Bentleysb: After some separate discussions it appears that there is a potential bug in area intersection. This seems like an ok solution while that is investigated further.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InvalidPiersCheck.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+        // Check if the pier has connections to ferry route or buildings
+        final boolean isConnectedToFerryOrBuilding = this.isConnectedToFerryOrBuilding(edge,
+                listOfEdgesFormingOSMWay, isPolygonal, osmWayAsPolygon);
+        // Check if the pier overlaps a highway or not
+        final boolean overlapsHighway = this.pierOverlapsHighway(edge, listOfEdgesFormingOSMWay,
+                osmWayAsPolygon, isPolygonal);
+        // Flag the pier if it overlaps a highway or is connected
+        // to a building or ferry route or overlaps a building
+        return overlapsHighway || isConnectedToFerryOrBuilding
```
danielduhh: Do you mind calling out "overlaps highway" and "is connected to very or building" in the instructions ?

sayas01: @danielduhh I had considered adding these in the instructions before but later on decided to drop it because we are flagging piers if
1) they have a highway tag or
2) is connected to ferry/building or
3) overlaps highway or
4) is polygonal pier with building tag or
5) is polygonal pier with amenity=ferry_terminal.
Essentially 5 different cases. There could be cases where two or more of these are true which would make the instructions very long. Also since we are trying to find linear piers and polygonal piers with no area tags, the specifics of which of the above conditions the pier meet, may be irrelevant to editors.

danielduhh: I think having 5 use cases is more reason to have specific instructions. With that, people could filter the flags by some string value in the instructions if they only cared about a single case. Thoughts?

sayas01: Sounds reasonable. I will update the instructions likewise.

sayas01:
@Bentleysb @danielduhh Could you review this again please? The latest commit includes updates to the flag instructions based on @danielduhh's comments.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InvalidPiersCheck.java
```diff
@@ -0,0 +1,362 @@
...
+        final Set<Location> locationsInOsmWay = new HashSet<>();
+        listOfEdgesFormingOSMWay
+                .forEach(osmWayEdge -> locationsInOsmWay.addAll(osmWayEdge.asPolyLine()));
+        final Polygon osmWayAsPolygon = new Polygon(locationsInOsmWay);
+        // Check if the OSM way has linear geometry or polygonal geometry
+        final boolean isPolygonal = this.hasPolygonalGeometry(listOfEdgesFormingOSMWay, edge);
+        final int instructionIndex;
+        // We can flag the edge if it has a highway tag with the right priority or is a polygonal
+        // pier with building tag or is a polygonal pier with amenity=ferry_terminal
+        if ((HighwayTag.highwayTag(edge).isPresent() && HighwayTag.highwayTag(edge).get()
```
danielduhh: Are there cases where a flag can contain multiple issues? Ideally, with each issue, we could concatenate the instructions and flag them all at the end

sayas01: Since checking for overlappingHighway and intersecting/overlapping buildings is computationally heavy tasks, I short circuit the check and return the flag if  the pier has a highway tag or is polygon with building tag/amenity=ferry terminal tag. If we want to include all the conditions in the instructions, then we need to check for intersections/overlaps even if it has a highway tag. Thoughts?

danielduhh: Ahh ic, I'm cool with that. We can revisit it later !

danielduhh:
Thanks @sayas01 !!

PR #210, opened 2019-10-25
[Flag Database] Create new schema if doesn't exist

danielduhh:
### Description:

When setting the `currentSchema` value in the connection uri to a *new* schema that didn't already exist, or have the postgis extension installed, the flad db command would throw an `NullPointerException` on `connection.getSchema()`.

This adds a `getSchema` function to the DatabaseConnection that parses the URI query parameters and return the value of `currentSchema`, or the default (`"public"`) if it's not provided.

### Potential Impact:

No downstream impact

### Unit Test Approach:

I added a few unit tests for fetching query parameter values from the connection URI.

### Test Results:

All tests pass as expected


src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java
```diff
@@ -29,6 +34,7 @@
     public DatabaseConnection(final String connectionUrl)
```
seancoulter: We could annotate ```connectionUrl``` with the @javax.validation.constraints.NotNull tag since it looks like we'd get unexpected behavior if ```connectionUrl``` is null. DatabaseConnection is a public class so we would want to avoid any misuse when creating objects

```public DatabaseConnection(@NotNull final String connectionUrl)```

**Comment by seancoulter** [comment]
LGTM, just had a suggestion for improving code documentation.

File: src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java (hunk 1 above)
Comment (danielduhh) [reply src/main/java/org/openstreetmap/atlas/checks/database/DatabaseConnection.java]: Passing in a null value for the connection url will throw a jdbc SQLException
```java
12:03:36.184 [main] DEBUG o.o.a.u.c.a.AbstractAtlasShellToolsCommand - Command using context 3
Oct 30, 2019 12:03:36 PM org.postgresql.Driver parseURL
WARNING: JDBC URL must contain a / at the end of the host or port: jdbc:postgresql://null
Oct 30, 2019 12:03:36 PM org.postgresql.Driver parseURL
WARNING: JDBC URL must contain a / at the end of the host or port: jdbc:postgresql://null
12:03:36.213 [main] ERROR o.o.a.c.d.FlagDatabaseSubCommand - Invalid connection string. host[:port]/database
java.sql.SQLException: No suitable driver found for jdbc:postgresql://null
	at java.sql.DriverManager.getConnection(DriverManager.java:689) ~[na:1.8.0_161]
	at java.sql.DriverManager.getConnection(DriverManager.java:270) ~[na:1.8.0_161]
```

PR #212, opened 2019-10-30
FlagDatabaseSubCommand: Add Meta Data Arguments

Bentleysb:
### Description:

This adds two new arguments to the FlagDatabaseSubCommand: `run_uri` and `software_version`. Both are optional arguments used to pass meta data about the generation of the flags that are being uploaded. 

This also changes the type of the software_version column to remove an arbitrary character limit. 

### Potential Impact:

None, the arguments are optional. 

### Unit Test Approach:

Updated two unit tests so that the arguments can be retrieved from the `executeFlagStatement` method. 

### Test Results:

Tested uploading data to a local database using the new arguments. 

Both before the update and when not passing the arguments the flag table looked like this:
Screen Shot 2019-10-30 at 9 18 42 AM

When the arguments were provided (tested using long and short form argument names) the flag table looked like this:
Screen Shot 2019-10-30 at 11 39 45 AM

### DIFF
--- a/src/main/resources/org/openstreetmap/atlas/checks/database/schema.sql
+++ b/src/main/resources/org/openstreetmap/atlas/checks/database/schema.sql
@@ -8,7 +8,7 @@ CREATE TABLE IF NOT EXISTS flag (
 	check_name text not null,
 	instructions text not null,
 	run_uri text,
-	software_version varchar(8),
+	software_version text,
 	date_created timestamp
 );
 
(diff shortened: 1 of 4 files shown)


src/main/resources/org/openstreetmap/atlas/checks/database/schema.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE IF NOT EXISTS flag (
 	check_name text not null,
 	instructions text not null,
 	run_uri text,
-	software_version varchar(8),
+	software_version text,
```
danielduhh: 👍

PR #216, opened 2019-11-11
[WIP]Atlas-checks global edit time filter

seancoulter:
### Description:

An optional, global tag-based time filter is injected into every check in the project. This is done at check instantiation time via BaseCheck. The idea here is to filter atlas objects (in BaseCheck.checkObjectFilter) for ones that have a last_edit_time tag within a specified time range. This time range is defined by min/max configurable in the project configuration file.

Note: the time range is defined under the ```_TimeFilter``` construct in the configuration file. During check loading time, this should not be read since it doesn't exist in the ```org.openstreetmap.atlas.checks.validation``` package. Another location for the filters could be in ```CheckResourceLoader```. Or if there's another place that's better, I'm happy to move it.

### Potential Impact:

Filters are optional. No impact when off, and the chance of fewer flags when they're on.

### Unit Test Approach:
Incoming

### Test Results:
Incoming

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -6,6 +6,10 @@
     "type": "org.openstreetmap.atlas.checks.base.BaseCheck",
     "enabled.value.default": true
   },
+  "_TimeFilter": {
+    "min": "2010-01-01",
+    "max": "2015-07-29"
+  },
   "PoolSizeCheck": {
     "surface": {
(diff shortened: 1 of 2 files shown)


seancoulter:
Closing as we explore using other methods to filter atlas objects by time

PR #220, opened 2019-11-19
[MapRouletteUploadCommand] Add generating check metadata to Task Geojson

seancoulter:
### Description:

This PR's purpose is primarily to add the Check and framework generator (atlas checks) which generated a Task to that Task's geojson when uploading to MapRoulette. This is particularly useful for cases where custom challenge names are used in the configuration; here the original name of the check was previously not necessarily associated with the Task. Now we ensure that it is.

There was also a minor bug with the custom challenge name which has been fixed. When the ISO country code was prepended to the challenge name, it would overwrite the custom challenge name in favor of the generating check name. This has now been fixed so the ISO code is prepended to the custom challenge name as expected.

### Potential Impact:

The generating check name now exists in a Task's geojson.

### Unit Test Approach:

Same unit tests. I added a sample geojson variable to the sample task used for test MR uploads (in MapRouletteClientTest.java). This was to increase test coverage for our existing unit tests, which previously did not generate MR-ready tasks (JsonObjects) for upload. With these additions the unit tests fail if the Task was not generated properly. This can happen if there was an exception when parsing the Task object to create an MR-ready task, mostly in Task.generateTask()

### Test Results:

Tests pass. 

I also did a sample project upload to Map Roulette. In that project is a single challenge, and if I choose a random task in that challenge:
image

We can see the "flag:check" and "flag:generator" fields in the Properties object.

In that task's geojson we can see the generating check name/framework in the properties object:
image


src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
```diff
@@ -120,13 +120,13 @@ protected void execute(final CommandMap commandMap,
                         {
                             try
                             {
-                                final Challenge challenge = this
-                                        .getChallenge(task.getChallengeName(), instructions);
+                                final Challenge challenge = this.getChallenge(check, instructions);
+                                final Challenge mockChallenge = new Challenge(challenge);
```
Bentleysb: Could this be combined with the line above, or would that break where you call challenge.getName() below?

danielduhh:
How do people feel about adding this to each CheckFlag feature, as opposed to only adding the `generator` property when we run this command? 
@sayas01 @seancoulter @Bentleysb

src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
seancoulter: Yeah that could be done, but multiple variable initialization actually throws Sonar error S1659 (declare variables on separate lines)

seancoulter:
@danielduhh made some changes to the PR. Now the generating check is associated with each Task's through the Task's properties, so it renders in MapRoulette (see description). Since the generating check can already be found in the log file for the run, I don't think we need to add it in log file creation time -- it doesn't get read in to a Task by default so that's why I injected it here.

src/main/java/org/openstreetmap/atlas/checks/maproulette/data/Task.java
```diff
@@ -58,6 +58,7 @@ public Location getLocation()
     protected static final String TASK_PARENT_ID = "parent";
     protected static final String TASK_TYPE = "type";
     private static final String KEY_DESCRIPTION = "description";
+    private static final String CHECK_GENERATOR = "check";
```
danielduhh: I believe this value will be overwritten if we have a feature with [the tag "check"](https://taginfo.openstreetmap.org/keys/check).  I suggest going with `flag:check`, [like the tippecanoe output](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java#L348) or `metadata`: {check: ""}, or something similar

danielduhh: We could also add: `flag:generator`: "Atlas Checks"

seancoulter: Done!

seancoulter: Now ```flag:check``` and ```flag:generator``` are in the properties field for a Task

PR #221, opened 2019-11-27
Sonar fixes

smaheshwaram:
### Description:

Sonar reported on blocker, critical and major issues. This PR is created to address blocker, critical and few major issues.

**Issues addressed in this PR:**
1. src/.../atlas/checks/atlas/predicates/TagPredicates.java
Move constants to a class or enum.

2. src/.../atlas/checks/atlas/predicates/TypePredicates.java
Move constants to a class or enum.

3. src/.../atlas/checks/atlas/predicates/ValuePredicates.java
Move constants to a class or enum.

4. src/.../atlas/checks/base/BaseCheck.java
Make "flaggedIdentifiers" transient or serializable.

5. src/.../atlas/checks/base/CheckResourceLoader.java
No need to call "toString()" method as formatting and string conversion is done by the Formatter.

6. src/.../atlas/checks/commands/AtlasChecksCommand.java
Catch Exception instead of Throwable.

7. src/.../atlas/checks/constants/CommonConstants.java
Move constants to a class or enum.

8. Rename field "EQUALS" to prevent any misunderstanding/clash with method "equals" defined in superclass "java.lang.Object".

9. src/.../atlas/checks/distributed/AtlasDataSource.java
Make "logger" transient or serializable.

10. src/.../checks/validation/linear/edges/SnakeRoadCheck.java
Remove this useless assignment; "walk" already holds the assigned value along all execution paths.

11. src/.../checks/validation/linear/edges/FloatingEdgeCheck.java
This block of commented-out lines of code should be removed.See Rule

12. src/.../checks/validation/linear/edges/FloatingEdgeCheck.java
This block of commented-out lines of code should be removed.

13. Add the missing @deprecated Javadoc tag.See Rule
Do not forget to remove this deprecated code someday.See Rule

14. String contains no format specifiers.

15. src/.../atlas/checks/configuration/ConfigurationResolver.java
Catch Exception instead of Throwable.

### Potential Impact:

1. No code smell, no failing checks on the PR.
2. Achieves clean code, chances of reducing potential bugs.
3. Refactoring on the code for more readability and maintainability.

### Unit Test Approach:

1. Since sonar issues were reported on various files, added coverage for new lines and modified lines.
2. Added new test class for ValuePredicates.java since the method used to test valid parenthesis is been completely replaced for clean code, fix potential bugs and sonar code smell.
3. No changes in the method names and method usages have been verified before making changes to the lines.

### Test Results:

1. Able to build the Atlas-checks project locally after PR changes.
2. Atlas-check configuration is run for different check filters.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -21,9 +21,9 @@ public final class ValuePredicates
      */
     public static boolean isValidParenthesis(final String s)
     {
-        char[] chars = s.toCharArray();
-        Deque<Character> stack = new ArrayDeque<>();
-        for (char c : chars)
+        final char[] chars = s.toCharArray();
+        final Deque<Character> stack = new ArrayDeque<>();
+        for (final char c : chars)
         {
             if ((c == ')' && (stack.isEmpty() || stack.pop() != '('))
                     || (c == ']' && (stack.isEmpty() || stack.pop() != '['))
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -4,7 +4,6 @@ import static org.openstreetmap.atlas.checks.constants.CommonConstants.CLOSED_PA
 import static org.openstreetmap.atlas.checks.constants.CommonConstants.OPEN_PARENTHESES_CHAR;
 
 import java.util.Stack;
-import java.util.function.Predicate;
 
 /**
  * Collection of value based predicates
@@ -23,8 +22,12 @@ public final class ValuePredicates
      * Tests if the {@link String} contains '(' and ')' that are balanced and closed. Balanced
      * meaning there are the same number of open parentheses as closed. Closed meaning that every
      * open parentheses is followed by a closed parentheses.
+     * 
+     * @param value
+     *            passed as an argument to check if valid parenthesis
+     * @return true if passed argument is valid parenthesis
      */
-    public static final Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
+    public static boolean isValidParenthesis(final String value)
     {
         final Stack<Character> characterStack = new Stack<>();
 
@@ -58,5 +61,5 @@ public final class ValuePredicates
             }
         }
         return characterStack.isEmpty();
-    };
+    }
 }
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -12,12 +12,6 @@ import java.util.Stack;
  */
 public final class ValuePredicates
 {
-    private ValuePredicates()
-    {
-        // default constructor to fix the error "Utility classes do have public or default
-        // constructor"
-    }
-
     /**
      * Tests if the {@link String} contains '(' and ')' that are balanced and closed. Balanced
      * meaning there are the same number of open parentheses as closed. Closed meaning that every
@@ -61,5 +55,11 @@ public final class ValuePredicates
             }
         }
         return characterStack.isEmpty();
+    }
+    
+    private ValuePredicates()
+    {
+        // default constructor to fix the error "Utility classes do have public or default
+        // constructor"
     }
 }
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -11,14 +11,20 @@ import java.util.function.Predicate;
  *
  * @author brian_l_davis
  */
-public interface ValuePredicates
+public final class ValuePredicates
 {
+    private ValuePredicates()
+    {
+        // default constructor to fix the error "Utility classes do have public or default
+        // constructor"
+    }
+
     /**
      * Tests if the {@link String} contains '(' and ')' that are balanced and closed. Balanced
      * meaning there are the same number of open parentheses as closed. Closed meaning that every
      * open parentheses is followed by a closed parentheses.
      */
-    Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
+    public static final Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
     {
         final Stack<Character> characterStack = new Stack<>();
 
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -56,7 +56,7 @@ public final class ValuePredicates
         }
         return characterStack.isEmpty();
     }
-    
+
     private ValuePredicates()
     {
         // default constructor to fix the error "Utility classes do have public or default
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
@@ -2,6 +2,7 @@ package org.openstreetmap.atlas.checks.atlas.predicates;
 
 import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.function.Predicate;
 
 /**
  * Collection of value based predicates
@@ -14,12 +15,8 @@ public final class ValuePredicates
      * Tests if the {@link String} contains '(' and ')' that are balanced and closed. Balanced
      * meaning there are the same number of open parentheses as closed. Closed meaning that every
      * open parentheses is followed by a closed parentheses.
-     * 
-     * @param value
-     *            passed as an argument to check if valid parenthesis
-     * @return true if passed argument is valid parenthesis
      */
-    public static boolean isValidParenthesis(final String value)
+    private static Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
     {
         final char[] chars = value.toCharArray();
         final Deque<Character> stack = new ArrayDeque<>();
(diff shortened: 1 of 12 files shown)


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -282,7 +288,7 @@ protected boolean acceptPier()
 
     protected void clearFlaggedIdentifiers()
     {
-        this.flaggedIdentifiers.clear();
+        getFlaggedIdentifiers().clear();
```
mgcuthbert: You should still have "this." prefixed to the method.

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -60,7 +60,7 @@
     private final List<String> countries;
     private final Map<String, List<String>> flagLanguageMap;
     // OSM Identifiers are used to keep track of flagged features
-    private final Set<T> flaggedIdentifiers = ConcurrentHashMap.newKeySet();
+    private transient Set<T> flaggedIdentifiers = ConcurrentHashMap.newKeySet();
```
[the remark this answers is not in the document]
mgcuthbert: Because you are using lazy initialization, I would suggest just setting this initially to null. It will kind of force people to use the method even if they are not thinking about serialization.

src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
```diff
@@ -38,7 +38,7 @@
 {
     private static final long serialVersionUID = -6407331424906155431L;
     private Atlas atlas = null;
-    private final Logger logger = LoggerFactory.getLogger(AtlasDataSource.class);
+    private final transient Logger logger = LoggerFactory.getLogger(AtlasDataSource.class);
```
mgcuthbert: This logger variable will deserialize to null and could cause issues.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
```diff
@@ -1,56 +1,47 @@
...
+     * @return true if passed argument is valid parenthesis
      */
-    Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
+    public static boolean isValidParenthesis(final String value)
     {
-        final Stack<Character> characterStack = new Stack<>();
-
-        for (final char character : value.toCharArray())
+        final char[] chars = value.toCharArray();
+        final Deque<Character> stack = new ArrayDeque<>();
```
seancoulter: Just curious, why the switch from Stack to Deque here?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
```diff
@@ -97,10 +97,10 @@ public boolean validCheckForObject(final AtlasObject object)
         this.markAsFlagged(object.getOsmIdentifier());
 
         // Instantiate the network walk with the starting edge
-        SnakeRoadNetworkWalk walk = initializeNetworkWalk(edge);
+        final SnakeRoadNetworkWalk walk = initializeNetworkWalk(edge);
 
         // Walk the road
-        walk = walkNetwork(edge, walk);
+        walkNetwork(edge, walk);
```
seancoulter: It looks like we don't do anything with the return value - maybe the function can be void

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
smaheshwaram: Sonar is complaining about it as a major issue because Stack is synchronized class and Deque is unsynchronized.

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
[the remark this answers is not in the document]
smaheshwaram: I agree. Assigning null to the variable.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
smaheshwaram: I will modify it to void.

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
smaheshwaram: Inside getFlaggedIdentifiers(), "this" keyword is used to set and return the instance variable "flaggedIdentifiers". So, even though different instances are accessing this BaseCheck class, getFlaggedIdentifiers() sets the value to its own instance variable. Please let me know if I am incorrect.

src/main/java/org/openstreetmap/atlas/checks/distributed/AtlasDataSource.java
smaheshwaram: I will assign null to it and add a getter method to set and get the LoggerFactory.getLogger(AtlasDataSource.class).

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -96,8 +95,8 @@
             "Saves intermediate atlas files created when processing OSM protobuf data.",
             Boolean::valueOf, Optionality.OPTIONAL, "false");
     private static final Switch<Set<OutputFormats>> OUTPUT_FORMATS = new Switch<>("outputFormats",
-            String.format(
-                    "Comma-separated list of output formats (flags, metrics, geojson, tippecanoe)."),
+            String.format("Comma-separated list of output formats %s and %s and %s and %s ",
+                    "flags", "metrics", "geoJson", "tippeCanoe"),
```
Bentleysb: I am guessing spectrum was complaining about using String.format without only one argument. All of these arguments are just string literals, so we can just use the original string without calling String.format.

smaheshwaram:
PR comments have been addressed. If no other suggestions or changes required, then the PR can be merged.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
```diff
@@ -1,56 +1,47 @@
 package org.openstreetmap.atlas.checks.atlas.predicates;
 
-import static org.openstreetmap.atlas.checks.constants.CommonConstants.CLOSED_PARENTHESES_CHAR;
```
danielduhh: Was sonar complaining about using these? If not, we can still leverage these characters

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
```diff
@@ -1,56 +1,47 @@
...
      * Tests if the {@link String} contains '(' and ')' that are balanced and closed. Balanced
      * meaning there are the same number of open parentheses as closed. Closed meaning that every
      * open parentheses is followed by a closed parentheses.
+     * 
+     * @param value
+     *            passed as an argument to check if valid parenthesis
+     * @return true if passed argument is valid parenthesis
      */
-    Predicate<String> ARE_PARENTHESES_BALANCED_AND_CLOSED = value ->
+    public static boolean isValidParenthesis(final String value)
```
danielduhh: What's the reason for changing from Predicate to boolean ?

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -175,9 +173,10 @@ private static void executeChecks(final String country, final Atlas atlas,
     private static Iterable<ComplexEntity> findComplexEntities(final BaseCheck check,
             final Atlas atlas)
     {
-        if (check.finder().isPresent())
+        final Optional<Finder> val = check.finder();
```
danielduhh: Please avoid using acronyms as variable names 🙏 
`val-> value` (or something more descriptive)

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
smaheshwaram: Sonar was complaining about the usage of Stack and also a line inside the logic. So, I modified the logic.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/ValuePredicates.java
smaheshwaram: Good catch. I modified it when changing the interface to class and forgot to keep Predicate back. Will add it.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
smaheshwaram: Sure.

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
seancoulter: I think he meant it should read ```this.getFlaggedIdentifiers()``` since that method is non static and belongs to the instance of the class.

src/main/java/org/openstreetmap/atlas/checks/configuration/ConfigurationResolver.java
```diff
@@ -161,9 +161,9 @@ public static Configuration loadConfiguration(final CommandMap commandMap,
                 {
                     source.close();
                 }
-                catch (final Throwable throwable)
+                catch (final IOException ioe)
```
seancoulter: ioe -> exception (or similar)

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
mgcuthbert: Yes you are correct, but in this particular case it is more about readability. "this" keyword is simply referencing the current object, and "this.getFlaggedIdentifiers()" is exactly the same as "getFlaggedIdentifiers()", it is really just for readability. Generally speaking if you are referencing a method or variable in a class you should prefix it with "this" for readability.

PR #224, opened 2019-12-06
InvalidTagsCheck

Bentleysb:
### Description:

This is a generic check for flagging features using TaggableFilters. 

The check uses a configurable list of filters. Each filter has 2 parts. The first is a list of AtlasEntity classes (node, edge, area, etc.). The second is a [TaggableFilter](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/filters/TaggableFilter.java). If a feature is one of the classes given and passes the TaggableFilter then it is flagged.
The TaggableFilter is also used to generate the instructions for a flag. For each TaggableFilter that an object passes an instruction is added to the flag. The instruction uses the tag keys from the TaggableFilter to give guidance on what the problem is. 
There is currently only one filter for this check. It is described in the filter example below.

**Filter Example:**

Areas and Relations with the tag `boundary=protected_area` should have a `protect_class` tag.  
A filter to flag this would look like the following:  
`["area,relation","boundary->protected_area&protect_class->!"]`  
The first string is a list of the 2 AtlasEntity classes we want to look for. The second string is the TaggableFilter that is looking for the combination of `boundary=protected_area` without a `protect_class` tag.

This would flag an osm feature like the following: [Way 673787307](https://www.openstreetmap.org/way/673787307).

The instruction for flags produced by this filter would look like: `Check the following tags for missing, conflicting, or incorrect values: boundary, protect_class`

### Potential Impact:

This check is similar to ConflictingTagCombinationCheck and ConflictingAreaTagCombination, but more generic. It is possible that those checks could at a later date be absorbed into this check. 

### Unit Test Approach:

Unit test check a number of combinations for each part of the configurable filter system. Multiple AtlasObject classes and TaggableFilters are tested. There is also a test for the dynamic instructions. 

### Test Results:

Tested in 7 countries and sampled 129 flags. All flags were true positive cases.


src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Gathers the keys from a {@link TaggableFilter} using regex.
+     *
+     * @param filter
+     *            a {@link TaggableFilter}
+     * @return the tag keys as a {@link Set} of {@link String}s
+     */
+    private static Set<String> getFilterKeys(final TaggableFilter filter)
+    {
+        return Arrays.stream(filter.toString().split("[|&]+"))
```
smaheshwaram: Regex can be made as a static final variable string.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
```diff
@@ -0,0 +1,159 @@
...
+                .filter(string -> string.contains(KEY_VALUE_SEPARATOR))
+                .map(tag -> tag.split(KEY_VALUE_SEPARATOR)[0]).collect(Collectors.toSet());
+    }
+
+    /**
+     * Convert a {@link String} of comma delimited
+     *
+     * @param classString
```
smaheshwaram: Description for params is missing.

PR #227, opened 2019-12-12
[SinkIslandCheck] sink islands surrounded by pedestrian ways

seancoulter:
### Description:

This PR does the following:

--Adds a switch to allow for candidate sink island edges (highway=service+ edges with valid accessibility tags) WITHOUT a motor_vehicle tag to be flagged if those candidate edges are surrounded by pedestrian ways.
--Ignores such candidate sink islands with motor_vehicle=yes tags that are surrounded by pedestrian ways.

### Potential Impact:
With the switch on, we should expect to see more flags as a result of sink islands surrounded by pedestrian ways being flagged.
Regardless of the switch, we should anticipate some diffs to account for the fact that we now ignore sink islands with motor_vehicle=yes tags that are surrounded by pedestrian ways.

### Unit Test Approach:

Added some new unit tests.
Pending more analysis

### Test Results:

Tests pass


config/configuration.json
```diff
@@ -799,6 +799,7 @@
   "SinkIslandCheck": {
     "tree.size": 50,
     "minimum.highway.type": "service",
+    "filter.pedestrian.network": true,
```
sayas01: Is the default config going to be true for "filter.pedestrian.network"?

seancoulter: It should be false to maintain baseline outputs. Thanks

seancoulter:
@danielduhh updated description with flag count expectations

PR #228, opened 2019-12-12
Road name gap check

smaheshwaram:
### Description:

This check identifies Edges with no name Tag that are between two other Edges with the same name Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.

i) This check should flag the edges who has NO name tag with below conditions:
   1. Is an Edge
   2. Has the Tag,
      a. highway=TERTIARY, SECONDARY, PRIMARY, TRUNK, or MOTORWAY
   3. Does NOT have the Tag,
      a. name=*
      b. highway=*_LINK
         1. TERTIARY_LINK, SECONDARY_LINK, PRIMARY_LINK, TRUNK_LINK, or MOTORWAY_LINK
      c. junction=ROUNDABOUT
   
ii) This check should flag the edges which has different Name Tag:
    1. Is an Edge
    2. Has the Tag,
        1. highway=TERTIARY, SECONDARY, PRIMARY, TRUNK, or MOTORWAY.
    3. name=* but is not the same as the other Edges it is connected to
    4. Does NOT have the Tag,
            a. highway=*_LINK
            b. junction=ROUNDABOUT

### Potential Impact:

A new check will be added to Atlas-checks project to identify and flag the edges with above mentioned description.

### Unit Test Approach:

Unit tests are added for the new check. Classes including RoadNameGapCheckTest.java and also a Ruler class i.e., RoadNameGapCheckTestRule to set up Atlas data.

### Test Results:

The code has been tested thoroughly to address the above mentioned description. Tested all the use cases. Tested all validations mentioned above in the description by taking 30MB data atlas files.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -373,6 +373,18 @@
       "difficulty":"NORMAL"
     }
   },
+  "InvalidTagsCheck": {
+    "filters.classes.tags": [
+      ["area,relation","boundary->protected_area&protect_class->!"]
+    ],
+    "challenge": {
+      "description": "Tasks containing features with tags containing missing, conflicting, or incorrect values",
+      "blurb": "Features with invalid tags",
+      "instruction": "Open your favorite editor and check that the listed tags are correct.",
+      "difficulty": "Medium",
+      "tags":"tags"
+    }
+  },
   "InvalidTurnRestrictionCheck": {
     "challenge": {
       "description": "Tasks containing invalid turn restrictions",
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -445,6 +445,7 @@
     }
   },
   "RoadNameGapCheck": {
+    "valid.highway.tag": ["primary","secondary","tertiary","trunk","motorway"],
     "challenge": {
       "description": "identifies Edges with no name Tag that are between two other Edges with the same name Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.",
       "blurb": "Flag the Edge with different name tag or no name tag which is between the same name tag edges",
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -444,6 +444,23 @@
       "difficulty": "HARD"
     }
   },
+  "RoadNameGapCheck": {
+    "tag.filters":[
+      "highway=TERTIARY",
+      "highway=SECONDARY",
+      "highway=PRIMARY",
+      "highway=TRUNK",
+      "highway=MOTORWAY",
+      "highway=!*_LINK"
+    ],
+    "challenge": {
+      "description": "identifies Edges with no name Tag that are between two other Edges with the same name Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.",
+      "blurb": "Flag the Edge with different name tag or no name tag which is between the same name tag edges",
+      "instruction": "Open your favorite editor and edit or add the name tag of the flagged edge(s) so they're consistent with connecting edges with same name tag.",
+      "difficulty": "EASY",
+      "tags":"highway"
+    }
+  },
   "RoadNameSpellingConsistencyCheck": {
     "distance.search.maximum": 500.0,
     "challenge": {
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -445,14 +445,6 @@
     }
   },
   "RoadNameGapCheck": {
-    "tag.filters":[
-      "highway=TERTIARY",
-      "highway=SECONDARY",
-      "highway=PRIMARY",
-      "highway=TRUNK",
-      "highway=MOTORWAY",
-      "highway=!*_LINK"
-    ],
     "challenge": {
       "description": "identifies Edges with no name Tag that are between two other Edges with the same name Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.",
       "blurb": "Flag the Edge with different name tag or no name tag which is between the same name tag edges",
--- a/docs/available_checks.md
+++ b/docs/available_checks.md
@@ -10,6 +10,7 @@ This document is a list of tables with a description and link to documentation f
 | [PedestrianAreaOverlappingEdgeCheck](checks/pedestrianAreaOverlappingEdgeCheck.md) | The purpose of this check is to identify pedestrian areas overlapping with roads that are not snapped to car navigable edges. |
 | [ShadowDetectionCheck](checks/shadowDetectionCheck.md) | The purpose of this check is to identify floating buildings. |
 | [SpikyBuildingCheck](checks/spikyBuildingCheck.md) | The purpose of this check is to identify buildings with extremely sharp angles in their geometry. |
+| [WaterbodyAndIslandSizeCheck](checks/waterbodyAndIslandSizeCheck.md) | The purpose of this check is to identify waterbodies and islands which are either too small or too large in size. |
 
 ## Highways
 | Check Name | Check Description |
@@ -69,4 +70,3 @@ This document is a list of tables with a description and link to documentation f
 | [GeneralizedCoastlineCheck](checks/generalizedCoastlineCheck.md) | The purpose of this check is to identify coastlines whose nodes are too far apart, have angles that are too sharp, and/or have _source=PGS_ Tag values. |
 | [IntersectingBuildingsCheck](checks/intersectingBuildingsCheck.md) | The purpose of this check is to identify buildings that intersect other buildings. |
 | [SelfIntersectingPolylineCheck](checks/selfIntersectingPolylineCheck.md) | The purpose of this check is to identify all edges/lines/areas in Atlas that have self-intersecting polylines, or geometries that intersects itself in some form. |
-| [WaterbodyAndIslandSizeCheck](checks/waterbodyAndIslandSizeCheck.md) | The purpose of this check is to identify waterbodies and islands which are either too small or too large in size. |
--- a/docs/available_checks.md
+++ b/docs/available_checks.md
@@ -56,6 +56,7 @@ This document is a list of tables with a description and link to documentation f
 | [InvalidAccessTagCheck](checks/invalidAccessTagCheck.md) | The purpose of this check is to identify invalid access tags. |
 | [InvalidLanesTagCheck](docs/checks/invalidLanesTagCheck.md) | The purpose of this check is to identify highways in OSM with an invalid lanes tag value. |
 | [MixedCaseNameCheck](checks/mixedCaseNameCheck.md) | The purpose of this check is to identify names that contain invalid mixed cases so that they can be edited to be the standard format. |
+| [RoadNameGapCheck](checks/RoadNameGapCheck.md) | The purpose of this check is to identify edge connected between two edges whose name tag is same. Flag the edge if the edge has a name tag different to name tag of edges connected to it or if there is no name tag itself.
 | [RoadNameSpellingConsistencyCheck](checks/RoadNameSpellingConsistencyCheck.md) | The purpose of this check is to identify road segments that have a name Tag with a different spelling from that of other segments of the same road. This check is primarily meant to catch small errors in spelling, such as a missing letter, letter accent mixups, or capitalization errors. |
 | [StreetNameIntegersOnlyCheck](checks/streetNameIntegersOnlyCheck.md) | The purpose of this check is to identify streets whose names contain integers only. |
 | [UnusualLayerTagsCheck](checks/unusualLayerTagsCheck.md) | The purpose of this check is to identify layer tag values when accompanied by invalid tunnel and bridge tags. |
--- a/docs/checks/RoadNameGapCheck.md
+++ b/docs/checks/RoadNameGapCheck.md
@@ -1,0 +1,37 @@
+# Road name gap check
+
+This is a new check that's designed to flag edges whose name tag is either different to the edges connecting to it with same name tag 
+or if there is no name tag itself for the edge which is connected between two edges with same name tag.
+The goal is to flag the above mentioned edges so that editors can either add the similar tag name to it or correct the existing tag name.
+
+i) This check should flag the edges who has NO name tag with below conditions:
+   1. Is an Edge
+   2. Has the Tag,
+      a. highway=TERTIARY, SECONDARY, PRIMARY, TRUNK, or MOTORWAY
+   3. Does NOT have the Tag,
+      a. name=*
+      b. highway=*_LINK
+         1. TERTIARY_LINK, SECONDARY_LINK, PRIMARY_LINK, TRUNK_LINK, or MOTORWAY_LINK
+      c. junction=ROUNDABOUT
+   
+ii) This check should flag the edges which has different Name Tag:
+    1. Is an Edge
+    2. Has the Tag,
+        1. highway=TERTIARY, SECONDARY, PRIMARY, TRUNK, or MOTORWAY.
+    3. name=* but is not the same as the other Edges it is connected to
+    4. Does NOT have the Tag,
+            a. highway=*_LINK
+            b. junction=ROUNDABOUT
+    
+### Live Examples
+
+Case 1: No name tag between two Edges with name tag (HKG)*
+a. Way 548524582 (https://www.openstreetmap.org/way/548524582) does not have the Tag, name=* and is 
+connected to Way 447864651 (https://www.openstreetmap.org/way/447864651) and Way 528677149 (https://www.openstreetmap.org/way/528677149) 
+that have the Tag, name=連翔道 Lin Cheung Road https://www.openstreetmap.org/way/548524582
+
+Case 2: Different name tag between two Edges with same name tag (ISO)
+ 
+### Code Review
+
+For the source code for this check, please refer to [RoadNameGapCheck](../../src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java)
--- a/docs/checks/RoadNameGapCheck.md
+++ b/docs/checks/RoadNameGapCheck.md
@@ -29,6 +29,9 @@ Case 1: No name tag between two Edges with name tag (HKG)*
 a. Way 548524582 (https://www.openstreetmap.org/way/548524582) does not have the Tag, name=* and is 
 connected to Way 447864651 (https://www.openstreetmap.org/way/447864651) and Way 528677149 (https://www.openstreetmap.org/way/528677149) 
 that have the Tag, name=連翔道 Lin Cheung Road https://www.openstreetmap.org/way/548524582
+b. Edge name: Tai. Incoming edge names: Tai , Shai. Outgoing edge name: Shai.
+c. Edge name: Tai. Incoming edge names: Shai. Outgoing edge name: Tai, Shai.
+d.: Edge name: Tai. Incoming edges names: Tai, Shai. Outgoing edge names: Shai, Pendler.
 
 Case 2: Different name tag between two Edges with same name tag (ISO)
  
--- a/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
@@ -55,25 +55,31 @@ public final class TagPredicates
             object, HighwayTag.class) && Validators.isOfType(object, AreaTag.class, AreaTag.YES);
 
     /**
-     * Tests if the {@link AtlasObject} has a highway tag that do not contain TERTIARY_LINK, SECONDARY_LINK,
-     * PRIMARY_LINK, TRUNK_LINK, or MOTORWAY_LINK
+     * Tests if the {@link AtlasObject} has a highway tag that do not contain TERTIARY_LINK,
+     * SECONDARY_LINK, PRIMARY_LINK, TRUNK_LINK, or MOTORWAY_LINK
      */
-    public static final Predicate<AtlasObject> IS_HIGHWAY_NOT_LINK_TYPE = object -> Validators.hasValuesFor(object,
-            HighwayTag.class) && Validators.isNotOfType(object, HighwayTag.class, HighwayTag.TERTIARY_LINK,
-            HighwayTag.SECONDARY_LINK, HighwayTag.PRIMARY_LINK, HighwayTag.TRUNK_LINK, HighwayTag.MOTORWAY_LINK);
-    
+    public static final Predicate<AtlasObject> IS_HIGHWAY_NOT_LINK_TYPE = object -> Validators
+            .hasValuesFor(object, HighwayTag.class)
+            && Validators.isNotOfType(object, HighwayTag.class, HighwayTag.TERTIARY_LINK,
+                    HighwayTag.SECONDARY_LINK, HighwayTag.PRIMARY_LINK, HighwayTag.TRUNK_LINK,
+                    HighwayTag.MOTORWAY_LINK);
+
     /**
-     * Tests if the {@link AtlasObject} has a highway tag that do contain TERTIARY, SECONDARY, PRIMARY, TRUNK, or MOTORWAY
+     * Tests if the {@link AtlasObject} has a highway tag that do contain TERTIARY, SECONDARY,
+     * PRIMARY, TRUNK, or MOTORWAY
      */
-    public static final Predicate<AtlasObject> VALID_HIGHWAY_TAG = object -> Validators.hasValuesFor(object,
-            HighwayTag.class) && Validators.isOfType(object, HighwayTag.class, HighwayTag.TERTIARY,
-            HighwayTag.SECONDARY, HighwayTag.PRIMARY, HighwayTag.TRUNK, HighwayTag.MOTORWAY);
+    public static final Predicate<AtlasObject> VALID_HIGHWAY_TAG = object -> Validators
+            .hasValuesFor(object, HighwayTag.class)
+            && Validators.isOfType(object, HighwayTag.class, HighwayTag.TERTIARY,
+                    HighwayTag.SECONDARY, HighwayTag.PRIMARY, HighwayTag.TRUNK,
+                    HighwayTag.MOTORWAY);
     /**
      * Tests if the {@link AtlasObject} has a junction tag not of type ROUNDABOUT.
      */
-    public static final Predicate<AtlasObject> NOT_ROUNDABOUT_JUNCTION = object -> Validators.hasValuesFor(object,
-            JunctionTag.class) && Validators.isNotOfType(object, JunctionTag.class, JunctionTag.ROUNDABOUT);
-    
+    public static final Predicate<AtlasObject> NOT_ROUNDABOUT_JUNCTION = object -> Validators
+            .hasValuesFor(object, JunctionTag.class)
+            && Validators.isNotOfType(object, JunctionTag.class, JunctionTag.ROUNDABOUT);
+
     /**
      * Tests if the {@link AtlasObject} has a minimum or maximum speed
      */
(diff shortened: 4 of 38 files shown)


src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
```diff
@@ -53,6 +53,26 @@ private TagPredicates()
     public static final Predicate<AtlasObject> IS_HIGHWAY_AREA = object -> Validators.hasValuesFor(
             object, HighwayTag.class) && Validators.isOfType(object, AreaTag.class, AreaTag.YES);
 
+    /**
+     * Tests if the {@link AtlasObject} has a highway tag that do not contain TERTIARY_LINK,
+     * SECONDARY_LINK, PRIMARY_LINK, TRUNK_LINK, or MOTORWAY_LINK
+     */
+    public static final Predicate<AtlasObject> IS_HIGHWAY_NOT_LINK_TYPE = object -> Validators
```
sayas01: `isLink()` method in HighwayTag can be leveraged to check if the highway type is a link type or not.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
```diff
@@ -53,6 +53,26 @@ private TagPredicates()
...
+            .hasValuesFor(object, HighwayTag.class)
+            && Validators.isNotOfType(object, HighwayTag.class, HighwayTag.TERTIARY_LINK,
+                    HighwayTag.SECONDARY_LINK, HighwayTag.PRIMARY_LINK, HighwayTag.TRUNK_LINK,
+                    HighwayTag.MOTORWAY_LINK);
+
+    /**
+     * Tests if the {@link AtlasObject} has a highway tag that do contain TERTIARY, SECONDARY,
+     * PRIMARY, TRUNK, or MOTORWAY
+     */
+    public static final Predicate<AtlasObject> VALID_HIGHWAY_TAG = object -> Validators
```
sayas01: I feel like this is specific to this check and should go within the check class than in the TagPredicates. Valid highways for this check may not be the same for other checks.

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+ * This check identifies Edges with no name Tag that are between two other Edges with the same name
+ * Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.
+ *
+ * @author sugandhimaheshwaram
+ */
+public class RoadNameGapCheck extends BaseCheck
+{
+
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
```
sayas01: This needs to be regenerated.

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return an optional {@link CheckFlag} object that
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Edge edge = (Edge) object;
+        final Set<Edge> inEdges = edge.inEdges().stream().filter(this::validCheckForObject)
```
sayas01: Since the inEdges and outEdges sets are used only to collect name tags, we can filter and collect the inEdgesNameTags and outEdgesNameTags instead of inEdges and outEdges and simply compare the two sets to get set of matching name tags. This will simplify the findInEdgeOutEdgeMatchingName().

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+        final Optional<String> edgeName = edge.getName();
+        if (edgeName.isPresent() && !edgeNames.contains(edgeName.get()))
+        {
+            final String instruction = "Edge name is different from in edge name and out edge name.";
+            return Optional.of(createFlag(object, instruction));
+        }
+        return Optional.empty();
+    }
+
+    private Set<String> findInEdgeOutEdgeMatchingName(final Set<Edge> inEdges,
```
sayas01: Please see my previous comment on simplifying this method.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
smaheshwaram: Sure. I will take a look.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
smaheshwaram: Makes sense. Will move it to current file.

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
smaheshwaram: Are you saying that there is a method which gets in inEdgeNameTags()? Please correct me if my understanding is wrong.

sayas01: What I meant is if we do something like this:
```final Set<String> inEdgesNameTags = edge.inEdges().stream().filter(inEdge -> this.validCheckForObject(inEdge) && NameTag.getNameOf(inEdge).isPresent()).map(inEdge -> NameTag.getNameOf(inEdge).get()).collect(Collectors.toSet());```
for both inEdges and outEdges, then we can simply compare the two sets to get matching name tags.

File: config/configuration.json [hunk 6]
```diff
@@ -444,6 +444,23 @@
       "difficulty": "HARD"
     }
   },
+  "RoadNameGapCheck": {
+    "tag.filters":[
```
danielduhh: Are these tag filters being used?

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
smaheshwaram: Got it. I just thought for a second that there is a default method which I am not aware of. Thanks.

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+                .collect(Collectors.toSet());
+        final Set<Edge> outEdges = edge.outEdges().stream().filter(this::validCheckForObject)
+                .collect(Collectors.toSet());
+
+        if (inEdges.isEmpty() || outEdges.isEmpty())
+        {
+            return Optional.empty();
+        }
+
+        final Set<String> edgeNames = findInEdgeOutEdgeMatchingName(inEdges, outEdges);
```
danielduhh: nit: Use `this.` when calling class methods

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+            final String instruction = "Edge name is empty.";
+            return Optional.of(createFlag(object, instruction));
+        }
+
+        // Create flag when in edge and out edge name tag matches and intermediate edge has
+        // different tag name.
+        final Optional<String> edgeName = edge.getName();
+        if (edgeName.isPresent() && !edgeNames.contains(edgeName.get()))
+        {
+            final String instruction = "Edge name is different from in edge name and out edge name.";
```
danielduhh: Adding both the flagged Edge's name and the name of the mismatched in/out Edges will be useful for community mappers and debugging.

danielduhh: `getLocalizedInstruction` is a useful function for handling instructions.

You can declare your instructions as a static global variables [like this](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java#L36)
and fetch the values [like this](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java#L145)

config/configuration.json
smaheshwaram: We are not using it. I thought its necessary to add this filters in json file. But, will take it out.

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
```diff
@@ -0,0 +1,134 @@
...
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge && Edge.isMasterEdgeIdentifier(object.getIdentifier())
+                && TagPredicates.IS_HIGHWAY_NOT_LINK_TYPE.test(object)
+                && TagPredicates.VALID_HIGHWAY_TAG.test(object)
+                && HighwayTag.isCarNavigableHighway(object) && !JunctionTag.isRoundabout(object);
```
seancoulter: Since all tags checked in ```VALID_HIGHWAY_TAG``` are contained in the set of car navigable highways, we could just call ```TagPredicates.VALID_HIGHWAY_TAG.test(object)``` without ```HighwayTag.isCarNavigableHighway(object)```

seancoulter: Also, would we want to check for the missing name tag here? For ```object```

src/test/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheckTestRule.java
```diff
@@ -0,0 +1,116 @@
...
+public class RoadNameGapCheckTestRule extends CoreTestRule
+{
+    private static final String TEST_1 = "47.2136626201459,-122.443275382856";
+    private static final String TEST_2 = "47.2138327316739,-122.44258668766";
+    private static final String TEST_3 = "47.2136626201459,-122.441897992465";
+    private static final String TEST_4 = "47.2138114677627,-122.440990166979";
+
+    @TestAtlas(
+            // nodes
+            nodes = { @TestAtlas.Node(coordinates = @TestAtlas.Loc(value = TEST_1), id = "0"),
```
seancoulter: nit: we've already imported TestAtlas.Node, so the compiler should recognize just ```@Node```

src/main/java/org/openstreetmap/atlas/checks/validation/RoadNameGapCheck.java
smaheshwaram: The requirement mentioned about all the valid tags along with car navigable highway so I was using the default method. Will take it out. Thanks.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheckTestRule.java
smaheshwaram: Makes sense.

docs/checks/RoadNameGapCheck.md
```diff
@@ -0,0 +1,37 @@
+# Road name gap check
```
sayas01: We also have a document https://github.com/osmlab/atlas-checks/blob/dev/docs/available_checks.md with all the available checks. Could you add add this check to the available_checks.md doc as well?

smaheshwaram: Done.

smaheshwaram:
@danielduhh Dan got the analysis report and there are 2 scenarios which I need to take care of. All the requirements are met. However, there are 2 things which we didn't think of as part of requirement and it needs to be handled. I will work on it. Please don't merge it until then. Thanks.

danielduhh:
@smaheshwaram Sounds good. Please prepend [DNM] to your PR name  as an indicator that your check is still a work in progress. Thanks!

smaheshwaram:
Please look into this below link to understand the cases. Only valid cases are case 1 and case 4. Ignore case 2 and 3.

https://critigen.quip.com/edilAybbS3Dn/RoadNameGapCheck-Analysis

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,201 @@
...
+
+        // Create flag when we have in edge and out edge with same name but intermediate edge
+        // doesn't have a name.
+        if (!edge.getName().isPresent())
+        {
+            return Optional.of(
+                    createFlag(object, this.getLocalizedInstruction(0, edge.getOsmIdentifier())));
+        }
+        final Optional<String> edgeName = edge.getName();
+
```
smaheshwaram: The code in this if condition is handling the 2 cases. Please take a look at getMatchingInAndOutEdgeNames() method and also finaMatchingEdgeNameWithAnyInEdgeOrOutEdge() method.

smaheshwaram:
@danielduhh Analysis on this check is done. Failing cases are resolved. Dan Castro please confirm it.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+    private static final long serialVersionUID = 7104778218412127847L;
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList("Edge name is empty.",
+            "Edge name {1} is different from in edge name and out edge name.");
+
+    /**
+     * Tests if the {@link AtlasObject} has a highway tag that do contain TERTIARY, SECONDARY,
+     * PRIMARY, TRUNK, or MOTORWAY
+     */
+    private static final Predicate<AtlasObject> VALID_HIGHWAY_TAG = object -> Validators
+            .hasValuesFor(object, HighwayTag.class)
```
Bentleysb: I think this is redundant when using Validators.isOfType().

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+            && Validators.isOfType(object, HighwayTag.class, HighwayTag.TERTIARY,
```
Bentleysb: It might be good to make these values configurable, in case someone want to expand the range of this check at some point.

src/main/java/org/openstreetmap/atlas/checks/atlas/predicates/TagPredicates.java
Bentleysb: Is this being used anymore?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+ * Tag, OR the Edge has a name Tag but does not equal the name Tag of the Edges that it is between.
+ *
+ * @author sugandhimaheshwaram
+ */
+public class RoadNameGapCheck extends BaseCheck
+{
+    /**
+     * This checks if the connected edges have same direction or not.
+     */
+    class EdgePredicate
```
seancoulter: Nice use of EdgeDirectionComparator. Why do we wrap EdgeDirectionComparator in EdgePredicate? Seems like we could just use the former directly

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // Create flag when we have in edge and out edge with same name but intermediate edge
+        // doesn't have a name.
+        if (!edge.getName().isPresent())
+        {
+            return Optional.of(
+                    createFlag(object, this.getLocalizedInstruction(0, edge.getOsmIdentifier())));
+        }
+        final Optional<String> edgeName = edge.getName();
```
seancoulter: We can just get() the name directly here since we've verified it exists in the block above

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+
+        if (edgeName.isPresent() && !matchingInAndOutEdgeNames.contains(edgeName.get()))
+        {
+            // Case 1: Edge name: Tai. Incoming edge names: Tai , Shai. Outgoing edge name: Shai.
```
seancoulter: These are cases we want to flag, right? If so, we can probably indicate that and I'd vote to move to the docs

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,218 @@
...
+            // Case 2: Edge name: Tai. Incoming edge names: Shai. Outgoing edge name: Tai, Shai.
+            // Case 3: Edge name: Tai. Incoming edges names: Tai, Shai. Outgoing edge names: Shai,
+            // Pendler.
+            final Set<Edge> connectedEdges = edge.connectedEdges().stream()
+                    .filter(this::validCheckForObject).collect(Collectors.toSet());
+            if (findMatchingEdgeNameWithConnectedEdges(connectedEdges, edgeName.get()))
```
seancoulter: Here it looks like we don't flag edges if one of its in/out edges shares its name. If that's the case I think this ```if``` block can be moved to line 113 or so since it doesn't seem necessary to wait for getMatchingInAndOutEdgeNames() to finish.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: Okay. I will take this out.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: Do you mean in the json file?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: I can only use edge.getName().get() directly instead of using optional which is resulting in sonar issue. So, used this approach to handle it.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: Moved it to RoadNameGapCheck.md file case examples.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: I want to do this validation only when there is at least one matching in and out edge name. Because, there was a failing case where there is matching in and out edge but the interconnecting edge is connected to another edge which has a different name and which doesn't have either incoming or outgoing edge for it.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
smaheshwaram: I use EdgePredicate class for reusability purpose. I was using filter in flag() method in two places one for inEdge and one for outEdge. Do not like to repeat the code. Cannot add a method outside it because I need two parameters for comparison.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
seancoulter: It throws a sonar error if you do ```final String edgeName = edge.getName().get()``` on line 130?

smaheshwaram: It says call Optional isPresent() before accessing the value. So, I coded this way.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
Bentleysb: Yes

smaheshwaram: Done

docs/available_checks.md
```diff
@@ -56,6 +56,7 @@ This document is a list of tables with a description and link to documentation f
 | [InvalidAccessTagCheck](checks/invalidAccessTagCheck.md) | The purpose of this check is to identify invalid access tags. |
 | [InvalidLanesTagCheck](docs/checks/invalidLanesTagCheck.md) | The purpose of this check is to identify highways in OSM with an invalid lanes tag value. |
 | [MixedCaseNameCheck](checks/mixedCaseNameCheck.md) | The purpose of this check is to identify names that contain invalid mixed cases so that they can be edited to be the standard format. |
+| [RoadNameGapCheck](checks/RoadNameGapCheck.md) | The purpose of this check is to identify edge connected between two edges whose name tag is same. Flag the edge if the edge has a name tag different to name tag of edges connected to it or if there is no name tag itself.
```
sayas01: typo: The purpose of this check is to identify edge connected between two edges whose name tag is ~same~ different.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,210 @@
...
+            return Optional.of(
+                    createFlag(object, this.getLocalizedInstruction(0, edge.getOsmIdentifier())));
+        }
+        final Optional<String> edgeName = edge.getName();
+
+        if (edgeName.isPresent() && !matchingInAndOutEdgeNames.contains(edgeName.get()))
+        {
+            final Set<Edge> connectedEdges = edge.connectedEdges().stream()
+                    .filter(this::validCheckForObject).collect(Collectors.toSet());
+            if (findMatchingEdgeNameWithConnectedEdges(connectedEdges, edgeName.get()))
```
sayas01: nit: Could be simplified using ternary operator.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,210 @@
...
+     * @param connectedEdges
+     *            connected edges for a given edge
+     * @param edgeName
+     *            Edge name
+     * @return True if there is matching edge name with connected edge or else false.
+     */
+    private boolean findMatchingEdgeNameWithConnectedEdges(final Set<Edge> connectedEdges,
+            final String edgeName)
+    {
+        for (final Edge connectedEdge : connectedEdges)
```
sayas01: Could make use of stream to replace this for loop. ```return connectedEdges.stream().anyMatch(....)```

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,210 @@
...
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return an optional {@link CheckFlag} object that
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Edge edge = (Edge) object;
+        final EdgePredicate edgePredicate = new EdgePredicate(edge);
+        final Set<Edge> inEdges = edge.inEdges().stream().filter(this::validCheckForObject)
+                .filter(edgePredicate::isSameHeading).collect(Collectors.toSet());
```
sayas01: Couldn't we just use a single filter and put validCheckForObject and isSameHeading in the same filter?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,210 @@
...
+     *            outgoing edges
+     * @return Set of matching names for both incoming and outgoing edges.
+     */
+    private Set<String> getMatchingInAndOutEdgeNames(final Set<Edge> inEdges,
+            final Set<Edge> outEdges)
+    {
+        final Set<String> edgeNames = new HashSet<>();
+        for (final Edge inEdge : inEdges)
+        {
+            if (!inEdge.getName().isPresent())
```
sayas01: It would be cleaner if we could add this condition in the line that collects the inEdge and outEdge. 
```edge.inEdges().stream().filter(edge -> edge.getName().isPresent() && this.validCheckForObject(edge) &&....)```

File: src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java [hunk 24]
```diff
@@ -0,0 +1,210 @@
...
+            {
+                continue;
+            }
+            for (final Edge outEdge : outEdges)
+            {
+                if (!outEdge.getName().isPresent())
```
sayas01: Same comment as above.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -0,0 +1,210 @@
...
+            return object != null
+                    && this.edgeDirectionComparator.isSameDirection(this.edge, otherEdge, false);
+        }
+    }
+
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 7104778218412127847L;
+    private static final List<String> VALID_HIGHWAY_TAG_DEFAULT = Arrays.asList("primary",
+            "secondary", "tertiary", "trunk", "motorway");
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList("Edge name is empty.",
```
sayas01: nit: It would be useful to have the flagged object's identifier in the instruction.

docs/available_checks.md
smaheshwaram: Actually, it is not a typo. I need to find an edge which is connected between two edges whose name tag is same. IF the intermediate edge has no name then flag it. If the edge has a name but is different to the matching in and out edge then check if the edge has any other connected edge which has same name. If no then create a flag.

smaheshwaram:
Screen Shot 2020-01-07 at 4 41 28 PM

@danielduhh Code coverage is showing that it covered 48 lines out of 51. However, sonar quality gates failing on 0% code coverage. Build was fine until today morning. I was just addressing PR comments. Not major changes in code. So, please check if something needs to be fixed from my end.

danielduhh:
Hey! I'm not sure why Sonar is throwing fits.. the unit tests are clearly there @smaheshwaram . We can try opening and closing the PR to retrigger the build

danielduhh:
@mgcuthbert do you have any idea what's going on with Sonar? It's reporting 0% code coverage...

yiqingj:
it does look like a sonar issue always reporting 0% coverage. I'll merge this one manually for now to unblock the release.

PR #232, opened 2019-12-30
Add CheckFlag Special Identifier Method 

Bentleysb:
### Description:

Some checks are creating flags that change from run to run but are the same other than their identifier. This happens when multiple AtlasObjects are flagged but the starting object can be any of them, and the flag id is based on the starting object. Depending on which object is the starting object, the flag id changes but the rest of the flag is the same. 

Some checks avoid this by creating a set of objects to flag and passing them to the flag constructor. However this defeats the purpose of having an `addObject()` method, and makes adding an object and instructions extra cumbersome as you can't use `addObject(AtlasObject, String)`. 

This creates an alternative by adding the `CheckFlag.setObjectIdentifiersAsFlagIdentifier()` method. This method sets the flags identifier equal to a sorted concatenation of the identifiers of the flags FlaggedObjects. This gives the option to use `addObject()` and use object ids for the flag id. It also maintains the option to use a custom identifier when a specific atlas id or another special id is required. 

This also updates checks that suffer from the aforementioned variance in flag ids. 

### Potential Impact:

Some checks will have different flag ids than previously. 

### Unit Test Approach:

Added a unit test to check that the new method works to both sort and concatenate the flagged object ids as expected. 

### Test Results:

Tested running all the affected checks and verified their new flag ids were as expected.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
@@ -1,5 +1,7 @@
 package org.openstreetmap.atlas.checks.flag;
 
+import static org.openstreetmap.atlas.checks.constants.CommonConstants.EMPTY_STRING;
+
 import java.io.BufferedWriter;
 import java.io.OutputStreamWriter;
 import java.io.Serializable;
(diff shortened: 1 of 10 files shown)


Bentleysb:
We can automatically update the id when calling `addObject`. However that would mean redefining what flag identifiers are to some extent. Currently a flag identifier can be any string, be that and atlas id, random characters, the name of the check and an incrementing number, or anything else.  It is most commonly the case to use one or more concatenated atlas ids, but that is not a requirement. By automatically adding atlas ids we would essentially be requiring that the flag identifier be the atlas ids of the flagged objects, unless the `addObject` method is specifically avoided.

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -481,6 +484,20 @@ public void setChallengeName(final String challengeName)
         this.challengeName = challengeName;
     }
 
+    /**
+     * Set the {@link CheckFlag}'s identifier equal to the sorted concatenation of all the ids of
+     * its flagged features.
+     *
+     * @return this {@link CheckFlag}
+     */
+    public CheckFlag setObjectIdentifiersAsFlagIdentifier()
```
seancoulter: Can we be sure that the sorted concatenation of 2 or more IDs is not itself the same as some other single feature's ID?

seancoulter: We could append the feature type after the identifier if this is a risk, for example as in your unit test the flag would be "1n2n"

Bentleysb:
Closing to investigate some other options.

PR #238, opened 2020-01-09
Fix sonar test coverage

matthieun:
### Description:

Fix sonar test coverage, similar as what was done in https://github.com/osmlab/atlas/pull/480

### Potential Impact:

Sonar reports accurate again

### Unit Test Approach:

N/A

### Test Results:

N/A


matthieun:
Hmm tried in atlas-generator and seems it still does not work...

matthieun:
Actually it turns out I was looking at the wrong metric. The "new code" in my PR shows 0% which is why i was concerned. However after running more manual tests I could confirm that the values in the XML report path get picked up. So once this is merged, other PRs should start showing the expected code coverage.

PR #243, opened 2020-01-16
Snake road check bug

smaheshwaram:
### Description:

This is a bug fix for the existing Snake Road Check.
Existing code is meeting all the requirements except for instead of doing heading difference for connected edges the current edge was doing heading difference with all the edges on the highway. This issue is fixed with new code changes.

Criteria to identify Snake Road is as follows:
- At some point, two consecutive edges making up the Snake road must have a heading difference of at least 60 degrees.
- The snake road must have at least 3 edges
- At least one edge making up the snake road should have a valence greater than 4
- To be considered a snake road, you should either have no road name or have any connected ways that share the 
same name that you do. This is done to prevent flagging strange portions of highways that may exhibit the behavior of a snake road.

### Potential Impact:

Changes made to Atlas checks.

### Unit Test Approach:

Existing tests.

### Test Results:

Ran the check on atlas files before and after code changes and made sure the number of flags created are less. Also, with the help of analyst making sure the generated flags are still satisfying criteria for Snake Road check.

Ran SnakeRoadCheck bug on approximately 65 MB data Atlas files and found below metrics.
With old code: 84 flags were created.
With new code changes: 34 flags were created.
Flag count is decreased as expected and also verified each flag generated is satisfying snake road check requirement.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
@@ -3,6 +3,7 @@ package org.openstreetmap.atlas.checks.validation.linear.edges;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import org.openstreetmap.atlas.checks.atlas.predicates.TagPredicates;
 import org.openstreetmap.atlas.checks.atlas.predicates.TypePredicates;
@@ -168,9 +169,14 @@ public class SnakeRoadCheck extends BaseCheck<Long>
             // Process it
             walk.visitEdge(current, connection);
 
+            final Set<Edge> oneLayerRemovedConnections = walk
+                    .getConnectedMasterEdgeOfTheSameWay(connection);
+            oneLayerRemovedConnections.forEach(
+                    onceRemovedConnection -> walk.checkIfEdgeHeadingDifferenceExceedsThreshold(
+                            connection, onceRemovedConnection));
+
             // Add its neighbors to the next layer
-            walk.populateOneLayerRemovedConnections(
-                    walk.getConnectedMasterEdgeOfTheSameWay(connection));
+            walk.populateOneLayerRemovedConnections(oneLayerRemovedConnections);
 
             // If we've processed all directly connected edges, check the next layer of connections
             if (walk.getDirectConnections().isEmpty())
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadNetworkWalk.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadNetworkWalk.java
@@ -48,7 +48,7 @@ public class SnakeRoadNetworkWalk
         // We use a TreeSet here to order the edges by their Atlas identifier. This helps us easily
         // grab the first and last Edge making up the road and check connections for false
         // positives.
-        this.visitedEdges = new TreeSet<AtlasObject>((one, two) ->
+        this.visitedEdges = new TreeSet<>((one, two) ->
         {
             return Long.compare(one.getIdentifier(), two.getIdentifier());
         });
@@ -186,7 +186,6 @@ public class SnakeRoadNetworkWalk
         this.visitedEdges.add(comingTo);
         this.setGreatestValence(comingTo.start());
         this.setGreatestValence(comingTo.end());
-        checkIfEdgeHeadingDifferenceExceedsThreshold(comingFrom, comingTo);
     }
 
     /**


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
```diff
@@ -160,6 +161,10 @@ private boolean networkWalkQualifiesAsSnakeRoad(final SnakeRoadNetworkWalk walk)
      */
     private void walkNetwork(final Edge current, final SnakeRoadNetworkWalk walk)
     {
+        final Set<Edge> directConnections = walk.getConnectedMasterEdgeOfTheSameWay(current);
+        directConnections.stream().forEach(connection -> walk
```
sayas01: Isn't this change performing the same operation as before? For all connections, ```checkIfEdgeHeadingDifferenceExceedsThreshold(comingFrom, comingTo);``` was called from walk.visitedEdge previously, and now it is called here. Was this to improve performance or is there something I am missing?

smaheshwaram: The difference is before it used to call checkIfEdgeHeadingDifferenceExceedsThreshold() from visitedEdge() for all the edges not just directly connected edges to a current edge. My code will make sure that it will check the heading difference of two directly connected edges to a current edge.

sayas01: ahh I think the names got me confused.. directConnections and walk.getDirectConnections() are different but I read it as same.

seancoulter: The requirements from your PR description say to me that the 'snakiness' can be determined from any 2 consecutive edges, so shouldn't we be calling lines 164-166 inside the while loop as well? Instead of passing ```current``` we pass ```connection```

smaheshwaram: Heading difference is verified only against the current edge and it direct connections(this will be done for each edge when it is processed). Valence is checked for whole highway so that we can qualify it as a snake road.
Moving these lines inside the while loop will make the check do the same thing repeatedly for an edge object that will be processed for snakeRoadCheck.

seancoulter: Sorry if I'm misunderstanding. But aren't we now only calling ```checkIfEdgeHeadingDifferenceExceedsThreshold(current, connection))``` between the starting edge and all of its connected edges? When we want to be calling it on every consecutive edge in the original Way. This might be doable if we reassign ```current``` whenever we move to the next layer of edges

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
```diff
@@ -160,6 +161,10 @@ private boolean networkWalkQualifiesAsSnakeRoad(final SnakeRoadNetworkWalk walk)
      */
     private void walkNetwork(final Edge current, final SnakeRoadNetworkWalk walk)
     {
+        final Set<Edge> directConnections = walk.getConnectedMasterEdgeOfTheSameWay(current);
```
seancoulter: to avoid reprocessing, we can just say ```final Set<Edge> directConnections = walk.getDirectConnections()``` here. Might need to change Set to a LinkedList doing this

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
smaheshwaram: yes, thats true.
My assumption is when snakeRoadcheck is invoked its invoked for every edge, if we do this inside for loop, heading difference is calculated repeatedly for all consecutive edges.
So by not doing this only the edge where heading difference is greater than threshold is flagged.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
```diff
@@ -168,9 +169,13 @@ private void walkNetwork(final Edge current, final SnakeRoadNetworkWalk walk)
             // Process it
             walk.visitEdge(current, connection);
 
+            final Set<Edge> oneLayerRemovedConnections = walk
+                    .getConnectedMasterEdgeOfTheSameWay(connection);
+            oneLayerRemovedConnections.forEach(
+                    conn -> walk.checkIfEdgeHeadingDifferenceExceedsThreshold(connection, conn));
```
seancoulter: conn -> onceRemovedConnection

seancoulter: (to avoid abbreviating)

seancoulter:
Looks good! Just had a tiny request.

smaheshwaram:
All review comments were addressed. Ready for final review and approvals from devs.

smaheshwaram:
All review comments were taken care of. Final review and approvals are needed from devs. Thanks in advance.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
smaheshwaram: Done

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SnakeRoadCheck.java
smaheshwaram: Done

smaheshwaram:
I have lot of commits for this branch. So, I force pushed the code which is why it is asking for approvals again. It had 2 approvals already.

PR #246, opened 2020-01-21
DuplicateNodeCheck, every overlapped node of a location is flagged.

smaheshwaram:
### Description:

This PR changes were related to fixing an existing bug for DuplicateNodeCheck.
Actual: Only one random node was flagged which got overlapped with many nodes of a specific location.
Expected: All overlapped nodes needs to be flagged for a specific location.

### Potential Impact:

Changes made to Atlas-checks.

### Unit Test Approach:

Fixing the existing unit test.

### Test Results:

Locally tested this check for Atlas files before and after making code changes. Working as expected.


src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicateNodeCheck.java
```diff
@@ -39,7 +40,7 @@ public DuplicateNodeCheck(final Configuration configuration)
     @Override
     public boolean validCheckForObject(final AtlasObject object)
     {
-        return object instanceof Node && !this.isFlagged(((Node) object).getLocation());
```
Bentleysb: This does accomplish flagging all overlapping nodes, but it would be better if there was only one flag per location. That way if someone was fixing these issues they were not seeing the same problem multiple times. 
All the Nodes in a location could be in the same flag, or just the same single Node for each location (highest or lowest identifier perhaps).

PR #247, opened 2020-01-23
[DNM] Edge crossing edge check

smaheshwaram:
### Description:

EdgeCrossingEdgeCheck flags all of the edges that cross a starting edge. Depending on when an edge is checked its crossing edges may or may not be flagged based on whether they are flagged already. This PR is a bug fix to flag all these edges.

### Potential Impact:

Code changes to the Atlas-checks.

### Unit Test Approach:

Fixing existing unit tests as per new change set.

### Test Results:

Ran it on few Atlas files locally to check how the edges are flagged. Working as expected.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheckTest.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheckTest.java
@@ -1,0 +1,92 @@
+package org.openstreetmap.atlas.checks.validation.intersections;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.openstreetmap.atlas.checks.configuration.ConfigurationResolver;
+import org.openstreetmap.atlas.checks.validation.verifier.ConsumerBasedExpectedCheckVerifier;
+
+/**
+ * @author mkalender
+ * @author sayana_saithu
+ */
+public class LineCrossingWaterBodyCheckTest
+{
+    private static LineCrossingWaterBodyCheck check = new LineCrossingWaterBodyCheck(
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/EdgeCrossingEdgeCheck.java
```diff
@@ -188,7 +181,7 @@ public boolean validCheckForObject(final AtlasObject object)
     private boolean isValidCrossingEdge(final AtlasObject object)
     {
         if (Edge.isMasterEdgeIdentifier(object.getIdentifier())
-                && !TagPredicates.IS_AREA.test(object) && !this.isFlagged(object.getIdentifier()))
```
Bentleysb: This does prevent an Edge from being inconsistently flagged, but it also creates a lot of duplication in the flags. Ideally we would want to avoid both those things.

smaheshwaram: I understand your concern. But, I thought we wanted to flag every edge unlike the node in DuplicateNodeCheck. Please let me know if you are suggesting to fix this bug in similar fashion of duplicate node check.

seancoulter: Just a suggestion, we could do something like grabbing all ```collectedEdges``` and flagging all the edges except for the one with the lowest OSM ID. This should ensure we have consistent flagging per set of collected edges, and flags don't depend on which edge is the starting edge.

smaheshwaram:
@danielduhh Functionality change required. Modified the status to DNM.

PR #251, opened 2020-02-05
[DNM] InconsistentRoadClassificationCheck bug fix

smaheshwaram:
### Description:

InconsistentRoadClassificationCheck flags Edges in a nondeterministic manor. It should only flag master Edges and use the OsmWayWalker. Fixed this issue in current PR.

### Potential Impact:

Changes made to Atlas checks.

### Unit Test Approach:

No additional tests are included.

### Test Results:

Tested by running the check on atlas files and the changes are working as expected.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InconsistentRoadClassificationCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InconsistentRoadClassificationCheck.java
@@ -233,6 +233,7 @@ public class InconsistentRoadClassificationCheck extends BaseCheck<Long>
             final Edge edge)
     {
         return edge.outEdges().stream()
+                .filter(Edge::isMasterEdge)
                 .filter(connectedEdge -> referenceHighwayType
                         .isOfEqualClassification(connectedEdge.highwayTag())
                         && this.areInTheSimilarDirection(edge, connectedEdge));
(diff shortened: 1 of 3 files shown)


seancoulter:
Nice job. To avoid picking up the referenceEdge's reverse edge we should filter for master edges here: https://github.com/osmlab/atlas-checks/pull/251/files#diff-f639779faa99d55536d586972f65211eR253 and another filter here: https://github.com/osmlab/atlas-checks/pull/251/files#diff-f639779faa99d55536d586972f65211eR345

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InconsistentRoadClassificationCheck.java
```diff
@@ -112,7 +114,7 @@ public boolean validCheckForObject(final AtlasObject object)
                 .findInconsistentEdges(edge);
         if (!inconsistentEdgeTuples.isEmpty())
         {
-            final CheckFlag flag = createFlag(item,
+            final CheckFlag flag = createFlag(new OsmWayWalker((Edge) item).collectEdges(),
```
seancoulter: The default OsmWayWalker (which is being used here) picks up reverse and master edges, so we'd want to change the ```nextCandidates``` function in the constructor.

Also, I'm not sure if this is the best place to use the WayWalker. Maybe it should be used in a larger sense to traverse the edge network looking for inconsistencies.

smaheshwaram: I understand your concern about adding OSMWayWalker. I added it because the bug description mentioned about it. We can take input from Bentley and fix it according to that.

smaheshwaram:
@danielduhh Closing this current PR because it got messed up with additional commits. Will create new PR for this bug.

PR #258, opened 2020-02-11
Add new AtGradeSignPostCheck

sayas01:
### Description:

This PR adds new AtGradeSignPostCheck. The check flags any at-grade intersections that are not part of a destination_sign relation. An at-grade intersection is defined as an intersection of two or more edges with valid highway classification, all at the same z-level. The intersecting node can be part of multiple destination_sign relation. So even if the node is a member of a destination_sign relation, the check ensures to flag any missing destination_sign relation the node can be part of, based on the intersecting edges. The check also flags destination_relations of at-grade intersections that are missing destination_sign tag. If the intersection is an exit node of a roundabout, the entire roundabout is flagged in addition to the intersecting node and roundabout exit edge.

Example: https://www.openstreetmap.org/node/892115562
image

### Potential Impact:

None. Just a new check to identify at-grade intersections with missing destination_sign tag.

### Unit Test Approach:

Added new unit tests.

### Test Results:

All tests passing.


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,644 @@
...
+                    instructionIndex = INSTRUCTION_INDEX_TWO;
+                    entitiesToBeFlagged.addAll(connectedEdgesNotFormDestinationRelation);
+                }
+            }
+        }
+        return new FlaggedIntersection(instructionIndex, entitiesToBeFlagged);
+    }
+
+    /**
+     * Get items part of intersection that is not part of any destination_sign relations
```
Bentleysb: This description seems a little confusing,as it seems that the inputs are already filtered for items that aren't part of destination sign relations.

sayas01: This method returns intersections that do not have destination sign relations. The params for this method are not filtered prior based on destination_sign relation. I have slightly updated the comment. Let me know if it is still unclear. I will be happy to be more descriptive.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,644 @@
...
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return an optional {@link CheckFlag} object that
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        if (String.valueOf(object.getOsmIdentifier()).equals("1720823768"))
+        {
+            System.out.println(object);
+        }
```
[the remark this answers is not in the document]
sayas01: Yp! Removed it! :)

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
Bentleysb: Perhaps I am just missing something obvious here, but I am not seeing where the filtering for destination signs is happening. It looks the only filtering is for roundabout tags, and then everything gets added to the entitiesToBeFlagged set.

sayas01: My bad! This method is called only if there is no destination_sign relation. This method simply flags the items in the input hashmaps. Since the flag method was getting long, I just pulled out the section that flags the items in the maps to this method. I have renamed the method and replaced the java doc for the method.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+                                final Optional<HighwayTag> atlasEntityHighway = HighwayTag
+                                        .highwayTag(atlasEntity);
+                                return atlasEntityHighway.isPresent() && validHighwayTypesOfOutEdge
+                                        .contains(atlasEntityHighway.get().getTagValue());
+                            }).collect(Collectors.toSet());
+                    // If any of the edges is a roundabout, add it to roundabout map
+                    if (filteredByHighways.stream().anyMatch(JunctionTag::isRoundabout)
+                            || JunctionTag.isRoundabout(inEdge))
+                    {
+                        roundAboutInEdgeToOutEdgeMap.put(inEdge, filteredByHighways);
```
seancoulter: Why don't we put only the edges in ```filteredByHighways``` that are explicitly roundabouts in ```roundAboutInEdgeToOutEdgeMap```? And the rest in ```inEdgeToOutEdgeMap```

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+    private Set<Relation> getRelationsWithMissingDestinationTag(
+            final Set<Relation> destinationSignRelations)
+    {
+        return destinationSignRelations.stream()
+                .filter(relation -> relation.tag(DestinationTag.KEY) == null)
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Collects all roundabout edges starting with the given given
```
seancoulter: type: given -> edge

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
sayas01: So if there is at least one roundabout edge, then the intersection should be in roundAboutInEdgeToOutEdgeMap. If one of the inEdge is an entrance to a roundabout and outEdge is exit to a roundabout, there will not be a roundabout tag to these edges and I do not want them to be added to inEdgeToOutEdgeMap. This way, I can be sure to assign roundabout instruction to the intersection of these edges.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+     *         input params
+     */
+    private FlaggedIntersection getFlaggedIntersection(
+            final Map<AtlasEntity, Set<AtlasEntity>> roundAboutInEdgeToOutEdgeMap,
+            final Map<AtlasEntity, Set<AtlasEntity>> inEdgeToOutEdgeMap)
+    {
+        final Set<AtlasEntity> entitiesToBeFlagged = new HashSet<>();
+        int instructionIndex = -1;
+        // Flag all in and out edges
+        if (roundAboutInEdgeToOutEdgeMap.isEmpty() && inEdgeToOutEdgeMap != null)
```
seancoulter: Is there any way we could have a nonempty ```roundAboutInEdgeToOutEdgeMap``` and ```inEdgeToOutEdgeMap```? If so there should be a case to handle flagging both

sayas01: No there will not be because populateInEdgeToOutEdgeMaps method populates either roundAboutInEdgeToOutEdgeMap or inEdgeToOutEdgeMap based on presence of roundabout edges in the output.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+        }
+        // For each inEdge, get the list of potentially matching out edges
+        // Matching out edges are based on z level and highway type.
+        // For each inEdge, store the inEdge and corresponding outEdges in inEdgeToOutEdgeMap
+        // If any of the out edge is a roundabout edge, store the roundabout edges and the inEdge in
+        // roundAboutInEdgeToOutEdgeMap
+        final Map<String, Map<AtlasEntity, Set<AtlasEntity>>> mapOfMatchingInAndOutEdges = this
+                .populateInEdgeToOutEdgeMaps(inEdges, outEdges);
+        final Map<AtlasEntity, Set<AtlasEntity>> inEdgeToOutEdgeMap = mapOfMatchingInAndOutEdges
+                .getOrDefault(NON_ROUNDABOUT_INTERSECTION_MAP, null);
```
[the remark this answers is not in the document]
smaheshwaram: Also, can we name the variable as nonRoundaboutInEdgeToOutEdgeMap instead of inEdgeToOutEdgeMap.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+        // Collect all destination sign relations, the node is a member of
+        final Optional<Set<Relation>> destinationSignRelations = this
+                .getParentDestinationSignRelations(intersectingNode);
+        final FlaggedIntersection flaggedIntersection = destinationSignRelations.isEmpty()
+                ? this.getFlaggedIntersection(roundAboutInEdgeToOutEdgeMap, inEdgeToOutEdgeMap)
+                : this.getIntersectionsWithIncompleteDestinationSignRelation(
+                        roundAboutInEdgeToOutEdgeMap, inEdgeToOutEdgeMap, intersectingNode,
+                        destinationSignRelations.get());
+        final int instructionIndex = flaggedIntersection.getInstructionIndex();
+        final Set<AtlasEntity> entitiesToBeFlagged = flaggedIntersection.getFlaggedItems();
```
smaheshwaram: Are we verifying that getFlaggedItems() is not null or not empty. Because, setOfFlaggedItems variable can be null.

smaheshwaram:
@sayas01 LGTM except for few minor comments. Nice job on the PR.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
```diff
@@ -0,0 +1,634 @@
...
+    private static final String ROUNDABOUT_INTERSECTION_MAP = "roundaboutMatchingEdgesMap";
+
+    private static final int INSTRUCTION_INDEX_ZERO = 0;
+    private static final int INSTRUCTION_INDEX_ONE = 1;
+    private static final int INSTRUCTION_INDEX_TWO = 2;
+    private static final int INSTRUCTION_INDEX_THREE = 3;
+    private static final int INSTRUCTION_INDEX_FOUR = 4;
+
+    private static final int MINIMUM_NODE_VALENCE = 3;
+    private static final Angle DEFAULT_OPPOSITE_DIRECTION_LOWER_LIMIT = Angle.degrees(171);
```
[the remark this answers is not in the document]
sayas01: The default values used in the edge direction comparator class was giving FPs. So I used custom angles based on the analysis of outputs for various angles.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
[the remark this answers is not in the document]
sayas01: done!

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/AtGradeSignPostCheck.java
sayas01: he instructionIndex is updated only if there is some item to be flagged and the line 217 ensures that Optional.empty() is returned if there are no flaggedItems.

smaheshwaram: Yes, you are right. instructionIndex will be -1 if no items to be flagged and also getFlaggedItems() can be null because setOfFlaggedItems is not set anywhere. So, please move the if condition above this line.

sayas01: flaggedIntersection.getFlaggedItems() wouldn't be null in any case. It would be an empty set. But if instructionIndex =-1, flaggedIntersection.getFlaggedItems() need not be called and so I moved the line as requested.

config/configuration.json
```diff
@@ -940,5 +940,13 @@
       "difficulty": "MEDIUM",
       "defaultPriority": "MEDIUM"
     }
+  },
+  "AtGradeSignPostCheck": {
+    "enabled": true,
```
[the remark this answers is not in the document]
sayas01: Removed it!

danielduhh:
Thanks for your reviews everyone !

PR #259, opened 2020-02-13
Sharded Integrity Checks Job

Bentleysb:
### Description:

This builds on the work in https://github.com/osmlab/atlas-checks/pull/186 to create a sharded implementation of the Atlas Checks framework. By using input data that has been cut into "shards" the job can make better use of parallelization and keep a lower memory footprint. This vastly improves performance when run on a Spark cluster with large scale data sets. 

Outside of hardware requirements, there are a few things of note when comparing this job to the current one. 

- The input data must be Atlas files in the structure input_folder/iso_country_code/iso_z-x-y.atlas. PBF input is not yet supported.
- This job requires a reference to how the Atlases are sharded. This can be provided via a sharding.txt file in that input path, or by using the `sharding` argument. Dynamic and SlippyTile sharding schemas are accepted.
- Due to the nature of the parallelization, some flags will be duplicated internally by this job. These flags are de-duplicated before being output, but this makes this job more sensitive to nondeterministic flagging behavior. 
- By default this job loads Atlas files using DynamicAtlases, and stores spark RDDs using memory first. Both of these settings work best when using the job at scale on a Spark cluster. For running on a single machine with more limited hardware, it is recommended to set the `multiAtlas` and `sparkStorageDiskOnly` to `true`. 
- This job is not currently able to upload flags directly to MapRoulette. 

Known Issue:

- Due to the way the atlas data is loaded per shard, with a limited expansion distance, some data is just too large to process correctly. This is generally only an issue for very large relations, where only a subset of their members are loaded. This can cause incorrect flagging of these features. 

There are some changes that effect the larger framework, by updating existing classes to better serve the new job:

- Common program arguments and functions are now stored in the IntegrityChecksCommandArguments class.
- A new unique identifier system has been added to CheckFlags, to allow for better de-duplication. This system generates a set of AtlasObject to identify the flags. This change has been propagated through the flag outputs and difference commands. The old identifier still remains, and this should be completely reverse compatible.
- Some deprecated classes in org.openstreetmap.atlas.checks.event have been removed, and all references have been replaced with the equivalent classes from org.openstreetmap.atlas.event in the Atlas library. This is a breaking change.
- The FlaggedObject class has a new method for converting to a CompleteEntity. This is used to disconnect the entity from its atlas, making it less memory intensive to perform the flag de-duplication process. This is a breaking change. 

### Potential Impact:

This new job will likely vary in flag counts, in comparison to the old job. This is due to the nondeterministic sensitivity and large feature loading issue. 
There are a couple breaking changes with the org.openstreetmap.atlas.checks.event package and FlaggedObject class. 

### Unit Test Approach:

Unit tests have been added to cover all major changes. Most of them are pretty strait forward. The unit tests for the new Spark Job are some what limited and nontraditional. This is due to known limitations with unit testing Spark. Testing is generally limited to running the entire job. These test do this using test inputs an checking for the correct number of outputs. 

### Test Results:

The new job was run at scale on a Spark cluster, and compared with the outputs of the current job. 
Run time for the new job was about 4 times faster than the current job.
Evaluation of the differences in flag output found there to be no variation outside of items  effected by the nondeterministic sensitivity and know issue with large features. That variation only effects about 0.1% of flags.


src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,434 @@
...
+
+        // Set up for unioning
+        final JavaPairRDD<String, UniqueCheckFlagContainer> firstCountryRdd = countryRdds.get(0);
+        countryRdds.remove(0);
+
+        // Add UI title
+        this.getContext().setJobGroup("0", "Conflate flags and generate outputs");
+        // union the RDDs, combine the UniqueCheckFlagContainers by country, and proccess the flags
+        // through the output event service
+        this.getContext().union(firstCountryRdd, countryRdds)
```
jklamer: Is getting Checks per country not going to be supported moving forward? // Is it depended on the being able to access the persisted RDD?

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,434 @@
...
+     * @param country
+     *            {@link String} country code
+     * @param configuration
+     *            {@link org.openstreetmap.atlas.generator.tools.spark.SparkJob} configuration map
+     * @return {@link Function} that fetches atlases/
+     */
+    private Function<Shard, Optional<Atlas>> atlasFetcher(final String input, final String country,
+            final Map<String, String> configuration)
+    {
+        final HadoopAtlasFileCache cache = new HadoopAtlasFileCache(input, configuration);
```
jklamer: I believe making a new one of these WILL work for every atlas fetcher, but if you can make it a transient static lazy initialized I believe that will mean each JVM working in the spark cluster will use its own, which may allow for optimizations.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
Bentleysb: I'm not sure I exactly follow the question, but all this has to do with creating job labels for the Spark UI. In order to run and label each country as a different job in spark this is persisting the RDD, calling count (just to force execution), and then unioning the results to get everything into one stream. So checks are fully by country.

src/main/java/org/openstreetmap/atlas/checks/base/CheckResourceLoader.java
```diff
@@ -60,6 +60,8 @@
     private final MultiMap<String, String> countryGroups = new MultiMap<>();
     private final Boolean enabledByDefault;
     private final String enabledKeyTemplate;
+    private final String countryWhitelistTemplate = "%s." + BaseCheck.PARAMETER_WHITELIST_COUNTRIES;
+    private final String countryBlacklistTemplate = "%s." + BaseCheck.PARAMETER_BLACKLIST_COUNTRIES;
```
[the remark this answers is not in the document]
Bentleysb: I am not sure why I made these like this. I think it might be best to make them static variable, as apposed to local. Any thoughts?

src/main/java/org/openstreetmap/atlas/checks/commands/AtlasChecksLogDiffSubCommand.java
```diff
@@ -78,19 +82,28 @@ public AtlasChecksLogDiffSubCommand()
...
-                .map(object -> object.getAsJsonObject().get(PROPERTIES).getAsJsonObject()
-                        .get(IDENTIFIER).getAsString())
-                .collectToSet();
+        final JsonObject flagProperties = flagJson.get(PROPERTIES).getAsJsonObject();
+        return flagProperties.has(IDENTIFIERS)
+                ? Iterables.stream(flagProperties.get(IDENTIFIERS).getAsJsonArray())
+                        .map(JsonElement::getAsString).collectToSet()
+                : Iterables.stream(flagJson.get(FEATURES).getAsJsonArray()).filter(object -> object
+                        .getAsJsonObject().get(PROPERTIES).getAsJsonObject().has(IDENTIFIER)
+                        || object.getAsJsonObject().get(PROPERTIES).getAsJsonObject().has("ItemId"))
```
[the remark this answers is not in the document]
Bentleysb: This is a bit of a funny one in that it is not a standard geojson key. It should probably be defined as a public constant somewhere in the output event classes. There are a bunch of similar items in the event classes that could use a refactor. It would probably be best to have a separate PR to update all of them.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -134,6 +81,27 @@ public static void main(final String[] args)
...
+    /**
+     * Gets complex entities
+     *
+     * @param check
+     *            A {@link BaseCheck} object
+     * @param atlas
+     *            An {@link Atlas} object
+     * @return An {@link Iterable} of {@link ComplexEntity}s
+     */
+    protected static Iterable<ComplexEntity> findComplexEntities(final BaseCheck check,
```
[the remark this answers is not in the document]
Bentleysb: Nope, its use has been replaced by `IntegrityChecksCommandArguments.objectsToCheck`. I will remove this.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -153,35 +121,11 @@ private static void executeChecks(final String country, final Atlas atlas,
     {
         final Pool checkExecutionPool = new Pool(checksToRun.size(), "Check execution pool",
                 POOL_DURATION_BEFORE_KILL);
```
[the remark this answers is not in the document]
Bentleysb: This is a constant value of 300 seconds. It configures the Pool to kill any check that takes longer then that time to run.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
Bentleysb: The caches are using a global namespace, so I think this should already effectively be one cache store per machine in the cluster.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
```diff
@@ -211,7 +133,7 @@ public String getName()
         return "Integrity Check Spark Job";
     }
 
-    @SuppressWarnings({ "rawtypes" })
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
sayas01: What are these warnings?

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
```diff
@@ -0,0 +1,189 @@
...
+    protected enum OutputFormats
+    {
+        FLAGS,
+        GEOJSON,
+        METRICS,
+        TIPPECANOE
+    }
+
+    @Deprecated
+    protected static final Switch<String> ATLAS_FOLDER = new Switch<>("inputFolder",
```
sayas01: Why is this deprecated?

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
```diff
@@ -0,0 +1,189 @@
...
+import org.openstreetmap.atlas.utilities.maps.MultiMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handles arguments and base functionality for integrity check sparkjobs generating commands
+ *
+ * @author jklamer
+ */
+public abstract class IntegrityChecksCommandArguments extends SparkJob
```
sayas01: nit: missing serialVersionUID

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessor.java
```diff
@@ -18,11 +18,13 @@
...
 /**
- * A {@link Processor} for {@link CheckFlagEvent}s to write them into GeoJson files
+ * A {@link org.openstreetmap.atlas.event.Processor} for {@link CheckFlagEvent}s to write them into
+ * GeoJson files
  *
  * @author brian_l_davis
  */
-public final class CheckFlagGeoJsonProcessor implements Processor<CheckFlagEvent>
+public final class CheckFlagGeoJsonProcessor
+        implements org.openstreetmap.atlas.event.Processor<CheckFlagEvent>
```
sayas01: Couldn't we import this?

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessor.java
```diff
@@ -111,7 +113,7 @@ public void process(final CheckFlagEvent event)
 
     @Override
     @Subscribe
-    public void process(final ShutdownEvent event)
+    public void process(final org.openstreetmap.atlas.event.ShutdownEvent event)
```
sayas01: Why are these changes?

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -451,6 +469,22 @@ public int hashCode()
         return new MultiIterable<>(getShapes()).iterator();
     }
 
+    /**
+     * Decouple the {@link CheckFlag} from any
+     * {@link org.openstreetmap.atlas.geography.atlas.Atlas}s by making all the
```
sayas01: nit: Could we do a static import here?

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java
```diff
@@ -17,14 +18,14 @@
  */
 public abstract class FlaggedObject implements Serializable, Located
 {
-    protected static final String COUNTRY_MISSING = "NA";
     protected static final String AREA_TAG = "Area";
+    protected static final String COUNTRY_MISSING = "NA";
```
sayas01: What are these changes?

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java
```diff
@@ -23,25 +23,29 @@
...
 /**
- * Wraps a {@link CheckFlag} for submission to the {@link EventService} for handling {@link Check}
- * results
+ * Wraps a {@link CheckFlag} for submission to the
+ * {@link org.openstreetmap.atlas.event.EventService} for handling {@link Check} results
  *
  * @author mkalender, bbreithaupt
  */
-public final class CheckFlagEvent extends Event
+public final class CheckFlagEvent extends org.openstreetmap.atlas.event.Event
```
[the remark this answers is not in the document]
Bentleysb: I'm able to import, not sure why that ended up like that. I will fix that.

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagEvent.java
```diff
@@ -143,6 +147,7 @@ else if (flaggedRelations.size() != 1 && !feature.has(GEOMETRY))
         flagProperties.add("feature_properties", featureProperties);
         flagProperties.add("feature_osmids", uniqueFeatureOsmIds);
         flagProperties.addProperty("feature_count", featureProperties.size());
+        flagProperties.add(IDENTIFIERS, GSON.toJsonTree(flag.getUniqueIdentifiers()));
```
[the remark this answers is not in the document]
Bentleysb: Yes it is converting it to a json array. 
I think there are a few advantages. It makes it easier to read and prettify when in json format. There is no issue with trying to deal with gigantic unbroken strings.  A lot of languages have no problem loading a json array into a native collection format, so they can then easily use the ids in a separated (search) or combined (id) manor. 
That said, it would probably work just about as well to have his as a csv string. For an example of how it is currently used by downstream tools look at the changes to the geojson and log diff tools in this PR.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
```diff
@@ -0,0 +1,189 @@
...
+    protected static final String OUTPUT_METRIC_FOLDER = "metric";
+    protected static final String OUTPUT_TIPPECANOE_FOLDER = "tippecanoe";
+    static final Switch<List<String>> CHECK_FILTER = new Switch<>("checkFilter",
+            "Comma-separated list of checks to run",
+            checks -> Arrays.asList(checks.split(CommonConstants.COMMA)), Optionality.OPTIONAL);
+    // Configuration
+    static final Switch<StringList> CONFIGURATION_FILES = new Switch<>("configFiles",
+            "Comma-separated list of configuration datasources.",
+            value -> StringList.split(value, CommonConstants.COMMA), Optionality.OPTIONAL);
+    static final Switch<String> CONFIGURATION_JSON = new Switch<>("configJson",
```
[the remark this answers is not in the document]
Bentleysb: It so you could pass the contents of a configuration file directly as json, instead of requiring the file.

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -48,18 +49,18 @@
...
  */
 public class CheckFlag implements Iterable<Location>, Located, Serializable
 {
-    private static final long serialVersionUID = -1287808902452203852L;
-    private static final Logger logger = LoggerFactory.getLogger(CheckFlag.class);
-
     private static final Distance TEN_METERS = Distance.meters(10);
-
-    private final String identifier;
+    private static final String NULL_IDENTIFIERS = "nullnull";
```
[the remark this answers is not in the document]
Bentleysb: It is used to filter out bad unique ids created from flagged points that are not from atlas entities.

src/main/java/org/openstreetmap/atlas/checks/commands/AtlasChecksLogDiffSubCommand.java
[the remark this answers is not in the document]
Bentleysb: In looking at this again I have realized this is the old identifier key that CheckFlagEvent used to use. It is still in this tool for reverse compatibility.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
Bentleysb: I'm not sure what rawtypes is for, but was already there. unchecked is for unchecked casts, that are happening when getting values from the command map.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
Bentleysb: That is a carryover from @jklamer's original PR. My assumption, and why I left it, is that it would be better to use the 'input' switch from the SparkJob base class.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
Bentleysb: I don't think you need this for an abstract class.

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagGeoJsonProcessor.java
Bentleysb: Changing to use the classes in atlas, instead of replicating them in atlas checks.

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
Bentleysb: It would not be static, as Atlas is a class. I could import the class, but that seems excessive for just a doc reference.

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedObject.java
Bentleysb: Automatic code rearrangement.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,434 @@
...
+                                .map(fetcher).filter(Optional::isPresent).map(Optional::get)
+                                .collect(Collectors.toList()));
+            }
+            else
+            {
+                final DynamicAtlasPolicy policy = new DynamicAtlasPolicy(fetcher,
+                        sharding.getValue(), Collections.singleton(task.getShard()),
+                        task.getShard().bounds().expand(shardDistanceExpansion))
+                                .withDeferredLoading(true).withAggressivelyExploreRelations(true)
+                                .withExtendIndefinitely(false);
```
seancoulter: Wondering how useful it'd be to have a configurable DynamicAtlasPolicy, or one defined in a file

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityCheckSparkJob.java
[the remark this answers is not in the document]
Bentleysb: Sorry, yes this is minutes.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
Bentleysb: Interesting! I don't foresee a need to serialize an entire spark job, but as the class is serializable I will add this.

src/main/java/org/openstreetmap/atlas/checks/distributed/IntegrityChecksCommandArguments.java
```diff
@@ -0,0 +1,189 @@
...
+            final Pattern atlasFilePattern = Pattern
+                    .compile(String.format(ATLAS_FILENAME_PATTERN_FORMAT, country));
+
+            // Go over shard files for the country and use file name pattern to find out shards
+            FileSystemHelper.listResourcesRecursively(countryDirectory, sparkContext, atlasFilter)
+                    .forEach(shardFile ->
+                    {
+                        final String shardFileName = shardFile.getName();
+                        final Matcher matcher = atlasFilePattern.matcher(shardFileName);
+                        if (matcher.find())
```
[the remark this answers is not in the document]
Bentleysb: I agree, I have added that.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,434 @@
...
+
+        // Check inputs
+        if (countries.isEmpty())
+        {
+            logger.error("No countries found to run");
+            return;
+        }
+        for (final String country : countries)
+        {
+            final Set<Check> checksLoadedForCountry = checkLoader.loadChecksForCountry(country);
```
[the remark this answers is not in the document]
Bentleysb: Country strings are only invalid if they don't have corresponding files. If that is the case then that is handled by the block on line 194.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,434 @@
...
+            }
+        }
+        if (this.countryChecks.isEmpty())
+        {
+            logger.error("No checks loaded for any of the countries provided. Skipping execution");
+            return;
+        }
+
+        // Find the shards for each country atlas files
+        final MultiMap<String, Shard> countryShards = countryShardMapFromShardFiles(
```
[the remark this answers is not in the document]
Bentleysb: There is a check for this on line 171.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
Bentleysb: This is definitely something I want to have eventually. That may be part of the solution for dealing with large relations. For now I think it is fine to have it hard coded while that solution is architected.

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -0,0 +1,435 @@
...
+        final Sharding sharding = AtlasSharding.forString(shardingFilePath,
+                this.configurationMap());
+        final Broadcast<Sharding> shardingBroadcast = this.getContext().broadcast(sharding);
+        final Distance distanceToLoadShards = (Distance) commandMap.get(EXPANSION_DISTANCE);
+
+        // Check inputs
+        if (countries.isEmpty())
+        {
+            logger.error("No countries found to run");
+            return;
```
Bentleysb: One of @seancoulter's comments made me realize that there are a lot of soft failures used in the start method to short circuit execution of the new spark job. These were added in the original PR, so I had just left them alone. I am now wondering if it might be better to throw exceptions, to make it more evident why the job did not do anything. Does anyone have an opinion on this?

danielduhh:
Looks good pending code smell clean up!

Bentleysb:
@danielduhh the remaining code smells are for the deprecated ATLAS_FOLDER switch.

danielduhh:
@sayas01 @seancoulter @smaheshwaram this good to merge?

danielduhh:
💯

PR #265, opened 2020-03-03
[MapRouletteUploadCommand] Encode challenge names with URIUtil

danielduhh:
### Description:

This updates the MapRoulette createChallenge function to use [URIUtil.encodeQuery](https://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/util/URIUtil.html#encodeQuery(java.lang.String)) to encode Challenges names in the /api/v2/project/{project_id}/challenge/{challange_name_string}.

This will fix an issue where the MapRouletteUpload command would return a 500 when uploading tasks to an existing challenge.
```
16:54:37.982 [main] DEBUG o.o.a.c.m.MapRouletteConnection - 500 - 
{
	"status": "KO",
	"message": "Challenge with name USA - SelfIntersectingPolylineCheck already exists in the database"
}
```

`URLEncoder.encode` replaces space characters with `+` instead of `%20`, which is not properly decoded by the MapRoulette API. 

### Potential Impact:

This will allow us to use MapRouletteUploadCommand on existing projects/challenges

### Unit Test Approach:

Ran the MapRouletteUploadCommand to upload challenges to a test project. Ran the command again and challenges were refreshed with the most recent data.

### Test Results:

Tests worked as expected

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteConnection.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteConnection.java
@@ -3,12 +3,13 @@ package org.openstreetmap.atlas.checks.maproulette;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URISyntaxException;
-import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


dcastrowa:
Nice! This will definitely be a useful added function.

PR #266, opened 2020-03-03
OceanBleedingCheck new check

seancoulter:
### Description:

A new check to flag streets (configurable), railways(configurable) and buildings that bleed into (intersect) oceans. Oceans are defined by a set of configurable tags. There are invalid and valid tags for an ocean that can be used to precisely define it. The goal is to bring attention to OSM features that should have their tags redefined or their geometry rearranged so they either validly intersect the ocean or no longer intersect it.

This PR also has a small configuration update for LineCrossingWaterBodyCheck.

### Potential Impact:

None, it's a new check

### Unit Test Approach:

New unit tests for various features that can bleed into the ocean

### Test Results:

Passing tests


config/configuration.json
```diff
@@ -964,5 +969,20 @@
       "difficulty":"NORMAL",
       "defaultPriority":"MEDIUM"
     }
+  },
+  "OceanBleedingCheck": {
```
sayas01: Could you also add the md doc for this check and update the available_checks.md file to include this check?

config/configuration.json
```diff
@@ -964,5 +969,20 @@
...
+    "ocean.valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock",
```
sayas01: nit: this could be 
```
"ocean" : {
     "valid": 
     "invalid":
  }
```

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/OceanBleedingCheckTest.java
```diff
@@ -0,0 +1,61 @@
...
+public class OceanBleedingCheckTest
+{
+    @Rule
+    public OceanBleedingCheckTestRule setup = new OceanBleedingCheckTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    @Test
+    public void invalidBuildingBleedingIntoOcean()
```
sayas01: Could you also add a check with non empty config?

config/configuration.json
```diff
@@ -964,5 +969,20 @@
...
+    "ocean.invalid": "man_made->breakwater,pier|natural->beach,marsh,swamp|water->marsh|wetland->bog,fen,mangrove,marsh,saltern,saltmarsh,string_bog,swamp,wet_meadow|landuse->*",
+    "highway.minimum": "path",
+    "highways.exclude": [
+      "bus_guideway"
+    ],
+    "lineItems.offending": "railway->rail,narrow_gauge,preserved,subway,disused,monorail,tram,light_rail,funicular,construction,miniature",
```
danielduhh: Do offending and "invalid" have the same definitions? If so, we should stick to one naming convention

seancoulter: They're not quite the same--offending refers to tags that must be found on a lineItem for it to be considered bleeding into the ocean. But if it were called valid/invalid I think that would make it less clear if they should be considered or not

Invalid refers to tags that cannot be found on an ocean for that ocean to be considered for the check.

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/OceanBleedingCheckTest.java
seancoulter: Done, see ed33586

config/configuration.json
seancoulter: That's more readable, agreed

config/configuration.json
```diff
@@ -964,5 +969,22 @@
       "difficulty":"NORMAL",
       "defaultPriority":"MEDIUM"
     }
+  },
+  "OceanBleedingCheck": {
+    "ocean": {
+      "valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock",
+      "invalid": "man_made->breakwater,pier|natural->beach,marsh,swamp|water->marsh|wetland->bog,fen,mangrove,marsh,saltern,saltmarsh,string_bog,swamp,wet_meadow|landuse->*"
+    },
+    "highway.minimum": "path",
```
danielduhh: This could be
highway:{
minimum: path,
exclude: []
}

PR #268, opened 2020-03-10
Overlapping edge check bug fix.

smaheshwaram:
### Description:

OverlappingEdgeCheck flags Edges that overlap, but it marks edges as flagged using their OSM ID. This causes an issue when more than one Edge with the same OSM ID has overlapped segments. Only one of the Edges will be flagged, and which one will vary. To fix this Edges should be marked as flagged using their atlas ID.

### Potential Impact:

Changes will be made to Atlas-Checks.

### Unit Test Approach:

Made sure existing tests working as expected.

### Test Results:

Code changes were tested on 50 MB data and analyzed flags on skinny MR to see if every overlapping edge has been flagged.


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
```diff
@@ -182,13 +182,13 @@ private boolean isPartOfClosedWay(final Edge object)
         // Loop through out going edges with the same OSM id
         while (nextEdge != null)
         {
-            wayIds.add(nextEdge.getIdentifier());
+            wayIds.add(nextEdge.getOsmIdentifier());
```
Bentleysb: It looks like the items in `wayIds` should be atlas identifiers, otherwise the search will return true for all multi edge ways.

smaheshwaram: Agreed. Will change it.

PR #271, opened 2020-03-18
Update 6.0.3

maraarae:
### Description:
Updated Atlas versions.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group=org.openstreetmap.atlas
-version=6.0.2-SNAPSHOT
+version=6.0.3-SNAPSHOT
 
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.openstreetmap.atlas
-version=6.0.2-SNAPSHOT
+version=6.0.3-SNAPSHOT
```
sayas01: Isn't this change included in https://github.com/osmlab/atlas-checks/pull/270?

maraarae: Yes, I'm still troubleshooting some issues with git but needed to get this PR out to update the Atlas versions as quickly as possible.

sayas01: Cool! Next time, please make the updates on the latest code by doing a git pull so that we do not make the same changes in multiple PRs.

PR #272, opened 2020-03-23
Add construction configuration

dcastrowa:
### Description:
This added enhancement will allow for the option to check if floating edges are connected to construction roads or not. By default, edges are considered floating if they are connected to construction roads and not connected to any other edges.

### Potential Impact:

No impact if configuration is left as default. If the configuration is changed to true, there will be less edges flagged.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/FloatingEdgeCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/FloatingEdgeCheck.java
@@ -48,6 +48,8 @@ public class FloatingEdgeCheck extends BaseCheck<Long>
     // class variable to store the minimum distance for the floating road
     private final Distance minimumDistance;
     private final HighwayTag highwayMinimum;
+    // check if floating edge is connected to construction road
+    private final boolean checkConstructionRoad;
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/FloatingEdgeCheck.java
```diff
@@ -107,6 +110,16 @@ public FloatingEdgeCheck(final Configuration configuration)
...
         // Consider navigable master edges
+        if (this.checkConstructionRoad) {
+            return TypePredicates.IS_EDGE.test(object) && ((Edge) object).isMasterEdge()
+                    && HighwayTag.isCarNavigableHighway(object) && isMinimumHighwayType(object)
+                    && !intersectsAirport((Edge) object)
+                    && ((Edge) object).connectedNodes().stream().noneMatch(
+                    node -> node.getAtlas()
+                            .linesContaining(node.getLocation(), line -> Validators.isOfType(line,
+                                    HighwayTag.class, HighwayTag.CONSTRUCTION))
+                            .iterator().hasNext());
```
Bentleysb: I think this makes more sense as part of the conditional on line 143.

PR #273, opened 2020-03-23
Making the PGS Tag filter a blank string.

ameliaewatts:
### Description:

Leaving a blank string for the PGS Tag filter. We still want to keep the filter in place, in case we want to filter for PGS Tags again.

### Potential Impact:

We should see an increase in flag counts. The plan is to generate all coastlines and use existing tools (flag db) to filter checks data moving forward.


Bentleysb:
It looks like there is a unit test that will need to be update to reflect this change.

ameliaewatts:
@seancoulter Were you going to be taking that part over? Referring to Bentley's comment.

PR #274, opened 2020-03-23
Leaving a blank string for the PGS Tag filter, under GeneralizedCoast…

ameliaewatts:
…lineCheck, in the config file.

### Description:

Leaving a blank string for the PGS Tag filter. We still want to keep the filter in place, in case we want to filter for PGS Tags again.

### Potential Impact:

We should see an increase in flag counts. The plan is to generate all coastlines and use existing tools (flag db) to filter checks data moving forward.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -256,7 +256,7 @@
   "GeneralizedCoastlineCheck": {
     "node.minimum.distance": 100.0,
     "node.minimum.threshold": 30.0,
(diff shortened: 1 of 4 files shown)


danielduhh:
Looks like this is failing the `oneLineSegmentGeneralizedSourceSurvey` assertion @seancoulter

danielduhh:
nice teamwork!

PR #276, opened 2020-04-03
ShardedIntegrityChecksSparkJob RDD Union Removal

Bentleysb:
### Description:
In earlier designs of the ShardedIntegrityChecksSparkJob it was necessary to gather all UniqueCheckFlagContainers together into a single RDD. This was required to perform a sorting and reduction process to countrify the results. Now that the tasks are parallelized in a countrified manor this is no longer necessary, and causes spark to perform a very large, unnecessary shuffle operation.  

This removes the RDD union in favor of running the flag reduction and file writing operations as part of each countrified stream. This should increase performance for the job on clusters by reducing the amount of communication necessary between executors. 

With the removal of the union it is no longer necessary to persist any RDDs. As such, the spark storage option is no longer needed, and has been removed. 

### Potential Impact:
Should improve performance and stability when running on a cluster. 

### Unit Test Approach:

No changes to unit tests.

### Test Results:
Tested locally on 5 small countries. Results were identical to those produced before the enhancement.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
@@ -2,7 +2,6 @@ package org.openstreetmap.atlas.checks.distributed;
 
 import static org.openstreetmap.atlas.checks.distributed.IntegrityCheckSparkJob.METRICS_FILENAME;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -13,11 +12,9 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import org.apache.spark.api.java.JavaPairRDD;
 import org.apache.spark.api.java.function.PairFunction;
 import org.apache.spark.api.java.function.VoidFunction;
 import org.apache.spark.broadcast.Broadcast;
-import org.apache.spark.storage.StorageLevel;
 import org.openstreetmap.atlas.checks.base.Check;
 import org.openstreetmap.atlas.checks.base.CheckResourceLoader;
 import org.openstreetmap.atlas.checks.configuration.ConfigurationResolver;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -219,46 +204,27 @@ public void start(final CommandMap commandMap)
                             .collect(Collectors.toList());
 
                     // Set spark UI job title
-                    this.getContext().setJobGroup("0",
-                            String.format("Running checks on %s",
-                                    tasksForCountry.get(0).getCountry()));
-                    // Run each task in spark, producing UniqueCheckFlagContainers
-                    final JavaPairRDD<String, UniqueCheckFlagContainer> rdd = this.getContext()
-                            .parallelize(tasksForCountry, tasksForCountry.size())
+                    this.getContext().setLocalProperty("callSite.short", String
```
danielduhh, line 207: What is `callSite.short` ?

Bentleysb: It is a special property that the Spark UI picks up for labeling. 
https://stackoverflow.com/questions/41903342/how-to-change-job-stage-description-in-web-ui

src/main/java/org/openstreetmap/atlas/checks/distributed/ShardedIntegrityChecksSparkJob.java
```diff
@@ -219,46 +204,27 @@ public void start(final CommandMap commandMap)
...
-                                    (Boolean) commandMap.get(MULTI_ATLAS)));
-                    // Save the RDD
-                    rdd.persist(storageLevel);
-                    // Use a fast spark action to overcome spark lazy elocution. This is necessary
-                    // to get the UI title to appear in the right spot.
-                    rdd.count();
-                    // Add the RDDs to the list
-                    countryRdds.add(rdd);
+                                    (Boolean) commandMap.get(MULTI_ATLAS)))
+                            .reduceByKey(UniqueCheckFlagContainer::combine)
```
seancoulter, line 214: What are we combining here, if not by country?

Bentleysb: It is by country, but there is only ever one country. This is just because the country code is passed through the streaming interface by using a pair RDD. The country code needs to be passed to `processFlags` via the RDD due to sparks lazy execution.

seancoulter: Ah that makes sense. Thanks

PR #280, opened 2020-04-14
Add Synthetic tag to Flagged Points

danielduhh:
### Description:

The CheckFlag constructor allows you to add lat/lng Location's to highlight specific areas of flagged features. When creating geojson for these types of CheckFlags, Point features are created with empty properties:
```json
{
	"type": "Feature",
	"geometry": {
		"type": "Point",
		"coordinates": [-63.0593661, 18.198648]
	},
	"properties": {}
}
```
Since these points are not actual OSM features, this can be confusing for downstream tools. To curb this confusion, I've added a `synthetic_highlight_point` (open to other names) tag to the properties of said flags.

```json
{
		"type": "Feature",
		"geometry": {
			"type": "Point",
			"coordinates": [-63.0593661, 18.198648]
		},
		"properties": {
			"synthetic_highlight_point": "yes"
		}
	}
```

### Potential Impact:

### Unit Test Approach:

Added a unit tests to verify the synthetic tag creation.
Ran SpikyBuildingCheck through sharded checks job & zgrep'd the outputs:

Screen Shot 2020-04-14 at 12 40 38 PM

### Test Results:

Unit Tests pass


biaozhang5:
can we change "yes" to a boolean - true?

danielduhh:
Sticking with `“yes”` as the value to follow the synthetic tag convention in atlas: https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/SyntheticBoundaryNodeTag.java#L18

src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedPoint.java
```diff
@@ -25,6 +24,7 @@
  */
 public class FlaggedPoint extends FlaggedObject
 {
+    protected static final String SYNTHETIC_POINT_TAG = "synthetic_highlight_point";
```
Bentleysb, line 27: It would be useful to have this as a tag enum, so other classes could access it (such as the diff commands).

danielduhh: In atlas ?

Bentleysb: As this is very checks specific I think it might make more sense to have it in the Atlas Checks project, though in the same vein as the tags in Atlas. Not sure exactly the best place for it to go in Atlas Check, I would suggest in the flag, constants, or utility packages.

danielduhh: Created a utility.tags package & added this as an enum

PR #282, opened 2020-04-15
LCWB update - multi polygon relation waterbodies

seancoulter:
### Description:

This is an update to LineCrossingWaterBodyCheck to include multipolygon relations with appropriate tagging as valid waterbodies. Tagging constraints are the same as for waterbody Areas. When an edge/Line/Building intersects a M.P. relation, it follows the same rules for flagging as previous, with the additional requirement that at least one intersection must occur at an outer member of the relation. This is to avoid flagging valid intersections between the edge/line/building and islands or landmasses tagged as inner members.

Multipolygon relations found at country borders are not flagged by this check. The way the relation's location is determined with respect to the border is by the SyntheticRelationMemberAdded tag. If this tag has values, then relation members were added during country slicing, which indicates the relation is at a border. The reason we exclude these relations from flagging is because their geometry is altered as a result of country slicing (i.e. the relation gets chopped at the border creating an inaccurate outer polygon), causing improper flagging when a building/line/edge intersects the synthetically-created outer polygon of the relation.

There is also a change to the check flag instruction so that duplicate OSM ids (from edges in the same Way) are not being printed.

### Potential Impact:

LCWB should have more flags to account for features that intersect multi polygon waterbodies. Eventually, as features flagged by InvalidMultipolygonRelationCheck are fixed, we might see more flags with LCWB. This is because LCWB fails on multi polygon waterbodies that are ill formed

### Unit Test Approach:

A unit test involving a multi polygon relation

### Test Results:

Passing


seancoulter:
Analysis for this with the update is in progress

PR #284, opened 2020-04-17
ShadowDetectionCheck Very Large Building Bug

Bentleysb:
### Description:

ShadowDetectionCheck was found to be very slow at processing areas that contain buildings with lots of correctly grounded parts. In this case the check would process all the parts multiple time. 

This fixes this by doing two things. One, all processed parts are marked as flagged so that they are not checked twice. Two, the BFS no longer short circuits upon finding a connection to the ground. Instead all connected parts are collected so that they can be marked as processed.

Example of a building that was extremely slow due to duplicate processing:
https://www.openstreetmap.org/relation/4156588

### Potential Impact:
Decrease the runtime of ShadowDetectionCheck

### Unit Test Approach:

No changes necessary.

### Test Results:

Tested on an area containing the example feature above. Run time was decreased from ~10 hours to ~1 minute. 
Ran on three smaller countries to test for changes in flags. All flags were the same.


danielduhh:
There's a code smell I think you can squash @Bentleysb

src/main/java/org/openstreetmap/atlas/checks/validation/areas/ShadowDetectionCheck.java
```diff
@@ -104,8 +105,8 @@ public boolean validCheckForObject(final AtlasObject object)
     protected Optional<CheckFlag> flag(final AtlasObject object)
     {
         // Gather connected building parts and check for a connection to the ground
-        final Set<AtlasObject> floatingParts = this.getFloatingParts(object);
-        if (!floatingParts.isEmpty())
+        final Tuple<Set<AtlasObject>, Boolean> connectedParts = this.getConnectedParts(object);
+        if (connectedParts.getSecond())
```
Bentleysb, line 109: @danielduhh, the code smell is because `connectedParts.getSecond()` is of type `Boolean` instead of `boolean`. The resolution seems to be to cast it to a `boolean` which seems a bit odd/redundant. Sonar is just worried about it being null, which can't happen in this case.

PR #285, opened 2020-04-20
OceanBleedingCheck update - coastlines and ferries

seancoulter:
### Description:

This is a 2 part enhancement to OceanBleedingCheck:

-- Flag features that intersect natural=coastline LineItems
--Do not flag streets that end at (intersect) ocean boundaries where the node at the location of intersection is tagged with amenity=ferry_terminal.

To accomplish the first part of the enhancement, a new configurable `ocean.boundary` was added which can be used to represent  tags found on ocean/land boundaries. This caused a change to the check's understanding of an ocean feature. Some changes had to be made to account for the separation of the definition of an ocean feature. Below are current definitions for an ocean feature and their flaggable scenarios.

-- A closed area which fully contains water, defined by water tagging. Flaggable items are found within or intersecting the feature.
-- A closed area which fully contains land (where the boundary is tagged natural=coastline). Flaggable items are found intersecting the feature.
-- A Line of land, defined by natural=coastline. Flaggable items are found intersecting the feature.
-- A Line of water, defined by water tagging. Flaggable items are found intersecting the feature.

To accomplish the second part of the enhancement, a check was added to make sure that if a LineItem ends at an ocean feature boundary, that point of intersection must be marked by an amenity=ferry_terminal tag.

Note: there's also a change to IntersectionUtilities.haveExplicitLocationsForIntersections so that it now takes in a PolyLine instead of Polygon. Polyline is Polygon's superclass, and outside of this check it's used by LineCrossingWaterbody and LineCrossingBuilding, which both give Polygons as args, so no refactoring required there.

### Potential Impact:

New configurable ocean.boundary, which should be used to define tags for ocean boundary line features that when crossed by streets/railways/buildings, should be flagged.

Should have an increase in flags due to introducing natural=coastline as a flaggable scenario. Should have a slight reduction of flag count because we no longer flag streets that end with amenity=ferry_terminal at ocean boundaries. Net increase expected.

### Unit Test Approach:

New unit tests

### Test Results:

Passing


src/main/java/org/openstreetmap/atlas/checks/utility/IntersectionUtilities.java
```diff
@@ -73,17 +73,17 @@ public static double findIntersectionPercentage(final Polygon polygon,
...
-     *            {@link Polygon} being crossed
+     *            {@link PolyLine} being crossed
      * @param crossingItem
      *            {@link LineItem} crossing
-     * @return whether given {@link Polygon} and {@link LineItem}'s intersections are actual
+     * @return whether given {@link PolyLine} and {@link LineItem}'s intersections are actual
      *         {@link Location}s for both items
      */
-    public static boolean haveExplicitLocationsForIntersections(final Polygon areaCrossed,
+    public static boolean haveExplicitLocationsForIntersections(final PolyLine areaCrossed,
```
Bentleysb, line 86: nit: areaCrossed -> lineCrossed

danielduhh:
Do you mind updating the documentation? https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/oceanBleedingCheck.md
@seancoulter

src/main/java/org/openstreetmap/atlas/checks/utility/IntersectionUtilities.java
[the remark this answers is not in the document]
seancoulter: @danielduhh Polyline is Polygon's superclass, and outside of this check it's only used by LineCrossingWaterbody and LineCrossingBuilding, which both give Polygons as args, so we should be good

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/OceanBleedingCheck.java
```diff
@@ -37,15 +45,17 @@
...
     private static final String DEFAULT_HIGHWAY_MINIMUM = "TOLL_GANTRY";
     private final HighwayTag highwayMinimum;
     private static final List<String> DEFAULT_HIGHWAYS_EXCLUDE = Collections.emptyList();
     private final List<HighwayTag> highwaysExclude;
-    private static final String OCEAN_INSTRUCTION = "Ocean feature {0,number,#} has invalid intersections.";
-    private static final String BLEEDING_BUILDING_INSTRUCTION = "Building {0,number,#} intersects the ocean feature.";
-    private static final String BLEEDING_LINEITEM_INSTRUCTION = "Way {0,number,#} intersects the ocean feature.";
+    private static final String OCEAN_INSTRUCTION = "Ocean feature {0,number,#} has invalid intersections. ";
+    private static final String BLEEDING_BUILDING_INSTRUCTION = "Building {0,number,#} intersects the ocean feature. ";
+    private static final String BLEEDING_LINEITEM_INSTRUCTION = "Way {0,number,#} intersects the ocean feature. ";
```
danielduhh, line 58: Is it possible to concatenate the ids into a single "Way" instruction? Same for "Building".
That way we can avoid instructions like:

Ocean feature 693833861 has invalid intersections.Way 781473903 intersects the ocean feature.Way 781473903 intersects the ocean feature.Way 781473903 intersects the ocean feature.Way 781473903 intersects the ocean feature.Way 373430804 intersects the ocean feature.Way 373430804 intersects the ocean feature.Way 373430800 intersects the ocean feature.Way 373430800 intersects the ocean feature.

PR #286, opened 2020-04-23
Update FlagDb & Task Deserializer to filter out synthetic point tags

danielduhh:
### Description:

This updates the FlagDB & TaskDeserializer classes to properly filter out synthetic highlight points after #280 was merged.

### Potential Impact:

Tools parsing the flag_id string in the flag table should expect new values

### Unit Test Approach:

Added a unit test to test the updated Flag Deserializer. Also tested the results in a flag db upload.

### Test Results:

Unit tests pass.
Flag DB upload shows `flag_id` values.

Screen Shot 2020-05-11 at 3 51 19 PM


src/main/java/org/openstreetmap/atlas/checks/database/FlagDatabaseSubCommand.java
```diff
@@ -239,7 +240,8 @@ public void executeFlagStatement(final PreparedStatement sql, final CheckFlag fl
     {
         try
         {
-            sql.setString(1, flag.getIdentifier());
+            sql.setString(1, flag.getUniqueIdentifiers().stream().map(String::toString)
+                    .collect(Collectors.joining(",")));
```
Bentleysb, line 244: The unique identifiers are unsorted so in this case it would be a good idea to call `.sorted()` on the stream before collecting it.

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
```diff
@@ -34,11 +39,27 @@ public CheckFlag deserialize(final JsonElement json, final Type typeOfT,
...
+     * Returns a comma delimited string of identifiers.
+     * 
+     * @param identifiers
+     *            - array of flag identifiers
+     * @return - comma delimited string
+     */
+    private String parseIdentifiers(final JsonArray identifiers)
+    {
+        return Arrays.asList(new Gson().fromJson(identifiers, String[].class)).stream()
+                .map(String::toString).collect(Collectors.joining(","));
```
Bentleysb, line 63: Same thing about `.sorted()`.

danielduhh: This should be fixed now

PR #287, opened 2020-04-23
LCWB enhancement - borders,ferries,fords

seancoulter:
### Description:

This PR has a few enhancements for LCWB:
-- Fix flagging at border regions
--Drop flagging for streets that cross waterbodies and have the FordTag
--Drop flagging for streets that have nodes at the location of intersection with a waterbody AND the node has a tag specified by the configurable `nodes.intersecting.non_offending`
--Add flagging for streets that end at waterbodies but are not marked by nodes at such endings, or the node is not one of the configurable tags specified by `nodes.intersecting.non_offending` 

For the border region fix:
Addresses an issue in LCWB where multipolygon relations at country borders were causing improper flagging. Specifically, in https://github.com/osmlab/atlas-checks/pull/282, we dropped flagging completely for multipolygon relations at borders, but this caused several true positive flag drops. So this PR aims to regain those TP while also keeping the false positive rate down by ensuring the intersecting street/railway/building feature intersects an outer polygon of the relation that was NOT created or edited by country slicing.

Two tags, IsoCountryTag and SyntheticNearestNeighborCountryCodeTag, are used to tell if the outer polygon was generated at country slicing time. these tags are synthetically added to atlas-generated features from country slicing. If an atlas feature has more than these two tags, it shouldn't have been created from country slicing. So only streets/buildings/railways that intersect features with more than the two tags should be flagged. If there is no such intersection, we can still flag the intersecting feature if it is entirely contained in the outer polygon, and does not intersect and is not contained by any inner member of the relation. Same IsoCountry/SyntheticNearestNeighborCountryCodeTag rules apply here as well

### Potential Impact:

Better flagging for LCWB on multipolygon relations at borders
No flagging for edges that end at OR cross (ford) waterbodies validly. Valid tags for streets that end at waterbodies defined in nodes.intersecting.non_offending.
More flags for edges that end at waterbodies but don't have proper tagging.

### Unit Test Approach:

Updated and added  unit tests

### Test Results:

passing

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
@@ -17,6 +17,7 @@ import org.openstreetmap.atlas.checks.flag.CheckFlag;
 import org.openstreetmap.atlas.checks.utility.IntersectionUtilities;
 import org.openstreetmap.atlas.geography.GeometricSurface;
 import org.openstreetmap.atlas.geography.MultiPolygon;
+import org.openstreetmap.atlas.geography.PolyLine;
 import org.openstreetmap.atlas.geography.Polygon;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.items.Area;
@@ -37,12 +38,12 @@ import org.openstreetmap.atlas.tags.NaturalTag;
 import org.openstreetmap.atlas.tags.NotesTag;
 import org.openstreetmap.atlas.tags.PlaceTag;
 import org.openstreetmap.atlas.tags.SourceTag;
-import org.openstreetmap.atlas.tags.SyntheticRelationMemberAdded;
 import org.openstreetmap.atlas.tags.Taggable;
 import org.openstreetmap.atlas.tags.annotations.validation.Validators;
 import org.openstreetmap.atlas.tags.filters.TaggableFilter;
 import org.openstreetmap.atlas.utilities.collections.MultiIterable;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.tuples.Tuple;
 
 /**
  * Flags line items (edges or lines) and optionally buildings that are crossing water bodies
@@ -217,12 +218,10 @@ public class LineCrossingWaterBodyCheck extends BaseCheck<Long>
     @Override
     public boolean validCheckForObject(final AtlasObject object)
     {
-        // We only consider water body areas or multipolygon relations that are not at borders, not
-        // linear water bodies
-        return (TypePredicates.IS_AREA.test(object) || (object instanceof Relation
-                && object.getTag(SyntheticRelationMemberAdded.KEY).isEmpty()
-                && ((Relation) object).isMultiPolygon())) && !INVALID_WATER_BODY_TAGS.test(object)
-                && VALID_WATER_BODY_TAGS.test(object);
+        // We only consider water body areas or multipolygon relations, not linear water bodies
+        return (TypePredicates.IS_AREA.test(object)
+                || (object instanceof Relation && ((Relation) object).isMultiPolygon()))
+                && !INVALID_WATER_BODY_TAGS.test(object) && VALID_WATER_BODY_TAGS.test(object);
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -329,29 +335,61 @@ private boolean intersectionsAreExplicit(final GeometricSurface waterbody,
...
+                .filter(member -> member.getRole().equals("outer"))
+                .map(member -> new Tuple<>(member.getEntity(),
+                        new Polygon(member.getEntity() instanceof Area ? (Area) member.getEntity()
+                                : (LineItem) member.getEntity())))
+                .anyMatch(member -> (intersectingFeature.intersects(member.getSecond())
+                        || (intersectingFeature.within(member.getSecond())
+                                && ((MultiPolygon) waterbody).inners().stream().noneMatch(
+                                        innerMember -> intersectingFeature.intersects(innerMember)
+                                                || intersectingFeature.within(innerMember))))
+                        && member.getFirst().lastEdit().isPresent());
```
danielduhh, line 371: While this seems like it works -- this solution feels a bit hacky. We've ran into this issue before and proposed parsing the identifier to verify whether or not the feature has been sliced at the border -- this too felt like a hack. 
I would defer to atlas to provide a better (reusable) solution to verify if a feature has been country sliced -- as its shown to be useful in other checks.

seancoulter: I agree that this should definitely be in Atlas. However, just to get a short term solution/discussion going, I did a minor change on this logic to be more in line with how country slicing works- we're now making sure the relation member has any tag other than the IsoCountryTag and SyntheticNearestNeighborCountryCodeTag. These two tags are synthetically added to a relation member that's been sliced at the border. The new logic should validate the same features as the previous method since all validated members should still have the LastEditTimeTag/other OSM specific tags.

This change is in 28a3ce6

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -329,29 +368,63 @@ private boolean intersectionsAreExplicit(final GeometricSurface waterbody,
                 .haveExplicitLocationsForIntersections(polygon, lineItem));
     }
 
+    /**
+     * If the waterbody is an area, the intersection is valid. If it's not an area, it's a
+     * multipolygon, and we make sure that either 1. The intersection is at an outer member of the
+     * multipolygon and the outer member was not synthetically created by the atlas slicing process.
+     * It's estimated that if the outer member doesn't have the {@link LastEditTimeTag}, it likely
```
danielduhh, line 375: I believe this comment is out of date

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -293,31 +302,61 @@ public boolean validCheckForObject(final AtlasObject object)
...
+        }
+
+        // All intersections between the street and waterbody
+        final Set<Location> intersections = waterbody instanceof Polygon
+                ? ((Polygon) waterbody).intersections(((Edge) crossingItem).asPolyLine())
+                : ((MultiPolygon) waterbody).outers().stream()
+                        .flatMap(polygon -> polygon
+                                .intersections(((Edge) crossingItem).asPolyLine()).stream())
+                        .collect(Collectors.toSet());
+        final Predicate<AtlasItem> hasFordTag = item -> !item.getTag(FordTag.KEY)
```
danielduhh, line 332: Can the ford use case addition be added to the `lines.non_offending` configurable list?

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheckTestRule.java
```diff
@@ -461,7 +461,8 @@
             @Node(coordinates = @Loc(value = AREA_LOCATION_3)),
             @Node(coordinates = @Loc(value = LOCATION_OUTSIDE_AREA_1)),
             @Node(coordinates = @Loc(value = AREA_LOCATION_BETWEEN_2_AND_3)) }, areas = {
-                    @Area(id = "100", coordinates = { @Loc(AREA_LOCATION_1), @Loc(AREA_LOCATION_2),
+                    @Area(id = "100", tags = "last_edit_time=20200422", coordinates = {
```
danielduhh, line 464: This could be updated to now check for
- synthetic nearest neighbor tag
- missing iso tag

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -329,29 +368,63 @@ private boolean intersectionsAreExplicit(final GeometricSurface waterbody,
...
+     * @param object
+     *            the AtlasEntity representation of the waterbody
+     * @param intersectingFeature
+     *            the intersecting building/road/line
+     * @return true if either of the 2 above conditions are satisfied, false otherwise
+     */
+    private boolean intersectsValidPartOfWaterbody(final GeometricSurface waterbody,
+            final AtlasObject object, final PolyLine intersectingFeature)
+    {
+        return !(waterbody instanceof MultiPolygon) || ((Relation) object).members().stream()
```
danielduhh, line 394: 🤯

danielduhh: I will leave this up to you but this is a bit hard to follow 😄

seancoulter: Added some comments and some clarification in the method docstring.. https://github.com/osmlab/atlas-checks/pull/287/commits/ec536cc2f39f151ad87b4e767e0316fa5ba4818f#diff-829f2e37322d264fb3b26e180abc731eR380

danielduhh: Awesome! Thanks for the updates

seancoulter:
Diffs for the ferry/ford/leisure enhancement (nodes.intersecting.non_offending) have been validated. This is ready for re-review now

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -329,29 +380,72 @@ private boolean intersectionsAreExplicit(final GeometricSurface waterbody,
...
+                        .anyMatch(member -> (intersectingFeature.intersects(member.getSecond())
+                                || (intersectingFeature.within(member.getSecond())
+                                        && ((MultiPolygon) waterbody).inners().stream()
+                                                .noneMatch(innerMember -> intersectingFeature
+                                                        .intersects(innerMember)
+                                                        || intersectingFeature
+                                                                .within(innerMember))))
+                                && (member.getFirst().getTags().keySet().stream().anyMatch(
+                                        key -> !(key.equals(ISOCountryTag.KEY) || key.equals(
+                                                SyntheticNearestNeighborCountryCodeTag.KEY)))));
```
Bentleysb, line 427: I think the use of SyntheticNearestNeighborCountryCodeTag in country slicing has been removed by https://github.com/osmlab/atlas/pull/627.

seancoulter: Good catch. Taking that out so we now only check if the feature has a tag other than the IsoCountryTag

seancoulter: According to https://github.com/osmlab/atlas/blob/f9ebfa344fb5891e3a21a2e790ea3f153230d4ec/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java#L442, synthetically added relation members at borders would only have the IsoCountryTag tag. So checking for any tag other than that is looking like a good/straightforward estimation for whether or not the feature was synthetically created, especially considering that features not synthetically created would have OSM tags as well.

danielduhh: nice! thanks for looking into it

danielduhh: So to clarify.. if the feature _only_ contains a iso country code, we can assume it was synthetically added.

seancoulter: as far as multipolygon relation members go, that's the idea.

PR #292, opened 2020-05-06
Update FlaggedRelation to Output Atlas Identifier and OSM Identifier as String

sayas01:
### Description:

For all FlaggedObjects, except FlaggedRelation, the "osmIdentifer" and "identifier" fields in the feature properties are string. In case of FlaggedRelation, these fields are populated as integers. This PR updates the "asGeoJsonFeature" method in FlaggedRelation to populate the  "osmIdentifer" and "identifier" fields as string instead of as integers.

The "identifier" and "osmIdentifier" keys in the properties of geojson features for FlaggedRelation will be string just like in other FlaggedObjects.

### Unit Test Approach:

Locally ran the updates on checks that flag relation and verified the geojson and log files

### Test Results:
Before updates:

After updates:

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedRelation.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedRelation.java
@@ -3,6 +3,7 @@ package org.openstreetmap.atlas.checks.flag;
 import java.util.Map;
 import java.util.Optional;
 
+import com.google.gson.JsonElement;
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.Location;
 import org.openstreetmap.atlas.geography.MultiPolygon;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedRelation.java
```diff
@@ -64,10 +67,15 @@ public FlaggedRelation(final Relation relation, final MultiPolygon geoJsonGeomet
     public JsonObject asGeoJsonFeature(final String flagIdentifier)
     {
         final JsonObject featureProperties = this.relation.getGeoJsonProperties();
+        final JsonElement osmIdentifier = featureProperties.get(OSM_IDENTIFIER_TAG);
+        final JsonElement identifier = featureProperties.get(IDENTIFIER);
+        featureProperties.remove(OSM_IDENTIFIER_TAG);
```
danielduhh, line 72: Do you mind adding a comment here on why we're deleting and adding the same properties/values ?

sayas01: @danielduhh I have added the comment.

PR #296, opened 2020-05-12
Enhancement to InvalidTagsCheck To Use Resource File For Filters

sayas01:
### Description:

This PR adds the following updates to the existing InvalidTagsChecks:

1) Updated the check to also handle filters passed through resource file
2) Addition of "filters.resource.override" and "filters.resource.append" config values. 
   
- If "filters.resource.override" is true and "filters.resource.append" is false, only the filters passed through the config ("filters.classes.tags") will be used to test the features to flag.

-  If "filters.resource.override" is false and  "filters.resource.append" is true, filters passed through the config will be added to default list of filters from the resource file.

 
### Potential Impact:
No update to existing check as the default filters are not changed. Just an enhancement to allow addition of filters through resource file.

### Unit Test Approach:

Updated and added new unit tests. Ran the check in different countries and analyzed the results.

### Test Results:

All unit tests passing. Analysis of the updates shows 0% FP.


docs/checks/invalidTagsCheck.md
```diff
@@ -1,17 +1,28 @@
 # Invalid Tags Check
 
-This check flags features based on configurable filters.   
+This check flags features for invalid tags based on a list of TaggableFilters. The list of TaggableFilters for this check is either a combination of filters passed through the 
+check configuration and default list of TaggableFilters or just the default filters. 
+The two configurable boolean values, "override.default.filters" and "append.to.default.filters", 
+determine the list of taggable filters that each of the atlas entities will be tested against. The rules are
+1) If "override.default.filters" is true and "append.to.default.filters" is false,
+only the filters that are passed through the configuration file will be used to test the atlas entities. 
+2) If "append.to.default.filters" is true and "override.default.filters" is false,
```
Bentleysb, line 9: If these filters are mutually exclusive then why not just have one switch?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
```diff
@@ -37,21 +52,87 @@
...
      */
-    private static Set<String> getFilterKeys(final TaggableFilter filter)
+    private static Set<String> getFilterKeys(final AtlasObject atlasObject,
+            final List<TaggableFilter> listOfFilters)
     {
-        return Arrays.stream(filter.toString().split("[|&]+"))
-                .filter(string -> string.contains(KEY_VALUE_SEPARATOR))
-                .map(tag -> tag.split(KEY_VALUE_SEPARATOR)[0]).collect(Collectors.toSet());
+        final Set<String> collectedKeys = new HashSet<>();
+        listOfFilters.stream().filter(taggableFilter -> taggableFilter.test(atlasObject)).forEach(
```
Bentleysb, line 104: This method was originally made to return the keys from a single filter so that an instruction could be made for each filter that is passed. Now this can return the combined keys of multiple filters, which could make the instructions more confusing.

docs/checks/invalidTagsCheck.md
sayas01: The check originally had empty default filters and I wanted to keep it that way and so used two switches. Since list of new filters added is long and couldn't be appended to the config file, I saved them in a resource file and made use of these two switches to either use the filters in the resource or add the config filters to the ones in resource.

danielduhh:
thoughts @sayas01 ?

sayas01:
@danielduhh The new updates make it possible for us to choose one of the below three options

-  just use the config filters
- just use the filters in the resource file
- use both the filters in the config as well as the resource file.

I added it this way so that it is easy to maintain and gives us flexibility to add new filters without having to change the resource file.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
sayas01: Hmm yeah. I am filtering the list of filters here so that only the keys of the filter which the atlasobject passes will be used in the instruction. But I think it makes sense to group the keys in the instruction to avoid confusion.

docs/checks/invalidTagsCheck.md
Bentleysb: I think if we are adding a resource then it makes sense to have that as the default, but I like having the option to overwrite or append to it.

sayas01: Yeah the only reason these filters are not in the config is because there are many. These are solely water-island related tag combinations and need not be default filters.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
sayas01: @Bentleysb I have updated the logic to pass taggable filter, instead of list of filters, to this method so that for each filter, a new instruction is created just as before.

config/configuration.json
```diff
@@ -388,8 +388,11 @@
     }
   },
   "InvalidTagsCheck": {
+    "override.default.filters": false,
```
danielduhh, line 391: This could be:
```json
"filters.defaultt": {
 "override": false,
 "append": true
}
```

src/main/resources/org/openstreetmap/atlas/checks/validation/tag/invalid-tags-check-filter.json
```diff
@@ -0,0 +1,52 @@
+{
+  "filters": [
+    "island->*&natural->spring,hot_spring&name->*",
```
danielduhh, line 3: If we go the resource route, the default values should be the tags in the original version

sayas01: @danielduhh https://github.com/osmlab/atlas-checks/pull/296/commits/96e197a3683865b16b6b502b8a558d0714fa312f should address this.

danielduhh:
@Bentleysb this good to go?

PR #297, opened 2020-05-12
InvalidMultiPolygonRelationCheck Geometry Enhancement

Bentleysb:
### Description:
This adds a number of new possible conditions to InvalidMultiPolygonRelationCheck  for flagging a feature. 

All of the new conditions are to check for invalid geometry combinations:
* Has any inner members that are not contained by an outer member
* Has any outer members that overlap
* Has any inner members that overlap with another inner member, but just touching is fine
* Has any inner relations that intersect an outer member

These combinations are all defined as invalid by the [OSM Wiki](https://wiki.openstreetmap.org/wiki/Relation:multipolygon)

Two other smaller changes are also included.
A configurable switch has been added to exclude one member relations from this check. This is meant to be used when running this check with OneMemberRelationCheck to prevent duplicate flags.
The `flag` method has been heavily refactored to reduce the cognitive complexity, as sonar was complaining. 

### Potential Impact:

Flags for this check are expected to fluctuate. There are large drops when the new switch is active, and lower numbers of additions from the new conditions. 

### Unit Test Approach:

Added new unit test for all the new conditions and the new switch. 

### Test Results:

Sampled flags from 7 countries. All results were as expected.


src/main/java/org/openstreetmap/atlas/checks/validation/relations/InvalidMultiPolygonRelationCheck.java
```diff
@@ -185,43 +161,254 @@ else if (!relationMember.getRole().equals(RelationTypeTag.MULTIPOLYGON_ROLE_INNE
...
+     */
+    private Set<Tuple<Polygon, Polygon>> checkInnerOverlap(
+            final MultiMap<Polygon, Polygon> outerToInners)
+    {
+        final Set<Tuple<Polygon, Polygon>> problematicPolygons = new HashSet<>();
+
+        outerToInners.entrySet().forEach(entry ->
+        {
+            // Loop through each combination of inner polygons only once to check for overlap
+            for (int index1 = 0; index1 < entry.getValue().size() - 1; index1++)
```
seancoulter, line 261: 👍

src/main/java/org/openstreetmap/atlas/checks/validation/relations/InvalidMultiPolygonRelationCheck.java
```diff
@@ -185,43 +161,254 @@ else if (!relationMember.getRole().equals(RelationTypeTag.MULTIPOLYGON_ROLE_INNE
...
-                        relation, openLocations);
+                logger.warn(
+                        "Unable to find members in multipolygonRelation {} containing the locations : {}",
+                        multipolygonRelation, openLocations);
+            }
+        }
+        catch (final CoreException exception)
+        {
+            // Catch multipolygon relations with no outer members
+            if (exception.getMessage().equals("Unable to find outer polygon."))
```
sayas01, line 213: Not sure if this is the best way to assign instructions for the different cases. The instructions could get changed in the future. Something to look out for.

Bentleysb: It's not ideal, but if it does change it should be caught by the unit tests.

src/main/java/org/openstreetmap/atlas/checks/validation/relations/InvalidMultiPolygonRelationCheck.java
```diff
@@ -185,43 +161,254 @@ else if (!relationMember.getRole().equals(RelationTypeTag.MULTIPOLYGON_ROLE_INNE
...
+                        Collections.singleton(this.getLocalizedInstruction(
+                                CLOSED_LOOP_INSTRUCTION_FORMAT_INDEX,
+                                multipolygonRelation.getOsmIdentifier(), memberIds, openLocations)),
+                        new HashSet<>(openLocations)));
             }
             else
             {
-                logger.warn("Unable to find members in relation {} containing the locations : {}",
-                        relation, openLocations);
+                logger.warn(
```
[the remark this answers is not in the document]
Bentleysb: This was just part of the refactor, not a new addition, so I had not really looked into that. I think you're right that it probably should be flagged. However, I would need to do some investigation to make sure it was not done like this to ignore atlas data errors, or something of that sort. I think that might be good to investigate outside this PR.

PR #304, opened 2020-06-02
InvalidGeometryCheck

Bentleysb:
### Description:

This is a new check to flag Ways with invalid geometries. 
It uses the Java Topology Suite (JTS) to test geometries for validity.

The check uses the GeometryValidator class to hadle JTS conversion and validation.
The class has been refactored to return details about invalid geometries. These are then used in the check instructions. 

An Atlas version bump is included as part of this PR. This is required to bring new synthetic tagging that is used to filter out false positive cases caused by Atlas country slicing. 

#### OSM Example
The Way [id:803496316](https://www.openstreetmap.org/way/803496316) would be flagged by this check because it is a self intersecting polygon. 

### Potential Impact:

The changes to the GeometryValidator class are breaking, requiring a small change in syntax to retain the current functionality. 

### Unit Test Approach:

Unit test cover a variety of invalid polyline and polygon cases.

### Test Results:

The check was tested in 7 small to medium countries, producing 46 flags. All flags were examined and found to be valid.

### DIFF
--- a/docs/available_checks.md
+++ b/docs/available_checks.md
@@ -89,6 +89,7 @@ This document is a list of tables with a description and link to documentation f
 | [~~DuplicateWaysCheck~~ (Deprecated)](checks/duplicateWaysCheck.md) | The purpose of this check is to identify Ways that have either had their entire length or a segment of their length duplicated or drawn multiple times. **This check has been deprecated and is no longer active.** |
 | [GeneralizedCoastlineCheck](checks/generalizedCoastlineCheck.md) | The purpose of this check is to identify coastlines whose nodes are too far apart, have angles that are too sharp, and/or have _source=PGS_ Tag values. |
 | [IntersectingBuildingsCheck](checks/intersectingBuildingsCheck.md) | The purpose of this check is to identify buildings that intersect other buildings. |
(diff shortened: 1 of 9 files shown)


docs/available_checks.md
```diff
@@ -89,6 +89,7 @@ This document is a list of tables with a description and link to documentation f
 | [~~DuplicateWaysCheck~~ (Deprecated)](checks/duplicateWaysCheck.md) | The purpose of this check is to identify Ways that have either had their entire length or a segment of their length duplicated or drawn multiple times. **This check has been deprecated and is no longer active.** |
 | [GeneralizedCoastlineCheck](checks/generalizedCoastlineCheck.md) | The purpose of this check is to identify coastlines whose nodes are too far apart, have angles that are too sharp, and/or have _source=PGS_ Tag values. |
 | [IntersectingBuildingsCheck](checks/intersectingBuildingsCheck.md) | The purpose of this check is to identify buildings that intersect other buildings. |
+| [InvalidGeometryCheck](checks/invalidGeometryCheck.md) | This check flags invalid polyline and polygon geometries. |
```
danielduhh, line 92: Thanks for adding this 🙏

Issue #302, opened 2020-05-28
[Bug] Incorrect behavior of isPartOfClosedWay method in OverlappingEdgeCheck


vladlemberg:
**Describe the bug**
Atlas check -> OverlappingEdgeCheck
Method -> isPartOfClosedWay
stream(nextEdge.outEdges()) returns both Master and negative edges. 
To fix the issue add .filter(Edge::isMasterEdge)

**To Reproduce**
Pass OSM Way: 97245856 
https://www.openstreetmap.org/way/97245856

vladlemberg:
fixed by filtering Master Edges

vladlemberg:
fixed. unit test passed.

PR #305, opened 2020-06-03
fix traversing via reversed edges.

vladlemberg:
### Description:
to fix bug https://github.com/osmlab/atlas-checks/issues/302 
Details: eliminate traversing via reversed Edges

### Potential Impact:

### Unit Test Approach:

Describe unit tests being added with that PR.

### Test Results:

Describe other (non-unit) test results here.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
@@ -128,8 +128,8 @@ public abstract class BaseCheck<T> implements Check, Serializable
         }
         catch (final Exception oops)
         {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -128,8 +128,8 @@ public BaseCheck(final Configuration configuration)
         }
         catch (final Exception oops)
         {
-            logger.error(String.format("%s failed on feature %s.", this.getCheckName(),
-                    object.getIdentifier()), oops);
+            logger.error(String.format("%s failed on feature %s (%s).", this.getCheckName(),
```
sayas01, line 131: @vladlemberg could you also mention this change in the PR description? Thanks!

PR #308, opened 2020-06-08
ApproximateWayCheck

brianjor:
### Description
This checks for ways that were done crudely, there is a discrepancy between the drawing and the real way especially in the curve. Roads are expected to be smooth and consistent, large changes in directions are suspect of poor drawing. Fixing these visually improves the map.

### Requirements
* Is an Edge
* Has a highway tag that is car navigable
* Edge has more than one segment

### Use Cases
Case 1: False Positive (US)

* This gets flagged due to the sharp angle in the edge, this edge is the same road throughout and matches the imagery
* https://www.openstreetmap.org/way/608161473

image

Case 2: True Positive (US)

* Curves are not smooth, poorly drawn
* https://www.openstreetmap.org/way/40195903

image

### Test Results
| Country | Total Ways Checked | False Positives | FP % |
|----------|:---------------------:|:---------------:|:-------:|
|   AUS     |              261            |           24         |  9.2%  |
|  TUR      |              274            |           29         | 10.6% |

Also tested on AIA, CYM,SMR but no flags

### Supported regions
All Countries

### References
https://wiki.openstreetmap.org/wiki/Osmose/issues#1190

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -4,54 +4,10 @@
       "org.openstreetmap.atlas.checks.validation"
     ],
     "type": "org.openstreetmap.atlas.checks.base.BaseCheck",
-    "enabled.value.default": true
-  },
-  "PoolSizeCheck": {
-    "surface": {
-      "maximum": 1000.0,
-      "minimum": 50.0
-    }
-  },
-  "SpikyBuildingCheck": {
-    "spiky.angle.maximum": 15.0,
-    "curve": {
-      "degrees": {
-        "maximum.single_heading_change": 25.0,
-        "minimum.total_heading_change": 10.0
-      },
-      "points.minimum": 4
-    },
-    "challenge": {
-      "description": "Find poorly digitized buildings with sharp angles in their geometry",
-      "blurb": "Fix buildings with spiky geometry",
-      "instruction": "Open your favorite editor and validate that the buildings are mapped correctly",
-      "difficulty": "NORMAL",
-      "tags":"building"
-    }
-  },
-  "OrphanNodeCheck": {
-
-  },
-  "AddressPointMatchCheck": {
-    "bounds.size": 150.0,
-    "challenge": {
-      "description": "Tasks contain nodes which either have no street name or incorrect street names",
-      "blurb": "Nodes without street names, or with incorrect street names",
-      "instruction": "Open your favorite editor and edit the node street names",
-      "difficulty": "EASY",
-      "defaultPriority": "LOW"
-    }
-  },
-  "AddressStreetNameCheck": {
-    "bounds.size": 100.0,
-    "challenge": {
-      "description": "Tasks contain nodes with addr:street names that don't match the surrounding roads",
-      "blurb": "Nodes with mismatched addr:street names",
-      "instruction": "Open your favorite editor and edit the node street names",
-      "difficulty": "HARD"
-    }
+    "enabled.value.default": false
   },
   "ApproximateWayCheck": {
+    "enabled": true,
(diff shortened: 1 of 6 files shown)


config/configuration.json
```diff
@@ -51,6 +51,12 @@
       "difficulty": "HARD"
     }
   },
+  "ApproximateWayCheck": {
+    "deviation.minimum.meters": 35.0,
+    "angle.minimum": 100.0,
+    "bezierStep": 0.01,
+    "highway.minimum": "service"
+  },
```
Bentleysb, line 59: Could you please add a `challenge` object to this configuration. You can look at other checks' configurations for examples. This information is used when creating MapRoulette challenges.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ApproximateWayCheck.java
```diff
@@ -0,0 +1,198 @@
...
+    {
+        final double startX = start.getLongitude().onEarth().asMeters();
+        final double startY = start.getLatitude().onEarth().asMeters();
+        final double anchorX = anchor.getLongitude().onEarth().asMeters();
+        final double anchorY = anchor.getLatitude().onEarth().asMeters();
+        final double endX = end.getLongitude().onEarth().asMeters();
+        final double endY = end.getLatitude().onEarth().asMeters();
+
+        double min = Double.POSITIVE_INFINITY;
+        for (double i = 0; i <= 1; i += bezierStep)
```
sayas01, line 183: nit: Please avoid one letter variables.

docs/checks/approximateWayCheck.md
```diff
@@ -0,0 +1,10 @@
+# ApproximateWayCheck
+
+#### Description
```
sayas01, line 3: Could you add a section to include examples? You can refer any of the existing docs for the details. Thanks!

docs/checks/approximateWayCheck.md
```diff
@@ -0,0 +1,10 @@
...
+
+The purpose of this check is to identify ways that are crudely drawn, there is a discrepancy between the drawing and the real way, especially for curves.
+
+#### Code Review
```
Bentleysb, line 7: It would be great if you could add a brief description of some off the more complex logic here. I think an explanation of your findAngle and quadraticBezier methods would be useful.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ApproximateWayCheck.java
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * This check flags edges that deviate from the assumed curve of a road by at least
+ * {@value DEVIATION_MINIMUM_LENGTH_DEFAULT} meters.
+ *
+ * @author v-brjor
+ */
+public class ApproximateWayCheck extends BaseCheck<Long>
+{
+
+    private static final long serialVersionUID = 1L;
```
sayas01, line 31: Please regenerate the UID.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ApproximateWayCheck.java
```diff
@@ -0,0 +1,216 @@
...
+    private static final String EDGE_DEVIATION_INSTRUCTION = "Way {0,number,#} is crude";
```
sayas01, line 32: The instruction could be more descriptive like "Way {0,number,#} is crude.  Add more nodes/rearrange current nodes".

PR #313, opened 2020-06-11
Conditional restriction check

lauracasiana:
### Description:

Added a new check that verifies the conditional restriction format. 

### Potential Impact:

No impact to existing code. 

### Unit Test Approach:

Added unit tests for possible situations the check might encounter in the ConditionalRestrictionCheckTest class. 

### Test results:

Tested on the following countries: AIA, AUS, CYM, DMA, NZL, TUR
Flags resulted only from  AUS - 45 different instruction cases. 
More details in the file below: 
[Conditional Restriction check validation.xlsx](https://github.com/osmlab/atlas-checks/files/4865508/Conditional.Restriction.check.validation.xlsx)


lauracasiana:
Example run on AUS: 
[ConditionalRestrictionCheck-1592398794894-561.geojson.zip](https://github.com/osmlab/atlas-checks/files/4796666/ConditionalRestrictionCheck-1592398794894-561.geojson.zip)

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final List<String> conditionalKeys = object.getOsmTags().keySet().stream()
+                .filter(key -> key.contains(CONDITIONAL)).collect(Collectors.toList());
+        for (final String key : conditionalKeys)
+        {
+            if (!isKeyValid(key))
+            {
+                return Optional.of(this.createFlag(object, this.getLocalizedInstruction(0, key)));
```
Bentleysb, line 81: By using a return statement here the check may miss flagging the value if it is also bad, and flagging other tags if an object has multiple bad conditional tags. Thus it could take multiple cycles of running the check and editing the feature to fully fix the problems. It would probably be better to add the instruction to a list of generated instructions for this feature, and flag the feature later if that list is not empty. You can see an example of this sort of arrangement in [RoundaboutValenceCheck](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutValenceCheck.java).

lauracasiana:
Run results after changes: 
[Conditional Restriction results.zip](https://github.com/osmlab/atlas-checks/files/4864018/Conditional.Restriction.results.zip)
Validation results: 
[Conditional Restriction check validation.xlsx](https://github.com/osmlab/atlas-checks/files/4864027/Conditional.Restriction.check.validation.xlsx)

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,223 @@
...
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Set<String> instructions = new HashSet<>();
+
+        final List<String> conditionalKeys = object.getOsmTags().keySet().stream()
+                .filter(key -> key.contains(CONDITIONAL)).collect(Collectors.toList());
+        for (final String key : conditionalKeys)
+        {
+            if (!isKeyValid(key))
```
sayas01, line 85: Please add `this` keyword when accessing the methods.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    private boolean containsTransportationMode(final String key)
+    {
+        final String[] parts = key.split(":");
```
sayas01, line 113: The CommonConstants class cab be leveraged here: https://github.com/osmlab/atlas-checks/blob/d070ac474ff5751c12b98fde8fe4620a3d8ae131/src/main/java/org/openstreetmap/atlas/checks/constants/CommonConstants.java#L28

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,223 @@
...
+            "dismount", "agricultural", "forestry", "discouraged", "official", "lane",
+            "share_busway", "opposite_share_busway");
+
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            "The conditional key \"{0}\" does not respect the \"<restriction-type>[:<transportation mode>][:<direction>]:conditional\" format",
+            "The conditional value \"{0}\" does not respect the format \"<restriction-value> @ <condition>[;<restriction-value> @ <condition>]\" ",
+            "The element with id {0,number,#} does not follow the conditional restriction pattern.");
+    public static final int TWO_PARTS = 2;
+    public static final int THREE_PARTS = 3;
+    public static final int FOUR_PARTS = 4;
```
sayas01, line 55: Any reason why these are set to access public?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private boolean isTransportationMode(final String value)
+    {
+        return TRANSPORTATION_MODE.contains(value);
+    }
+
+    private boolean isValueValid(final String value, final String key)
+    {
+        boolean isValid = true;
+        final Pattern valuePattern = Pattern.compile(
```
sayas01, line 192: This could be made static to avoid recompilation for each instance of the class.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/ConditionalRestrictionCheck.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    private boolean isTransportationMode(final String value)
+    {
+        return TRANSPORTATION_MODE.contains(value);
+    }
+
+    private boolean isValueValid(final String value, final String key)
+    {
+        boolean isValid = true;
```
sayas01, line 191: I think this variable is not required. We can simply return true/false based on the condition. For example: ` if (!trimmedPart.isEmpty() && !isAccessValue(trimmedPart))
                        {
                            return false;
                        }`

Issue #306, opened 2020-06-03
[New Check] Identify missing roundabouts


vladlemberg:
**Description**
This check identifies potential roundabouts that are not properly tagged.

**Requirements**
This check is to look for suspicious circular junctions that may be roundabout without the “junction=roundabout” tag. 

1. Check Candidate (example)
a) Closed way e.g. start and end location shared same lat/lon 
b) Intersecting at least with two navigable roads 
c) circular shape: angle between sections is > 10 or < 40 degree

**Use Cases**
This section summarizes the various use cases for this check and provides potential enhancement suggestions if needed. The uses cases should act as a test in a way, such that when the analysis/enhancement is completed, these types of use cases should be found or eliminated.

For each use case, please add a screenshot of the feature & openstreetmap.org link (or other sources). If necessary, edit the screenshot to better illustrate the use case.

Example:
tbd

**Supported regions**
globally

PR #315, opened 2020-06-11
Vlad missing roundabout tag

vladlemberg:
### Description:
Develop new Atlas check: https://github.com/osmlab/atlas-checks/issues/306

### Potential Impact:

N/A

### Unit Test Approach:

Unit Test comprises: Positive, False positive and Negative cases. 

### Test Results:

passed


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,196 @@
...
+        while (nextEdge != null)
+        {
+            wayIds.add(nextEdge.getIdentifier());
+            final List<Edge> nextEdgeList = Iterables.stream(nextEdge.outEdges())
+                    .filter(Edge::isMasterEdge)
+                    .filter(outEdge -> outEdge.getOsmIdentifier() == edge.getOsmIdentifier())
+                    .collectToList();
+            nextEdge = nextEdgeList.isEmpty() ? null : nextEdgeList.get(0);
+            // If original edge is found, the way is closed
+            if (nextEdge != null && wayIds.contains(nextEdge.getIdentifier()))
```
Bentleysb, line 189: Because you know you are dealing with the first edge section, you could just check that the in Edges contain an edge with the same OSM id. That would save you having to traverse along the whole Way.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,196 @@
...
+            final List<Tuple<Angle, Location>> maxOffendingAngles = originalGeom
+                    .anglesGreaterThanOrEqualTo(this.maxAngleThreshold);
+            // check minimum angle
+            final List<Tuple<Angle, Location>> minOffendingAngles = originalGeom
+                    .anglesLessThanOrEqualTo(this.minAngleThreshold);
+
+            if (maxOffendingAngles.isEmpty() && minOffendingAngles.isEmpty())
+            {
+                this.markAsFlagged(object.getOsmIdentifier());
+                return Optional.of(createFlag(new OsmWayWalker(edge).collectEdges(),
```
Bentleysb, line 106: You use the way walker 3 times to collect the entire way (here, in buildOriginalOsmWayGeometry, and in intersectingWithMoreThan). While this is not a super expensive function, its repeated use can add up. It might be a bit better to call it once and then pass the edge set to each of the places that need it.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
```diff
@@ -184,6 +184,7 @@ private boolean isPartOfClosedWay(final Edge object)
         {
             wayIds.add(nextEdge.getIdentifier());
             final List<Edge> nextEdgeList = Iterables.stream(nextEdge.outEdges())
+                    .filter(Edge::isMasterEdge)
```
Bentleysb, line 187: Why is this being added?

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -128,8 +128,8 @@ public BaseCheck(final Configuration configuration)
         }
         catch (final Exception oops)
         {
-            logger.error(String.format("%s failed on feature %s.", this.getCheckName(),
-                    object.getIdentifier()), oops);
+            logger.error(String.format("%s failed on feature %s (%s).", this.getCheckName(),
+                    object.getIdentifier(), object.getOsmIdentifier()), oops);
```
Bentleysb, line 132: Is there a reason for this change? It appears to be redundant as the atlas identifier already contains the OSM id.

vladlemberg: yes. for debugging purpose its very convenient to copy original OSM id instead of playing with sectioned id in cases where original ID ends with 0 or small.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
vladlemberg: its a bug. i can submit it separately if needed. Iterables.stream(nextEdge.outEdges() returns both "master" and "reverse" edges -> traverse via "reverse". -> than returns to original edge -> and flagged it as closed. which is incorrect.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: Thanks for suggestion!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: Thanks for suggestion!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: compilation time. tested against China data:
1 - use WayWalker in each method - 7,48 mi (current)
2 - call WayWalker only ones and pass Set<Edge> edges to every method - 9,64 min
3 - make Set<Edge> static - 8,50 minutes.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: In this case every way will be closedWay. Instead of "contain" we need:
return Edges.get(0).start().getLocation().equals(Edges.get(Edges.size()-1).end().getLocation());

it slower than Edge Walker.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
Bentleysb: For tracking purposes, I think this would be better in a separate PR.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
Bentleysb: What about using `edge.inEdges().stream().filter(e -> e.getOsmIdentifier() == edge.getOsmIdentifier()).count() == 1`?

vladlemberg: slick suggestion! need to add our favorite .filter(Edge::isMasterEdge) to make it complete. thanks a lot!!

src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
sayas01: @Bentleysb this change was integrated in https://github.com/osmlab/atlas-checks/pull/305/files

docs/available_checks.md
```diff
@@ -28,6 +28,7 @@ This document is a list of tables with a description and link to documentation f
 | [~~RoundaboutClosedLoopCheck~~ (Deprecated)](checks/roundaboutClosedLoopCheck.md) | The purpose of this check is to identify Roundabout Edges that are bi-directional or have an end Node with less than 2 connections.  **This check has been deprecated and is no longer active.** |
 | [RoundaboutConnectorCheck](checks/roundaboutConnectorCheck.md) | The purpose of this check is to identify roads that connect to a roundabout at too sharp an angle |
 | [RoundaboutValenceCheck](checks/roundaboutValenceCheck.md) | The purpose of this check is to identify OpenStreetMap (OSM) tagged roundabouts that have an unusual number of edges connected to them. |
+| [RoundaboutMissingTag](checks/roundaboutMissingTagCheck.md) | The purpose of this check is to identify Roundabouts with missing junction=roundabout tag. Candidate must be navigable, closed and round shape OSM Way that intersects with at least two navigable roads. |
```
sayas01, line 31: 👍

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
sayas01: @Bentleysb This one has also been added. I think the branch is out of date with the dev branch.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.JunctionTag;
+import org.openstreetmap.atlas.tags.ServiceTag;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Angle;
+import org.openstreetmap.atlas.utilities.tuples.Tuple;
+
+/**
+ * Flags edges that are closed shaped and have minimum of two intersections with navigable roads.
```
sayas01, line 29: Could you update this description to include "missing junction=roundabout" ?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
sayas01: @Bentleysb This has also been added in a prev PR. I think this branch is out of date with the dev branch.

docs/available_checks.md
sayas01, line 31: 👍

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+public class RoundaboutMissingTagCheck extends BaseCheck<Long>
+{
+    // Instructions
+    public static final String MISSING_JUNCTION_TAG_INSTRUCTION = "This edge might be a roundabout";
+    // Minimum intersection with Navigable Roads
+    private static final int MINIMUM_INTERSECTION = 2;
+    private static final int MODULUS = 10;
+    private static final int FIRST_EDGE_SECTION = 1;
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(MISSING_JUNCTION_TAG_INSTRUCTION);
```
sayas01, line 46: nit: asList could be replaced with Collections.singletonList as it has just one item in it.

config/configuration.json
```diff
@@ -711,6 +711,17 @@
       "difficulty": "NORMAL"
     }
   },
+  "RoundaboutMissingTagCheck": {
+    "enabled": true,
```
sayas01, line 715: Looks like the config is missing the "threshold.degrees" keys.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+     * @param configuration
+     *            the JSON configuration for this check
+     */
+    public RoundaboutMissingTagCheck(final Configuration configuration)
+    {
+        super(configuration);
+        this.maxAngleThreshold = configurationValue(configuration, "threshold.degrees",
+                MAX_THRESHOLD_DEGREES_DEFAULT, Angle::degrees);
+        this.minAngleThreshold = configurationValue(configuration, "threshold.degrees",
+                MIN_THRESHOLD_DEGREES_DEFAULT, Angle::degrees);
```
sayas01, line 64: Shouldn't there be two separate config keys, "threshold.degrees.maximum" and "threshold.degrees.minimum" ?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+     */
+    private PolyLine buildOriginalOsmWayGeometry(final Edge edge)
+    {
+        // Identify all sections of original OSM way
+        final Set<Edge> edgesFormingOSMWay = new OsmWayWalker(edge).collectEdges().stream()
+                .filter(Edge::isMasterEdge).collect(Collectors.toSet());
+
+        // Sort sectioned Edges by Ids
+        final List<Edge> sortedEdges = new ArrayList<>(edgesFormingOSMWay);
+        sortedEdges.sort(Comparator.comparingLong(AtlasObject::getIdentifier));
```
sayas01, line 131: nit: these could be combined to one statement  `new OsmWayWalker(edge).collectEdges().stream()
                .filter(Edge::isMasterEdge).distinct()
                .sorted(Comparator.comparingLong(AtlasObject::getIdentifier))
                .collect(Collectors.toList());`

docs/checks/roundaboutMissingTagCheck.md
```diff
@@ -0,0 +1,39 @@
...
+
+Our first goal is to validate the incoming Atlas Object.
+* Must be a first section of valid master Edge
+* Must have not already been flagged
+* Must be car navigable
+* Must have not `junction` and `area` tag
+* Must be a closed way [OSM-wiki:Closed Way](https://wiki.openstreetmap.org/wiki/Item:Q4669)
+* Must intersect at minimum with two navigable ways 
+
+After the preliminary filtering of features we need to rebuild original OSM way geometry.
```
seancoulter, line 29: If we're making sure a valid edge is the first section of a way to ensure that the way has more than 2 edges, we could note that here

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+
+        // Build original OSM polyline
+        PolyLine geometry = new PolyLine(sortedEdges.get(0).getRawGeometry());
+        for (int i = 1; i < sortedEdges.size(); i++)
```
sayas01, line 135: Please avoid single letter variable names.

seancoulter, line 135: Nice, would be cool to have this pulled out into the Edge class or something sometime down the line

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+        final List<Tuple<Angle, Location>> maxOffendingAngles = originalGeom
+                .anglesGreaterThanOrEqualTo(this.maxAngleThreshold);
+        // check minimum angle
+        final List<Tuple<Angle, Location>> minOffendingAngles = originalGeom
+                .anglesLessThanOrEqualTo(this.minAngleThreshold);
+
+        if (maxOffendingAngles.isEmpty() && minOffendingAngles.isEmpty())
+        {
+            this.markAsFlagged(object.getOsmIdentifier());
+            return Optional.of(createFlag(new OsmWayWalker(edge).collectEdges(),
```
seancoulter, line 103: We could cache the waywalker call at the beginning of this method

seancoulter: Or potentially in the validCheckForObject method, as I see now the usage is out of this scope

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+    private boolean intersectingWithMoreThan(final Edge edge)
+    {
+        // Identify all sections of original OSM way
+        final Set<Edge> edgesFormingOSMWay = new OsmWayWalker(edge).collectEdges().stream()
+                .filter(Edge::isMasterEdge).collect(Collectors.toSet());
+        final Set<Long> connectedEdges = new HashSet<>();
+        edgesFormingOSMWay.forEach(obj -> obj.connectedEdges().stream().filter(Edge::isMasterEdge)
+                // intersection with navigable roads
+                .filter(HighwayTag::isCarNavigableHighway)
+                .filter(e -> e.getTag(ServiceTag.KEY).isEmpty())
```
sayas01, line 159: Same as my prev comment. `.filter(edge -> edge.getTag(ServiceTag.KEY).isEmpty())`

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param edge
+     *            entity to check
+     * @return original Way geometry polyline
+     */
+    private PolyLine buildOriginalOsmWayGeometry(final Edge edge)
+    {
+        // Identify all sections of original OSM way
+        final Set<Edge> edgesFormingOSMWay = new OsmWayWalker(edge).collectEdges().stream()
+                .filter(Edge::isMasterEdge).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
vladlemberg: Thanks!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: Thanks!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: thanks! added.

config/configuration.json
vladlemberg: thanks! added.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: added

docs/checks/roundaboutMissingTagCheck.md
vladlemberg: its noted: 
* Must be a first section of valid master Edge
* Must intersect at minimum with two navigable ways

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
vladlemberg: yes. tested it, however it significantly slowing down the performance. 
compilation time. tested against China data on local:
1 - use WayWalker in each method - 7,48 min (current)
2 - cache WayWalker and pass Set Edges to every method - 9,64 min

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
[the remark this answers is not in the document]
vladlemberg: thanks!

src/test/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheckTestRule.java
```diff
@@ -0,0 +1,267 @@
...
+    private static final String TEST_NODE_SHP_10 = "37.3293534,-121.9155234";
+    private static final String TEST_NODE_SHP_11 = "37.3293683,-121.9155658";
+    private static final String TEST_NODE_SHP_12 = "37.3293686,-121.9156274";
+    // connected way nodes
+    private static final String TEST_NODE_5 = "37.3303971,-121.9139893";
+    private static final String TEST_NODE_6 = "37.3281316,-121.9170493";
+
+    @TestAtlas(
+            // nodes
+            nodes = { @TestAtlas.Node(coordinates = @TestAtlas.Loc(value = TEST_NODE_VERTEX_1)),
```
sayas01, line 38: Just one more request to clean up this class a bit. You can import @TestAtlas.Node, @TestAtlas.Node etc. here.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/RoundaboutMissingTagCheck.java
```diff
@@ -0,0 +1,169 @@
...
+     * avoiding duplicate flags)
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return object instanceof Edge && !this.isFlagged(object.getOsmIdentifier())
+                && object.getIdentifier() % MODULUS == FIRST_EDGE_SECTION
+                && ((Edge) object).isMasterEdge() && HighwayTag.isCarNavigableHighway(object)
+                && object.getTag(JunctionTag.KEY).isEmpty() && object.getTag(AreaTag.KEY).isEmpty()
+                && isPartOfClosedWay((Edge) object) && intersectingWithMoreThan((Edge) object);
```
sayas01, line 76: nit: **this**.isPartOfClosedWay((Edge) object) && **this**.intersectingWithMoreThan((Edge) object)

docs/checks/roundaboutMissingTagCheck.md
seancoulter: Right -- i was more referring to the "why" we need to ensure a valid edge is the first section of the edge, as opposed to some other part of the edge. I might be alone in thinking that level of documentation is necessary so up to you!

PR #321, opened 2020-06-25
Disable RoundaboutMissingTagCheck

sayas01:
### Description:

This PR disables the newly added RoundaboutMissingTagCheck by default. The analysis of this check is still in progress and should be disabled until it is complete.

### Potential Impact:

Check will be disabled by default and will not be run until it is enabled.

### Unit Test Approach:

NA

### Test Results:

NA

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -723,6 +723,7 @@
     }
   },
   "RoundaboutMissingTagCheck": {
+    "enabled": false,
     "angle.threshold": {
       "maximum_degree": 40.0,
       "minimum_degree": 10.0


config/configuration.json
```diff
@@ -723,6 +723,7 @@
     }
   },
   "RoundaboutMissingTagCheck": {
+    "enabled": false,
```
danielduhh, line 726: Unless there's a glaring issue with performance, I'm okay with leaving all checks enabled in this repo

sayas01: The check has high FP based on recent feedback from analysis team so was considering disabling it here. There isn't any known issue with performance.

sayas01:
Closing as there isn't known issue with performance.

PR #322, opened 2020-06-29
[MapRouletteUploadCommand] Overwrite checkInCommentPrefix with additional switch

danielduhh:
### Description:

Adds a new `checkinComment` switch to the MapRouletteUploadCommand. This allows us to either:
1. Set the `checkinComment` value
2. Set the prefix for the [checkinComment](url) value
The `checkinComment value overwrites the `checkInCommentPrefix` value.

Why do we need both?

This switch is available for users that do not want the check name and iso country code in their check-in comment.

### Potential Impact:

More options for MapRouletteUploadCommand

### Unit Test Approach:

Uploaded some tests to the public staging environment

### Test Results:

Challenge values look good 👍

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.StringUtils;
 import org.openstreetmap.atlas.checks.maproulette.data.Challenge;
 import org.openstreetmap.atlas.checks.maproulette.data.ChallengeStatus;
 import org.openstreetmap.atlas.checks.maproulette.data.Task;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteUploadCommand.java
```diff
@@ -61,8 +62,11 @@
             string -> Arrays.asList(string.split(",")), Optionality.OPTIONAL);
     private static final Switch<String> CHECKIN_COMMENT_PREFIX = new Switch<>(
             "checkinCommentPrefix",
-            "MapRoulette checkinComment prefix. This will be prepended to the check name",
+            "MapRoulette checkinComment prefix. This will be prepended to the check name (e.g. #prefix: [ISO - CheckName] ).",
             String::toString, Optionality.OPTIONAL, Challenge.DEFAULT_CHECKIN_COMMENT);
```
seancoulter, line 66: Nice that we have a default to fall back on

dcastrowa:
Nice solution

PR #326, opened 2020-07-06
LineCrossingWaterbodyCheck performance improvements

seancoulter:
### Description:

Recent LCWB runs have shown that it suffers from slow runtimes. This is reflected in the sharded and nonsharded atlas checks frameworks. This PR aims to tune up the performance of this check in a few different ways. 

One change is related to redundant geometry operations. LCWB had a few geometry-related function calls that were called 2 to 3 times on a given object. These are now cached and passed around. Minor improvements include checking object tagging before more contextual geometry checks. I've also created an overloaded `IntersectionUtilities. haveExplicitLocationsForIntersections` to handle the case where certain parameters are given and do not need to be calculated.

The other change, which will likely result in TP flag subtractions, is limiting the size of the multipolygon waterbodies processed by the check. This is done by summing the number of points in each outer polygon of a given multipolygon relation. If this exceeds some threshold, the multipolygon is skipped and not flagged. This is to limit the expensive geometry operations to small/medium sized waterbodies. There is a risk of data loss from doing this. However the runtime speedup is profound and is considered the lesser of both evils

The final change is a minor bug fix. Previously we had the potential to pick up land features that intersected the Polygon representation of LineItem members of multipolygon relations. Such Polygon representations are not true to the original LineItem geometry in that they have an extra synthetic segment connecting the start and end nodes of the line item. Thus we were at risk of picking up FP land features that intersected that segment. The fix was to check for intersection against PolyLines, not Polygons.

### Potential Impact:

LCWB should run more efficiently

### Unit Test Approach:

Same unit tests as before.

Did some flag analysis in RUS and runtime analysis for full country runs. Configurables used: `shapepoints.min`: `(default)` and `shapepoints.max`: `5000`

### Test Results:

Unit tests pass

RUS metrics:

Before:
iso	shard		name							duration (ms)	minutes
RUS	7-75-36		LineCrossingWaterBodyCheck		20671628	344.527133333333
RUS	7-100-40	LineCrossingWaterBodyCheck		6246947		104.115783333333
RUS	7-113-42		LineCrossingWaterBodyCheck		4001990		66.6998333333333
RUS	8-186-81	LineCrossingWaterBodyCheck		3621906		60.3651
RUS	8-225-86	LineCrossingWaterBodyCheck		2927794		48.7965666666667
RUS	8-152-68	LineCrossingWaterBodyCheck		2154004		35.9000666666667
RUS	9-334-155	LineCrossingWaterBodyCheck		2118323		35.3053833333333
RUS	8-166-80	LineCrossingWaterBodyCheck		1937125		32.2854166666667
RUS	9-325-163	LineCrossingWaterBodyCheck		1928588		32.1431333333333
RUS	8-166-79	LineCrossingWaterBodyCheck		1853373		30.88955
RUS	8-165-79	LineCrossingWaterBodyCheck		1847557		30.7926166666667
RUS	8-152-69	LineCrossingWaterBodyCheck		1813876		30.2312666666667
RUS	8-165-80	LineCrossingWaterBodyCheck		1749364		29.1560666666667
RUS	7-104-42	LineCrossingWaterBodyCheck		1698242		28.3040333333333
RUS	7-75-31		LineCrossingWaterBodyCheck		1676469		27.94115
RUS	8-224-87	LineCrossingWaterBodyCheck		1650159		27.50265
RUS	8-167-80	LineCrossingWaterBodyCheck		1612443		26.87405
RUS	8-153-69	LineCrossingWaterBodyCheck		1597237		26.6206166666667
RUS	7-113-43		LineCrossingWaterBodyCheck		1557904		25.9650666666667
RUS	8-225-87	LineCrossingWaterBodyCheck		1545796		25.7632666666667
RUS	8-155-76	LineCrossingWaterBodyCheck		1515924		25.2654

The above table shows a collection of shards by descending runtime in the pre-enhancement baseline.

After:
image
This table shows shards by descending runtime, post-enhancement. The longest-taking shard took just under 12 minutes. This is contrasted by pre-enhancement run times, where the longest shard runtime was over 300 minutes and it was not uncommon to see 20+ minute runtimes for other shards. The average post-enhancement runtime across all shards is 35019ms or 35 seconds. 

It took approximately 3.5 hours to run the full sharded checks suite on over 200 countries in the post-enhancement, while it took 6.5+ hours for the pre-enhancement suite.

Note that the shape points limiter, which is a big contributor to final LCWB runtimes, can be tweaked via the configuration. The tradeoff is faster runtimes for dropped TPs vs. slower runtimes for more comprehensive flagging for larger waterbodies


danielduhh:
Hey @seancoulter! Do you have any before/after metrics for some of the problematic shards?

Issue #331, opened 2020-07-23
Waterway checks (does not end in sink, crosses without intersection, closed waterway)


taylorsmock:
**Description**
This is largely a migration of Osmose 1070/1220 (waterway overlaps, waterway intersecting waterway without junction, unconnected waterway, and closed waterway).

**Requirements**
* Is an `Edge` or a `Line` (logic should use `LineItem` where ever possible), since waterways may become an `Edge` in the future (specifically navigable waterways, so rivers, not small streams), and has valid `waterway` water values.

**Use Cases**
This section summarizes the various use cases for this check and provides potential enhancement suggestions if needed. The uses cases should act as a test in a way, such that when the analysis/enhancement is completed, these types of use cases should be found or eliminated.

For each use case, please add a screenshot of the feature & openstreetmap.org link (or other sources). If necessary, edit the screenshot to better illustrate the use case.

Example:

*Case 1: Sinkhole area ([Colorado, US](https://www.openstreetmap.org/way/538940188))*
 - This should not be flagged, since sinkholes act as an exit for waterways.
 - https://www.openstreetmap.org/way/538940188
![image](https://user-images.githubusercontent.com/45215054/88223411-ac4f7980-cc24-11ea-851d-b34b70f7fbe2.png)

*Case 2: Sinkhole point ([Utah, US](https://www.openstreetmap.org/node/7704038746))*
 - This should not be flagged, since sinkholes act as an exit for waterways. Note that this is apparently not a sinkhole (see the `description` tag, but that is a separate issue).
 - https://www.openstreetmap.org/node/7704038746
![image](https://user-images.githubusercontent.com/45215054/88223459-c426fd80-cc24-11ea-8ec2-9eba46ffc96b.png)

*Case 3: Closed waterway ([Colorado, US](https://www.openstreetmap.org/way/46115766))*
- This should be flagged. It looks like it should be `natural=water` + `water=canal` instead of `waterway=ditch` (it looks more like a canal than a ditch).
- https://www.openstreetmap.org/way/46115766
![image](https://user-images.githubusercontent.com/45215054/88223519-d7d26400-cc24-11ea-8b27-6254960b7881.png)

*Case 4: Waterway ending in an ocean ([Florida, US](https://www.openstreetmap.org/node/1846978542))*
- This should not be flagged, since water ending inside an ocean has a sink
- https://www.openstreetmap.org/way/556331627 (Coastline)
[...]

PR #332, opened 2020-07-23
Basic waterway checks

taylorsmock:
### Description:

This does basic waterway checks (closed waterway, waterway does not end in sink, crossing waterway).
See #331.

There are also two new utility classes, `ElevationUtilities` and `CompressionUtilities`. The `ElevationUtilities` reads HGT files (for example, those provided by the SRTM NASA mission), and when given a point, returns the elevation for the appropriate location (only as accurate as the source data -- it _should_ be able to take arbitrarily accurate HGT files). `ElevationUtilities` reads from `elevation` directory in the directory that the check run is started from. This can be customized. `CompressionUtilities` is used to decompress file streams, so that if a file is in a `tar.gz`, `zip`, `gz`, or isn't compressed at all, the user can be relatively certain that the stream is now usable if the file stream is passed into the utility method. This also helps reduce the size on disk for elevation files (i.e., if there is a high resolution HGT file that takes up 1 GiB uncompressed).

To support usage of SRTM elevation files, there is a script in `scripts/elevationData` called `srtm-hgt.py`. This file fetches the SRTM NASA files at a slow rate, so that the serving USGS servers are not significantly burdened by the download. The script tries to ensure that if it is killed, the appropriate actions are taken upon the next start (skip/restart/continue). These files have (at best) ~90m resolution, which means that they are not useful for short lengths, or any method which requires more precise locations. For these instances, more precise data may be provided manually.

### Potential Impact:

Better waterway navigation will be possible. Closed waterways can be very confusing to data consumers (how does it work?), waterways that don't end in a sink (so an ocean, drain, or sinkhole) may or may not be navigable, but it cannot be determined easily. Crossing waterways may allow navigation between them, but only if they are connected (what happens if one is in a tunnel?).

### Unit Test Approach:

The unit tests primarily test for potential false positives, along with some potential false negatives.

### Test Results:

| AIA | 0 | 0 | n/a | n/a | n/a |n/a |
|  BLZ   |    311 |    35      |    11% |   35 | 0 | 0% |
| DMA | 30 | 30 | 100% | 30 | 0 | 0% |
| CYM | 7 | 7 | 100% | 7 | 0 | 0% [1] |
| [NZL](https://maproulette.org/challenge/13975)[2] | 12182 | 40 | 0.3% | 38 | 2 | 5% |
| [NZL](https://maproulette.org/challenge/13975) (second take, same challenge rebuilt) | 12191 | 156 | 1.2% | 132 | 9 | 5.7% |
| SMR | 10 | 10 | 100% | 5 | 5 | 50%[3] |

[1] 3 of the TP detections were difficult to determine (e.g., waterway in wetland -- spec doesn't allow for them to end, but it makes sense, waterway may go multiple directions, etc.). However, the wiki specification does not allow for them. No actual waterway visible (still an issue).
[2] NZL had an import of waterways (and other land cover information). Most of the issues found are waterways without sinks, and much of the time, there are multiple related issues (so a series of waterways may have the wrong direction). In the sample layer, 63 waterways were modified, so ~55 issues were actually fixed (some were crossing waterways, but very few were). If this were to be released publicly, it would be a good idea (tm) to update the MapRoulette task weekly (if people actually used it). A false positive is from a waterway ending on a lake where I was unable to find an outlet for the lake (there is probably one somewhere). Another false positive is from a waterway continuing through a coast and into the ocean a fair ways. I've linked to the MapRoulette challenge I used for the spot checking -- please don't use it for editing at this time.
[3] The false positives can be categorized into two separate categories:
1. Boundary issues (the atlas way appears cut at an administrative boundary in Atlas, 3/5 fp)
2. Layer issues (layer=-1 -> no layer, 2/5 fp). Both issues could also be due to boundaries, since a specific test did not fail with layer issues. Having opened the source atlas file, both are valid detections (the connecting waterways do not appear in the file).

### False Negatives:
[Way 350806025](https://www.openstreetmap.org/way/350806025): This way is not closed, but when taken with other ways, it forms a loop. [Osmose](http://osmose.openstreetmap.fr/en/map/#zoom=17&lat=-43.842208&lon=170.434695&item=xxxx&level=1%2C2%2C3&tags=&fixable=) does not find it either. This may be partially ameliorated by the addition of height data (e.g., from the SRTM Nasa mission).

### Notes
There is an ongoing discussion on the mailing lists about abruptly ending waterways. This test will probably need to be updated once a consensus is reached (i.e., tagging for when the waterway spreads out and is soaked up by the earth).

|branch|time|notes|
|-------|----|------|
|dev| 2h 28m 26s | `./gradlew clean build run` with `AIA,CYM,DMA,NZL,SMR` |
|feature/waterway-check| 2h 34m 2s | `./gradlew clean build run` with `AIA,CYM,DMA,NZL,SMR` |

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -1091,12 +1091,18 @@
     }
   },
   "WaterWayCheck": {
-    "ocean.boundary": "natural->coastline",
-    "ocean.valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock",
-    "waterway.elevation.distance.min.start.end": 450.0,
-    "waterway.elevation.resolution.min.uphill": 1.0,
-    "waterway.sink.tags.filters": "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain",
-    "waterway.tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised",
+    "ocean": {
+      "boundary": "natural->coastline",
+      "valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock"
+    },
+    "waterway": {
+      "elevation": {
+        "distance.min.start.end": 450.0,
+        "resolution.min.uphill": 1.0
+      },
+      "sink.tags.filters": "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain",
+      "tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised"
+    },
     "challenge": {
       "description": "Waterways that have circular flows, are crossing, do not have a sink, or are going uphill (if elevation data was provided)",
       "blurb": "Fix waterways such that they make physical sense",
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -1096,6 +1096,13 @@
     "waterway.elevation.distance.min.start.end": 450.0,
     "waterway.elevation.resolution.min.uphill": 1.0,
     "waterway.sink.tags.filters": "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain",
-    "waterway.tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised"
+    "waterway.tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised",
+    "challenge": {
+      "description": "Waterways that have circular flows, are crossing, do not have a sink, or are going uphill (if elevation data was provided)",
+      "blurb": "Fix waterways such that they make physical sense",
+      "instruction": "Open your favorite editor and edit the waterways",
+      "difficulty": "NORMAL",
+      "defaultPriority": "LOW"
+    }
   }
 }
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -1089,5 +1089,13 @@
       "instruction": "Open your favorite editor and edit the features overlapping the ocean so they either validly overlap or do not overlap.",
       "difficulty": "MEDIUM"
     }
+  },
+  "WaterWayCheck": {
+    "ocean.boundary": "natural->coastline",
+    "ocean.valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock",
+    "waterway.elevation.distance.min.start.end": 450.0,
+    "waterway.elevation.resolution.min.uphill": 1.0,
+    "waterway.sink.tags.filters": "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain",
+    "waterway.tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised"
   }
 }
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
@@ -43,6 +43,8 @@ public final class ElevationUtilities implements Serializable
      */
     public static final short NO_ELEVATION = Short.MIN_VALUE;
 
+    private static final short[][] EMPTY_MAP = new short[][] {};
+
     /** Just an int for converting a decimal to a percentage */
     private static final int DECIMAL_TO_PERCENTAGE = 100;
 
@@ -102,7 +104,7 @@ public final class ElevationUtilities implements Serializable
     public short getElevation(final Location location)
     {
         final short[][] map = getMap(location);
-        if (map == null)
+        if (Arrays.equals(EMPTY_MAP, map))
         {
             return NO_ELEVATION;
         }
@@ -138,11 +140,15 @@ public final class ElevationUtilities implements Serializable
      *
      * @param location
      *            The location to get the resolution of
-     * @return The resolution of the data
+     * @return The resolution of the data, or {@link Distance#MAXIMUM} if there is no data.
      */
     public Distance getResolution(final Location location)
     {
         final short[][] map = getMap(location);
+        if (Arrays.equals(EMPTY_MAP, map))
+        {
+            return Distance.MAXIMUM;
+        }
         final float difference = ((float) this.srtmExtent) / map.length;
         final Location temp = new Location(location.getLatitude(),
                 Longitude.degrees(location.getLongitude().asDegrees() + difference));
@@ -162,7 +168,7 @@ public final class ElevationUtilities implements Serializable
     {
         final short[][] mapOne = getMap(one);
         final short[][] mapTwo = getMap(two);
-        if (Arrays.equals(mapOne, mapTwo))
+        if (Arrays.equals(mapOne, mapTwo) && !Arrays.equals(EMPTY_MAP, mapOne))
         {
             final int[] indexOne = getIndex(one, mapOne.length);
             final int[] indexTwo = getIndex(two, mapTwo.length);
@@ -268,7 +274,7 @@ public final class ElevationUtilities implements Serializable
         }
         if (!path.toFile().isFile())
         {
-            return null;
+            return EMPTY_MAP;
         }
         try (InputStream is = CompressionUtilities
                 .getUncompressedInputStream(Files.newInputStream(path)))
@@ -277,7 +283,7 @@ public final class ElevationUtilities implements Serializable
         }
         catch (final IOException e)
         {
-            return null;
+            return EMPTY_MAP;
         }
     }
 
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
@@ -66,7 +66,7 @@ public final class ElevationUtilities implements Serializable
      */
     private static String configurationKey(final String key)
     {
-        return formatKey("elevationutilities", key);
+        return formatKey(ElevationUtilities.class.getSimpleName(), key);
     }
 
     private static <U, V> V configurationValue(final Configuration configuration, final String key,
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
@@ -36,7 +36,7 @@ public final class ElevationUtilities implements Serializable
     /**
      * The assumed extent of a HGT SRTM file (lat/lon) in degrees
      */
-    private static final int SRTM_EXTENT = 1;
+    private static final double SRTM_EXTENT = 1;
     /**
      * A non-number when there is no elevation data available. This is currently returns
      * {@link Short#MIN_VALUE}.
@@ -51,7 +51,7 @@ public final class ElevationUtilities implements Serializable
     /** A map of {lat, lon} pairs with a loaded srtm in a byte array */
     private final Map<Pair<Integer, Integer>, short[][]> loadedSrtm = new HashMap<>();
 
-    private final int srtmExtent;
+    private final double srtmExtent;
 
     private final String srtmExtension;
 
@@ -88,11 +88,28 @@ public final class ElevationUtilities implements Serializable
      */
     public ElevationUtilities(final Configuration configuration)
     {
-        this.srtmExtent = configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT,
-                i -> i);
-        this.srtmExtension = configurationValue(configuration, "elevation.srtm_ext", SRTM_EXTENSION,
-                i -> i);
-        this.srtmPath = configurationValue(configuration, "elevation.path", "elevation", i -> i);
+        this(configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT, i -> i),
+                configurationValue(configuration, "elevation.srtm_ext", SRTM_EXTENSION, i -> i),
+                configurationValue(configuration, "elevation.path", "elevation", i -> i));
+    }
+
+    /**
+     * Create a configured ElevationUtilities
+     *
+     * @param srtmExtent
+     *            The extent of the files. The units are in degrees, and the default is 1 degree.
+     * @param srtmExtension
+     *            The extension of the files. Archive/compression file endings are not required
+     *            (they will be automatically detected).
+     * @param srtmPath
+     *            The path for the files.
+     */
+    public ElevationUtilities(final double srtmExtent, final String srtmExtension,
+            final String srtmPath)
+    {
+        this.srtmExtension = srtmExtension;
+        this.srtmExtent = srtmExtent;
+        this.srtmPath = srtmPath;
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
@@ -88,9 +88,13 @@ public final class ElevationUtilities implements Serializable
      */
     public ElevationUtilities(final Configuration configuration)
     {
-        this(configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT, i -> i),
-                configurationValue(configuration, "elevation.srtm_ext", SRTM_EXTENSION, i -> i),
-                configurationValue(configuration, "elevation.path", "elevation", i -> i));
+        this(configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT,
+                Function.identity()),
+                configurationValue(configuration, "elevation.srtm_ext", SRTM_EXTENSION,
+                        Function.identity()),
+                configurationValue(configuration, "elevation.path", "elevation",
+                        Function.identity()));
+
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
@@ -1,0 +1,346 @@
+package org.openstreetmap.atlas.checks.utility;
+
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Serializable;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.openstreetmap.atlas.geography.Location;
+import org.openstreetmap.atlas.geography.Longitude;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * Used to determine assumed direction for some features (e.g. waterways). Assumptions are that any
+ * SRTM file follows the specification here:
+ * https://dds.cr.usgs.gov/srtm/version2_1/Documentation/Quickstart.pdf
+ *
+ * @author Taylor Smock
+ */
+public final class ElevationUtilities implements Serializable
+{
+    private static final long serialVersionUID = -5929570973909280629L;
+    /**
+     * The assumed file extension
+     */
+    private static String SRTM_EXT = "hgt";
+    /**
+     * The assumed extent of a HGT SRTM file (lat/lon) in degrees
+     */
+    private static int SRTM_EXTENT = 1;
+    /**
+     * A non-number when there is no elevation data available. This is currently returns
+     * {@link Short#MIN_VALUE}.
+     */
+    public static final short NO_ELEVATION = Short.MIN_VALUE;
+
+    /** Just an int for converting a decimal to a percentage */
+    private static final int DECIMAL_TO_PERCENTAGE = 100;
+
+    /** A map of {lat, lon} pairs with a loaded srtm in a byte array */
+    private final Map<Pair<Integer, Integer>, short[][]> loadedSrtm = new HashMap<>();
+
+    private final int srtmExtent;
+
+    private final String srtmExt;
+
+    private final String srtmPath;
+
+    /**
+     * Configuration Keys in the Integrity Framework are based on the check simple classname.
+     *
+     * @param key
+     *            key part for a specific configuration item defined for this class
+     * @return complete key for lookup
+     */
+    private static String configurationKey(final String key)
+    {
+        return formatKey("elevationutilities", key);
+    }
+
+    private static <U, V> V configurationValue(final Configuration configuration, final String key,
+            final U defaultValue, final Function<U, V> transform)
+    {
+        return configuration.get(configurationKey(key), defaultValue, transform).value();
+    }
+
+    private static String formatKey(final String name, final String key)
+    {
+        return String.format("%s.%s", name, key);
+    }
+
+    /**
+     * Create a configured ElevationUtilities
+     *
+     * @param configuration
+     *            A configuration which should (at a minimum) have a file path for elevation files.
+     */
+    public ElevationUtilities(final Configuration configuration)
+    {
+        this.srtmExtent = configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT,
+                i -> i);
+        this.srtmExt = configurationValue(configuration, "elevation.srtm_ext", SRTM_EXT, i -> i);
+        this.srtmPath = configurationValue(configuration, "elevation.path", "elevation", i -> i);
+    }
+
+    /**
+     * Get the elevation of a location
+     *
+     * @param location
+     *            The location to get the elevation of
+     * @return The meters of the elevation (does not show fractions of meters)
+     */
+    public short getElevation(final Location location)
+    {
+        final short[][] map = getMap(location);
+        if (map == null)
+        {
+            return NO_ELEVATION;
+        }
+        final int[] index = getIndex(location, map.length);
+        return map[index[0]][index[1]];
+    }
+
+    /**
+     * Get the incline between two points
+     *
+     * @param start
+     *            The start of the slope
+     * @param end
+     *            The end of the slope
+     * @return The incline from the start to the end (may be negative). May also return
+     *         {@link Double#NaN} if one of the elevations cannot be obtained.
+     * @see <a href="https://wiki.openstreetmap.org/wiki/Key:incline">OSM Wiki Incline</a>
+     */
+    public double getIncline(final Location start, final Location end)
+    {
+        final short startElevation = getElevation(start);
+        final short endElevation = getElevation(end);
+        if (startElevation == NO_ELEVATION || endElevation == NO_ELEVATION)
+        {
+            return Double.NaN;
+        }
+        final double distance = end.distanceTo(start).asMeters();
+        return ((endElevation - startElevation) / distance) * DECIMAL_TO_PERCENTAGE;
+    }
+
+    /**
+     * Get the resolution of the data at the location.
+     *
+     * @param location
+     *            The location to get the resolution of
+     * @return The resolution of the data
+     */
+    public Distance getResolution(final Location location)
+    {
+        final short[][] map = getMap(location);
+        final float difference = ((float) this.srtmExtent) / map.length;
+        final Location temp = new Location(location.getLatitude(),
+                Longitude.degrees(location.getLongitude().asDegrees() + difference));
+        return temp.distanceTo(location);
+    }
+
+    /**
+     * Check if two locations are in the same data point (i.e., same pixel in a HGT)
+     *
+     * @param one
+     *            A location to check
+     * @param two
+     *            Another location to check
+     * @return {@code true} if they are in the same grid and location
+     */
+    public boolean inSameDataPoint(final Location one, final Location two)
+    {
+        final short[][] mapOne = getMap(one);
+        final short[][] mapTwo = getMap(two);
+        if (Arrays.equals(mapOne, mapTwo))
+        {
+            final int[] indexOne = getIndex(one, mapOne.length);
+            final int[] indexTwo = getIndex(two, mapTwo.length);
+            if (Arrays.equals(indexOne, indexTwo))
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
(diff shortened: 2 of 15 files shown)


danielduhh:
Awesome, thanks @taylorsmock!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,453 @@
...
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final LineItem line = (LineItem) object;
+        final Location last = line.asPolyLine().last();
+        final Location first = line.asPolyLine().first();
+        CheckFlag flag = null;
+        final double incline = this.elevationUtils.getIncline(first, last);
+        final boolean uphill = !Double.isNaN(incline) && incline > 0
+                && last.distanceTo(first).isGreaterThan(this.minDistanceStartEndElevationUphill);
+        if (line.isClosed())
```
seancoulter, line 353: nit: looks like we could move this check up a few lines before the elevation checks

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,453 @@
...
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return !isFlagged(object.getOsmIdentifier()) && object instanceof LineItem
+                && this.waterwayTagFilter.test(object);
+    }
+
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
```
seancoulter, line 344: General question: is it possible to get a waterway that is an offender in multiple ways to this Check? E.g. a waterway that ends in a sink and crosses itself. If so we could append instructions together in one flag instead of returning only one instruction set per flag.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,453 @@
...
+
+    private boolean doesWaterwayEndInSink(final LineItem line)
+    {
+        return doesLineEndOnWaterway(line) || doesLineEndInSink(line) || doesLineEndInOcean(line);
+    }
+
+    private LineItem intersectsAnotherWaterWay(final LineItem line)
+    {
+        final Atlas atlas = line.getAtlas();
+        final Iterable<LineItem> intersectinWaterways = atlas.lineItemsIntersecting(line.bounds(),
```
seancoulter, line 424: nit: intersectingWaterways

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,453 @@
...
+        {
+            flag = createFlag(object,
+                    this.getLocalizedInstruction(FALLBACK_INSTRUCTIONS.indexOf(CIRCULAR_WATERWAY),
+                            object.getOsmIdentifier()),
+                    Collections.singletonList(line.asPolyLine().first()));
+        }
+        else if (uphill && this.minResolutionDistance
+                .isGreaterThanOrEqualTo(this.elevationUtils.getResolution(first)))
+        {
+            flag = createUphillFlag(object, first);
```
seancoulter, line 363: We probably want to add that this check flags uphill waterways as well (docs/javadoc)
I noticed it in a couple places, but wasn't sure if it was consistently mentioned

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: It is probably technically possible (so crossing waterways, waterway not properly connected to another waterway, and waterway going in the wrong direction).

I'll look into appending instructions tomorrow.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: pick: nit (I'll fix it, it would bug me as well if I didn't...)

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: I'll do that (I wrote the docs, then wrote the test, then checked NZL, where there are _many_ waterways that are going the wrong direction, and _then_ wrote the elevation check code).

taylorsmock:
> A couple thoughts so far. Thanks for making those utility classes!
Well, I kind of needed them. There are going to be some other use cases for them, eventually. (For example, depending upon the resolution of the elevation data, it could be possible to determine if a building could be 200+ stories tall).

As an FYI, I just rebased onto `dev` (again). I also ran the check against New Zealand with appended instructions, and there were some waterways that had multiple instructions, so that has been pushed as well.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
```diff
@@ -0,0 +1,361 @@
...
+        return String.format("%s.%s", name, key);
+    }
+
+    /**
+     * Create a configured ElevationUtilities
+     *
+     * @param configuration
+     *            A configuration which should (at a minimum) have a file path for elevation files.
+     */
+    public ElevationUtilities(final Configuration configuration)
```
Bentleysb, line 89: I think this would be more reusable if the 3 srtm values were given as the constructor parameters, instead of needing to be in a configuration.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * @author Taylor Smock
+ */
+public final class ElevationUtilities implements Serializable
+{
+    private static final long serialVersionUID = -5929570973909280629L;
+    /**
+     * The assumed file extension
+     */
+    private static final String SRTM_EXT = "hgt";
```
Bentleysb, line 35: Nit:
```suggestion
    private static final String SRTM_EXTENTION = "hgt";
```

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
```diff
@@ -0,0 +1,361 @@
...
+
+    /** Just an int for converting a decimal to a percentage */
+    private static final int DECIMAL_TO_PERCENTAGE = 100;
+
+    /** A map of {lat, lon} pairs with a loaded srtm in a byte array */
+    private final Map<Pair<Integer, Integer>, short[][]> loadedSrtm = new HashMap<>();
+
+    private final int srtmExtent;
+
+    private final String srtmExt;
```
Bentleysb, line 56: Nit
```suggestion
    private final String srtmExtension;
```

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,467 @@
...
+        {
+            final Collection<Pair<Segment, Segment>> segs = getIntersectingSegments(linePolyline,
+                    lineItem.asPolyLine());
+            final Segment max = segs.stream().map(Pair::getKey).distinct().max(segmentComparator)
+                    .orElse(null);
+            final Collection<Segment> crosses = segs.stream().filter(p -> p.getLeft().equals(max))
+                    .map(Pair::getValue).collect(Collectors.toSet());
+            final PolyLine coast = new PolyLine(crosses.stream()
+                    .flatMap(c -> Stream.of(c.first(), c.last())).toArray(Location[]::new));
+            if (isRightOf(coast, linePolyline.last())
```
Bentleysb, line 226: I am a bit confused what is happening at this point (some comments throughout this method would help). Is this considering a polyline to end in the ocean if it is to the east of a coast?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,467 @@
...
+     *
+     * @param line
+     *            The line to check
+     * @return {@code true} if the line ends on a waterway (the last node is part of a waterway)
+     */
+    public boolean doesLineEndOnWaterway(final LineItem line)
+    {
+        final List<LineItem> waterways = new ArrayList<>();
+        line.getAtlas()
+                .lineItemsIntersecting(line.asPolyLine().last().boxAround(Distance.ONE_METER),
```
Bentleysb, line 285: I think you can simplify this method by using `Atlas.lineItemsContaining(Location)`, and filtering from there.

Bentleysb (commented):
Nice work dealing with some pretty complex stuff in this check. I just had a few questions and comments.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
taylorsmock: I've expanded EXT/Ext to the appropriately cased extension.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
taylorsmock: I'll go ahead and make another constructor (I'd like to leave the configuration option in).

config/configuration.json
```diff
@@ -1068,5 +1068,13 @@
       "difficulty": "EASY",
       "defaultPriority": "LOW"
     }
+  },
+  "WaterWayCheck": {
```
Bentleysb, line 1072: This needs a challenge object with the MapRoulette information.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
Bentleysb: That sounds good. 
For the configuration constructor, where in the configuration are you expecting the values to be? Are they part of a check object, their own object, or top level entities? It would also be good to have a unit test to make sure custom values are read in properly.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: There are two assumptions at play here:
1) `natural=coastline` follows the specification, and the ocean is to the right of the coastline way.
2) The geometry isn't so complicated that the waterway loops around the coastline (I can make a contrived example, but I haven't seen any real-world examples that couldn't be more accurately modeled).

I've gone ahead and added some documentation.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: I didn't see that when I was coding. I probably similar code in the other water check PR I made.

Thanks. :)

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
taylorsmock: I'm expecting them to be part of their own configuration section, e.g.
```json
{
    "elevationutilities": {
        "elevation.srtm_extent": "1" (degree),
        "elevation.srtm_ext": "hgt" (file extension),
        "elevation.path": "elevation"
    }
}
```
(I copied/pasted from `docs/utilities/elevationUtilities`).
I'll see if I can make a specific test for it tomorrow (I thought I had one for that somewhere, but it doesn't look like it).

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
Bentleysb: I see, thanks for the clarification and the docs. It might be worth it to add a note about the ocean specification assumption in either the flag instructions or the MapRoulette challenge instructions in the configuration. That way editors may be able to identify incorrectly modeled coastlines that are getting caught by this check.

config/configuration.json
```diff
@@ -1068,5 +1068,20 @@
...
+    "ocean.boundary": "natural->coastline",
+    "ocean.valid": "natural->strait,channel,fjord,sound,bay|harbour->*&harbour->!no|estuary->*&estuary->!no|bay->*&bay->!no|place->sea|seamark:type->harbour,harbour_basin,sea_area|water->bay,cove,harbour|waterway->artificial,dock",
+    "waterway.elevation.distance.min.start.end": 450.0,
+    "waterway.elevation.resolution.min.uphill": 1.0,
+    "waterway.sink.tags.filters": "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain",
+    "waterway.tags.filters": "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised",
```
danielduhh, line 1078: You can simplify these configs with:
```
ocean: {
  boundary: "",
  valid: ""
},
waterway: {
 elevation: {
  distance.min.start.end: "",
  resolution.min.uphill: ""
 },
 sink.tags.filters: "",
 tags.filters: ""

}
```

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+            final int segment2index = this.lineSegments.indexOf(segment2real);
+            return segment1index - segment2index;
+        }
+    }
+
+    private static final long serialVersionUID = 2877101774578564205L;
+    private static final String WATERWAY_SINK_TAG_FILTER_DEFAULT = "natural->sinkhole|waterway->tidal_channel,drain|manhole->drain";
+    private static final String WATERWAY_TAG_FILTER_DEFAULT = "waterway->river,stream,tidal_channel,canal,drain,ditch,pressurised";
+
+    private static final String DEFAULT_VALID_OCEAN_TAGS = "natural->strait,channel,fjord,sound,bay|"
```
danielduhh, line 88: Is this copied from OceanBleedingCheck? https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/OceanBleedingCheck.java#L40

If so, lets pull them both out into a tag util that can be shared

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+    private static final String CROSSES_WATERWAY = "The waterway {0} crosses the waterway {1}.";
+    private static final String GOES_UPHILL = "The waterway {0} probably does not go up hill.\nPlease check (source elevation data resolution was about {1} meters).";
+    private static final Distance MIN_RESOLUTION_DISTANCE = Distance.ONE_METER;
+
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(DOES_NOT_END_IN_SINK,
+            CIRCULAR_WATERWAY, CROSSES_WATERWAY, GOES_UPHILL);
+
+    private final TaggableFilter waterwaySinkTagFilter;
+    private final TaggableFilter waterwayTagFilter;
+
```
danielduhh, line 106: nit: extra space

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+    private final TaggableFilter validOceanTags;
+
```
danielduhh, line 108: nit: extra space

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+        return Optional.of(flag);
+    }
+
+    @Override
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    private CheckFlag createUphillFlag(final CheckFlag flag, final AtlasObject object,
```
danielduhh, line 442: Do you mind adding java docs here?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+                .collectToSet();
+        sameLayerWays.removeIf(line::equals);
+        if (sameLayerWays.isEmpty())
+        {
+            return null;
+        }
+        return sameLayerWays.iterator().next();
+    }
+
+    private boolean waterwayConnects(final LineItem line, final LineItem potential)
```
danielduhh, line 478: Same comment here on documentation

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,472 @@
...
+        flag.addInstruction(instruction);
+        return flag;
+    }
+
+    private boolean doesWaterwayEndInSink(final LineItem line)
+    {
+        return doesLineEndOnWaterway(line) || doesLineEndInSink(line) || doesLineEndInOcean(line);
+    }
+
+    private LineItem intersectsAnotherWaterWay(final LineItem line)
```
danielduhh, line 461: Java docs here as well 😄

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,493 @@
...
+     * @return {@code true} if the object ends on a boundary
+     */
+    private static boolean endsWithBoundaryNode(final AtlasObject object)
+    {
+        if (!(object instanceof LineItem))
+        {
+            return false;
+        }
+        final LineItem lineItem = (LineItem) object;
+        final Atlas atlas = object.getAtlas();
```
danielduhh, line 155: `getAtlas()` can be an expensive operation. You should be able to store the atlas in the flag function in pass it to your helpers that need access to the entire atlas.

src/main/java/org/openstreetmap/atlas/checks/utility/ElevationUtilities.java
```diff
@@ -0,0 +1,379 @@
...
+
+    /**
+     * Create a configured ElevationUtilities
+     *
+     * @param configuration
+     *            A configuration which should (at a minimum) have a file path for elevation files.
+     */
+    public ElevationUtilities(final Configuration configuration)
+    {
+        this(configurationValue(configuration, "elevation.srtm_extent", SRTM_EXTENT, i -> i),
```
danielduhh, line 91: Please avoid single letter variables

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,493 @@
...
+    public boolean doesLineEndOnWaterway(final LineItem line)
+    {
+        final List<LineItem> waterways = new ArrayList<>();
+        line.getAtlas().lineItemsContaining(line.asPolyLine().last(), this.waterwayTagFilter::test)
+                .forEach(waterways::add);
+
+        waterways.removeIf(line::equals);
+        final Location last = line.asPolyLine().last();
+        return waterways.stream().anyMatch(
+                l -> l.asPolyLine().contains(last) && !last.equals(l.asPolyLine().last()));
```
danielduhh, line 316: nit: please avoid single letter variables

danielduhh (commented):
Hey @taylorsmock, what's the workflow for generating accurate data in this check? Does it require running the python script before hand? If so, how do you suggest we do this on a cluster?

taylorsmock:
> Hey @taylorsmock, what's the workflow for generating accurate data in this check? Does it require running the python script before hand? If so, how do you suggest we do this on a cluster?

This check does _not_ require the SRTM (.hgt) files. It does, however, improve directions and find additional cases.

I would _not_ recommend running the script per run, although I did take some pains to ensure that multiple runs wouldn't place undo burden on the public servers. I would run it once and then store the data somewhere (probably in the same location as the atlas files).

Initial workflow
* Run script to get SRTM files

Ongoing workflow
* Run your standard script (change configuration to point to the appropriate location to get the SRTM files)

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: IIRC, it is. I think I copied it, with the intention of seeing if I would have to change it. I never did.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,529 @@
...
+    {
+        final PolyLine tLine = new PolyLine(location);
+        final Segment closest = line.segments().stream()
+                .min(Comparator.comparingDouble(s -> s.shortestDistanceTo(tLine).asMeters()))
+                .orElse(null);
+        if (closest != null)
+        {
+            final PolyLine testLine = new PolyLine(closest.first(), closest.last(), location);
+            final Angle difference = testLine.headingDifference().orElse(null);
+            if (difference != null && difference.asDegrees() > 0)
```
sayas01, line 336: Could be simplified as :  `return difference != null && difference.asDegrees() > 0;`

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,529 @@
...
+     * Check if a line crosses an coastline line
+     *
+     * @param line
+     *            The line to check
+     * @return {@code true} if the line crosses a coastline
+     */
+    private boolean doesLineCrossCoast(final LineItem line)
+    {
+        final List<LineItem> lines = new ArrayList<>();
+        line.getAtlas()
```
sayas01, line 201: Same comment as https://github.com/osmlab/atlas-checks/pull/332/files#r485768561 on storing the atlas and passing it to helpers.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,529 @@
...
+     * Check if the waterway ends in a sink (i.e., a location that can reasonably expected to have
+     * no outflow).
+     *
+     * @param line
+     *            The waterway item to check
+     * @return {@code true} if the waterway ends in a sink.
+     */
+    private boolean doesWaterwayEndInSink(final LineItem line)
+    {
+        return doesLineEndOnWaterway(line) || doesLineEndInSink(line) || doesLineEndInOcean(line);
```
sayas01, line 476: Missing `this` keyword.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,526 @@
...
+     * @return {@code true} if the waterway can reasonably be expected to end
+     */
+    public boolean doesLineEndInSink(final LineItem lineItem)
+    {
+        // If the way is a sink, it obviously ends in a sink...
+        if (this.waterwaySinkTagFilter.test(lineItem))
+        {
+            return true;
+        }
+        if (lineItem instanceof Line)
```
seancoulter, line 256: I think we could have the same treatment of Lines and Edges here. If either feature has a node or point at its end that's a sink, return true

seancoulter (commented):
Looking good, just some minor suggestions

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: That wasn't part of the style guide. (The guide being `./gradlew spotlessCheck`). Should look into updating that to force methods to have `this` as a qualifier?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,526 @@
...
+        final Atlas atlas = line.getAtlas();
+        CheckFlag flag = null;
+        if (line.isClosed())
+        {
+            flag = createFlag(object,
+                    this.getLocalizedInstruction(FALLBACK_INSTRUCTIONS.indexOf(CIRCULAR_WATERWAY),
+                            object.getOsmIdentifier()),
+                    Collections.singletonList(line.asPolyLine().first()));
+        }
+        final double incline = this.elevationUtils.getIncline(first, last);
```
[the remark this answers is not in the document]
taylorsmock: I elected to not do this for two reasons:
1) The elevation data may not be "fine" enough to determine elevation between two nearby points (and won't be, in most cases).
2) There may be a ditch that is lower than the surrounding terrain that is not accounted for in the elevation data (highly likely with SRTM data).

Both (1) and (2) would be ameliorated if the elevation data is sufficiently fine. However, public domain data from the SRTM NASA mission have ~30m resolution, at best. Other data sources may be in the public domain and may have sufficient resolution for this, but I was more focused on general incline for the line instead of specific incline between points. I suppose we could check that the incline between each point in the line is downwards, and then have a specific instruction set for that.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,526 @@
...
+        final Set<LineItem> sameLayerWays = Iterables.stream(intersectingWaterways)
+                .filter(potential -> LayerTag.areOnSameLayer(line, potential)
+                        && !waterwayConnects(line, potential))
+                .collectToSet();
+        sameLayerWays.removeIf(line::equals);
+        if (sameLayerWays.isEmpty())
+        {
+            return null;
+        }
+        return sameLayerWays.iterator().next();
```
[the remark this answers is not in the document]
taylorsmock: I was trying to "flag" each object as quickly as possible, to avoid significantly increasing processing times. Since then, I started working on a generic check for invalid tags (using TagInfo data and OSM Wiki Data items), and I learned a whole new meaning for "long processing time" (for reference, my 5 country test runs went from ~2 hours to 14+ hours, and in one case, almost 23 hours). I've gotten that back down to ~3 hours, and I'm looking for more optimizations for that, but this isn't in the same category at all.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
```diff
@@ -0,0 +1,526 @@
...
+        if (!(object instanceof LineItem))
+        {
+            return false;
+        }
+        final LineItem lineItem = (LineItem) object;
+        final Atlas atlas = object.getAtlas();
+        final Location last = lineItem.asPolyLine().last();
+        final Stream<Point> points = Iterables.asList(atlas.pointsAt(last)).stream();
+        final Stream<Node> nodes = Iterables.asList(atlas.nodesAt(last)).stream();
+        return Stream.concat(points, nodes).anyMatch(SyntheticBoundaryNodeTag::isBoundaryNode);
```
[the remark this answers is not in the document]
taylorsmock: I'm not seeing this in sonar. Am I looking in the wrong place?

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas-checks&pullRequest=332&resolved=false

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: This goes back to trying to optimize to avoid significant increases in processing time. Again, compared to my latest in-progress check, this is peanuts. :)

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: I was kind of surprised by this. I expected that call to be pretty cheap. :(

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
taylorsmock: I created another Utility class specifically for common filters.

dcastrowa:
@taylorsmock is this ready for another round of reviews?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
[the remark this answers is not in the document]
seancoulter: Ah I misread -- your code is fine, Sonar would only complain if an operation was applied to the stream that wasn't a terminal operation. You've got the right place, and you should be able to integrate Sonar into your IDE as well depending on your IDE. check out this for IntelliJ if you haven't already - https://www.sonarlint.org/intellij/index.html

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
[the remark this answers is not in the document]
seancoulter: That's fair! To avoid over complicating the elevation part of the check I think it's fine as is. If we ever come across a dataset with finer resolution we could pull this out into a new check to look at the problem more closely

seancoulter (dismissed):
LGTM, on the condition the other convos are resolved. Thanks for this exciting contribution!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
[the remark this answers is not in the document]
taylorsmock: I use Eclipse, and I do have it integrated.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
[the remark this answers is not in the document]
taylorsmock: I've added a comment to indicate that it can be expanded in the future, if it makes sense with available data sets.

taylorsmock:
> @taylorsmock is this ready for another round of reviews?

I think so. The only outstanding conversation is with @sayas01, with regard to using `this` for method calls. If `this` is required for method calls, I'm going to have to make some modifications so that `spotless` fixes/checks for that.

Bentleysb (approved):
LGTM, pending the 'this' keyword discussion.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/lines/WaterWayCheck.java
sayas01: I think it would be good to add it to spotlessCheck. @taylorsmock

sayas01 (approved):
LGTM! 👍 We can include the `this `qualifier check to spotless in future PR.

PR #346, opened 2020-08-20
Water Area checks

### LINKED ISSUE #345: Water Area checks (waterway area overlaps, riverbank without river)
**Description**
 This is largely a migration of Osmose 1070/1220 (waterway area overlaps, riverbank without river). This is a followup of #331 . This is in a separate pull request due to having slightly different requirements, and likely needing some work on the `intersects` logic. The waterway overlaps actually falls under Osmose 1150.

**Requirements**

* Is an Edge or a Line (logic should use LineItem where ever possible), in case there is an non-closed waterway area
* Is an Area or Relation (relations should be checked just in case, but it _should_ be an area before it is passed to the check).
and
* Has a waterway area tag (mapcss fragment)
```
area[waterway=riverbank],
area[natural=water],
area[natural=coastline]
```

**Use Cases**
This section summarizes the various use cases for this check and provides potential enhancement suggestions if needed. The uses cases should act as a test in a way, such that when the analysis/enhancement is completed, these types of use cases should be found or eliminated.

For each use case, please add a screenshot of the feature & openstreetmap.org link (or other sources). If necessary, edit the screenshot to better illustrate the use case.

Example:

* Case 1: Riverbank without waterway ([Belize](https://www.openstreetmap.org/way/661564606))
Riverbank without river
[Osmose issue](http://osmose.openstreetmap.fr/en/error/acd78df8-c883-08c1-9603-cab85341eb77)

* Case 2: Overlapping water area ([Belize](https://www.openstreetmap.org/way/448755487))
Overlapping pond
[Osmose issue](http://osmose.openstreetmap.fr/en/error/487bd6ba-5907-6194-3796-402391459946)

**Further investigation**
* Osmose seems to ignore `natural=water` + `water=river` for riverbank area missing river way.

**Supported regions**
All countries

#331 has the rest of the relevant checks, and therefore a pull request for this issue will not duplicate those checks, and hence does not support them.

**References**
* OSM Wiki Waterways: https://wiki.openstreetmap.org/wiki/Waterways
* OSM Wiki Riverbank: https://wiki.openstreetmap.org/wiki/Tag:waterway%3Driverbank
* Waterway way checks: https://github.com/taylorsmock/atlas-checks/tree/feature/waterway-check

taylorsmock:
### Description:

This should fix #345.
This pull request should depend upon #332 .

### Potential Impact:

There should be no down-stream impact at this time. This may change (intersection logic).

### Unit Test Approach:
Unit tests are largely for false negatives, but there are some false positive checks as well.

### Test Results:

| AIA | 0 | 0 | N/A | 0 | 0 | N/A |
| CYM | 0 | 0 | N/A | 0 | 0 | N/A |
| DMA | 0 | 0 | N/A| 0 | 0 | N/A |
| NZL[1] | 594 | 61 | 10.2% | 60 | 1 | 1.6% |
| SMR[2] | 2 | 2 | 100% | 0 | 2 | 100% |

[1] As with #332, most of the issues are from a poorly done import. The single FP from the sample was from a `natural=water + water=stream_pool`. This is a true false positive, and could be ameliorated by changing the `WATER_FILTERS` static variable. (something like `natural->water&water->*&water->!stream_pool|waterway->riverbank`).
[2] Like the previous water area check (#332 ), this appears to be due to clipping at the border, and not due to logic issues.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
@@ -78,7 +78,6 @@ public class WaterAreaCheckTest
         this.verifier.actual(this.setup.getOverlappingPonds(),
                 new WaterAreaCheck(ConfigurationResolver.emptyConfiguration()));
         this.verifier.verifyExpectedSize(1);
-        // TODO verify actual check
     }
 
     @Test
--- a/src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
@@ -1,0 +1,91 @@
+package org.openstreetmap.atlas.checks.validation.areas;
+
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.openstreetmap.atlas.checks.configuration.ConfigurationResolver;
+import org.openstreetmap.atlas.checks.validation.verifier.ConsumerBasedExpectedCheckVerifier;
+
+/**
+ * Tests for WaterAreaCheck
+ *
+ * @author Taylor Smock
+ */
+public class WaterAreaCheckTest
+{
+    @Rule
+    public WaterAreaCheckTestRule setup = new WaterAreaCheckTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    /**
+     * This was a test for a false-false positive. The false positive occurred due to the age of the
+     * Belize pbf, where a waterway was drawn in April 2019 was not present.
+     */
+    @Test
+    public void testBrazilRiverFalsePositive()
+    {
+        this.verifier.actual(this.setup.getBrazilRiverFalsePositive(),
+                new WaterAreaCheck(ConfigurationResolver.emptyConfiguration()));
+        this.verifier.verifyEmpty();
+    }
+
+    @Test
+    public void testMangoCreekAndUnnamedWaterwayBad()
+    {
+        this.verifier.actual(this.setup.getMangoCreekAtlasBad(),
+                new WaterAreaCheck(ConfigurationResolver.emptyConfiguration()));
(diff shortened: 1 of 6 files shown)


Bentleysb (commented):
This is looking really good. Just a couple of things. One, we generally try to avoid acronyms and 1 letter variable names in this repo, even in lambda functions. Two, you have most of the logic for all three use cases in the flag method. It would make things a lot cleaner and more readable if they were each separated into their own methods.

Bentleysb (commented):
This looks good. If you can just add an MD doc for this and add it to https://github.com/osmlab/atlas-checks/blob/dev/docs/available_checks.md, then I think it will be good to go.

src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
```diff
@@ -0,0 +1,91 @@
...
+        this.verifier.verifyEmpty();
+    }
+
+    @Test
+    public void testOverlappingPonds()
+    {
+        this.verifier.actual(this.setup.getOverlappingPonds(),
+                new WaterAreaCheck(ConfigurationResolver.emptyConfiguration()));
+        this.verifier.verifyExpectedSize(1);
+        // TODO verify actual check
```
sayas01, line 81: Is this going to be addressed in this PR?

sayas01 (commented):
Nice work @taylorsmock ! Thanks for adding the detailed unit tests. LGTM once @Bentleysb's comment is resolved.

taylorsmock:
> This looks good. If you can just add an MD doc for this and add it to https://github.com/osmlab/atlas-checks/blob/dev/docs/available_checks.md, then I think it will be good to go.

I've added the docs.

Bentleysb (approved):

src/test/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheckTest.java
taylorsmock: I was intending to go and verify that I got one of the expected objects, but the flagged objects don't (currently) allow me to get the original AtlasObject back out. I could do some reflection to get it working, but I'd prefer to avoid that when I'm not unit testing that specific class.

sayas01 (approved):
LGTM! 👍

Bentleysb (approved):

PR #347, opened 2020-08-20
Code smells (sonar/NPE)

taylorsmock:
### Description:

This updates some code to use new (Java 8+) features, and fixes an NPE that was noticed.

### Potential Impact:

This also has some modernization efforts as well (so generics are used in some locations). This may require some downstream projects to `atlas-checks` to recompile, but should not require any significant source code changes.

### Unit Test Approach:

N/A -- this is largely cosmetic changes (but ones that will be useful).


Bentleysb (approved):
This all looks good. Thanks for doing this!

danielduhh (approved):

PR #354, opened 2020-09-04
Replace some non-inclusive language

jwpgage:
### Description:

This update replaces some non-inclusive language in the codebase. The master branch is replaced by the main branch in travis scripts.

### Potential Impact:

One parameter has been renamed from 'master' to 'cluster', and accordingly any use of that parameter should be updated to reflect that.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
@@ -45,17 +45,17 @@ import com.google.gson.GsonBuilder;
 public abstract class BaseCheck<T> implements Check, Serializable
 {
     public static final String PARAMETER_ACCEPT_PIERS = "accept.piers";
-    public static final String PARAMETER_BLACKLIST_COUNTRIES = "countries.blacklist";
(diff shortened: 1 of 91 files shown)


src/main/java/org/openstreetmap/atlas/checks/base/BaseCheck.java
```diff
@@ -261,17 +261,18 @@ public void logStatus()
     @Override
     public boolean validCheckForCountry(final String country)
     {
-        // whitelist is valid if country list is empty or country within the country list. It always
-        // takes precedence over blacklist and blacklist is essentially ignored if whitelist is non
+        // permitlist is valid if country list is empty or country within the country list. It
+        // always
```
matthieun, line 265: Maybe re-arrange the comments there in the right line?

matthieun (approved):
Just one small comment, otherwise this looks great @jwpgage! Thanks for the update!

matthieun (changes requested):
Oh I just realized you also need to update the `.travis.yml` file too, otherwise the build fails:
https://github.com/osmlab/atlas/pull/674/files#diff-354f30a63fb0907d4ad57269548329e3R22-R44

matthieun (approved):

Issue #334, opened 2020-07-27
InvalidTagsRegexCheck - bad source


mm-ciub:
**Description**
The InvalidTagsRegexCheck could have a similar approach as the InvalidTagsCheck but would use regex to verify certain tag values. 
To begin with it should verify if the source tag describes a legal source of data imported in OSM.
Examples of invalid sources: Google, Here. 

**Requirements**
- Any OSM primitive that has a **source** tag should be feasible for this check. 

**Use Cases**
There are multiple examples of cases where certain information in OSM comes from illegal sources like:
- Google
- Here
- vworld
[OSM blacklist](https://www.openstreetmap.org/api/capabilities)

Example:

*Case 1: [Cycleway originating from Google.] (Romania)*
 - The way has a google maps link as a source
 - (https://www.openstreetmap.org/way/560117102)

*Case 2: [Park originating from Google.] (Italy)*
 - The park has google as source
 - (https://www.openstreetmap.org/way/276514644)

*Case 3: [Building originating from VWorld.] (South Korea)*
 - The building has VWorld as source
 - (https://www.openstreetmap.org/way/368940416)

*Case 4: [Building originating from Here.] (Germany)*
 - The building has Here as source
 - (https://www.openstreetmap.org/node/4062814842)

**Further investigation**

**Supported regions**
All regions should be supported.

**What's not supported (optional):**

**References**
https://wiki.openstreetmap.org/wiki/Key:source
https://wiki.openstreetmap.org/wiki/Tag:boundary=administrative

PR #356, opened 2020-09-07
BadValueTagCheck

mm-ciub:
### Description:

The main focus for this check is to verify the source tag for illegal sources like "Google" or "Here".
The check can be extended with other filters containing different tag names and regex patterns to be checked against.
https://github.com/osmlab/atlas-checks/issues/334

### Potential Impact:

No impacts.

### Unit Test Approach:

Added unit tests that check the sources for nodes and edges with illegal tags such as maps link or Here maps.

### Test Results:

The check was run on shards from AIA, AUS, CYM, DMA, NZL, TUR, with results only for AUS, TUR and NZL.

|  AUS,TUR,NZL   |    47         |    38      |     80%       |   35 |   3 | 8%   |

There are still some open cases if certain values should be allowed or not, for example google drive links

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -427,10 +427,7 @@
     }
   },
   "InvalidTagsCheck": {
-    "filters.resource": {
-      "override": true,
-      "append": false
-    },
+    "filters.resource.override": true,
     "filters.classes.tags": [
       ["area","boundary->protected_area&protect_class->!"],
       ["relation","boundary->protected_area&protect_class->!"]
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -443,6 +443,15 @@
       "tags":"tags"
     }
   },
+  "BadValueTagCheck": {
+    "challenge": {
+      "description": "Tasks containing features with tags containing unacceptable values",
+      "blurb": "Features with unacceptable tag values",
+      "instruction": "Open your favorite editor and check that the listed tag values are accepted by OSM.",
+      "difficulty": "Medium",
+      "tags":"tags"
+    }
+  },
   "InvalidTurnRestrictionCheck": {
     "challenge": {
       "description": "Tasks containing invalid turn restrictions",
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/tag/BadValueTagCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/tag/BadValueTagCheck.java
@@ -35,7 +35,10 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonParser;
 
 /**
- * TODO
+ * This check allows verifying certain tag values using regex patterns based on configurable filters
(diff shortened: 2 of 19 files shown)


mm-ciub:
[Bad Source check validation.pdf](https://github.com/osmlab/atlas-checks/files/5187805/Bad.Source.check.validation.pdf)

config/configuration.json
```diff
@@ -440,6 +440,15 @@
       "tags":"tags"
     }
   },
+  "BadValueTagCheck": {
```
sayas01, line 443: Could you also add this to the https://github.com/osmlab/atlas-checks/blob/dev/docs/available_checks.md doc?

config/configuration.json
```diff
@@ -440,6 +440,15 @@
...
+    "challenge": {
```
sayas01, line 444: A nice to have would be a config to pass invalid regex pattern similar to InvalidTagsCheck.

sayas01 (commented):
Overall this looks good to me. I feel like the RegexTaggableFilter class could be pulled out and added to the Atlas repo as I can see it being useful in other checks.

danielduhh (commented):
Hey @mm-ciub, thanks for contributing!
Outside of the use of regex, are there any additional differences between this and InvalidTagsCheck? Do you think it makes sense to have a separate check for this use case?
Just wondering if we can simplify this by consolidating with the existing general tags check. Let me know what you think!

mm-ciub:
> Hey @mm-ciub, thanks for contributing!
> Outside of the use of regex, are there any additional differences between this and InvalidTagsCheck? Do you think it makes sense to have a separate check for this use case?
> Just wondering if we can simplify this by consolidating with the existing general tags check. Let me know what you think!

Hey! I initially wanted to extend the InvalidTagsCheck but upon consideration and discussions it was decided that this should be a new check were all regex related validation will be added. It can still be moved back if that no longer seems to be the case.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/BadValueTagCheck.java
```diff
@@ -0,0 +1,219 @@
...
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    /**
+     * The {@code RegexTaggableFilter} predicate is used to test an Atlas Object tag with multiple
+     * regex patterns.
+     */
+    private static class RegexTaggableFilter implements Predicate<AtlasObject>, Serializable
```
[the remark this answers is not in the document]
mm-ciub: I completely agree, it would make sense to exist in the same place as TaggableFilter. I only added it here to speed up the process.

mm-ciub:
Logic contained in https://github.com/osmlab/atlas-checks/pull/368 per comment suggestions.

PR #357, opened 2020-09-08
Generic Tag Checking (Osmose 3050, 3150, and 4090)

### LINKED ISSUE #352: Bad tags (osmose 3050, 3150, and 4090)
**Description**
Check for bad tags, tags that have a typo, and unwanted tags.

For unwanted tags, the tag should exist in [OSM Wiki Data](https://wiki.openstreetmap.org/wiki/Data_items) (e.g., https://wiki.openstreetmap.org/w/api.php?action=wbgetentities&sites=wiki&titles=Key:lanes&languages=en&format=json -- human readable [here](https://wiki.openstreetmap.org/wiki/Item:Q372)). This also affects tags that have a typo (if the key indicates that it has well-known values (example: [Key:bridge:movable](https://wiki.openstreetmap.org/wiki/Item:Q104)), if there is not a wiki item for that value, then more work must be done). This requirement is due to the [Meta](https://wiki.openstreetmap.org/wiki/Data_items#Meta_item) information that each key can have (`de facto`, `in use`, `approved`, `rejected`, `voting`, `draft`, `abandoned`, `proposed`, `obsolete`, `deprecated`, `discardable`, and `imported`). `discardable` and `rejected` are definitely unwanted, while `obsolete` and `deprecated` should be used to indicate that there is a newer tagging method to the user.

**Requirements**
1. Check Candidate
- Has tags
- Is not already flagged by an existing bad tag/key/value check (this may be difficult to do, see `ConditionalRestrictionCheck`, a utility class/parent class may be necessary to register keys that have _specific_ checks, but probably calling `validCheckForObject` will be good enough).

**Use Cases**
This section summarizes the various use cases for this check and provides potential enhancement suggestions if needed. The uses cases should act as a test in a way, such that when the analysis/enhancement is completed, these types of use cases should be found or eliminated.

For each use case, please add a screenshot of the feature & openstreetmap.org link (or other sources). If necessary, edit the screenshot to better illustrate the use case.

Example:

*Case 1: Invalid Highway Tag ([Togo](https://www.openstreetmap.org/node/5343281075))*
 - `highway=Station de carburant pour voiture` should be flagged (and should already be flagged). This check is primarily for instances where hardcoded values are not available (as it is with `highway`).
 - https://www.openstreetmap.org/node/5343281075
Station de carburant pour voiture

**Further investigation**
It is expected that there will be false positives where a proposal has recently been accepted, but a Wiki Data item has not been created. An additional script should be provided to parse the json for items that may need to be added to the wiki (automatically done for keys by [bot](https://wiki.openstreetmap.org/wiki/Data_items#Item_creation_process) when a tag matches a regex AND there are more than 10 instances).

**Supported regions**
All countries

Autocorrections -- this may be possible with the addition of Levenshtein distance calculations, when there is expected to be a well-known value (so names need not apply).

**Intended methodology:**
* `InvalidTagsCheck` will be modified to check the key and value of an object if it is not flagged already (the assumption being that the specific flags are more accurate than those on the wiki)
* If a local TagInfo database exists, query the database for the key and value incidence. Record this to a variable.
* For each key, get the appropriate wiki data item (preferably from a data dump, _not_ the live server, a utility script similar to [this](https://github.com/Sophox/sophox/blob/metabot/metabot/metabot/DataItems.py#L29) should be written).
  * If the key has the `well-known` values classification, check that the tag page exists. Else if there is a `regex`, check the value against the regex. Else continue on.
  * If the key has a `status` of `discardable`, `rejected`, `abandoned`, or `imported` (`imported` is a maybe), flag for removal.
  * If the key does not exist in wiki data, and the key is over a specific (configurable?) threshold, write instructions for a wiki editor
  * If the key does not exist in wiki data, and the key is under a specific (configurable?) threshold, write instructions for modification/removal
  * If the key is over a specific (configurable?) threshold, is _not_ `well-known`, flag for a wiki editor to add a regex expression checker for (so a `name` tag could be expected to match `.*`).
  * If the key is prohibited on the current element (node/way/area/relation), add instructions to that effect
* A script must be written to remove and consolidate instructions for wiki editors, to separate those from those that should be fixed.

**References**
* [OSM Data Items](https://wiki.openstreetmap.org/wiki/Data_items)
* [OSM Wiki Dump](https://wiki.openstreetmap.org/dump/) (either this, making lots of network calls)
* [Value Validation Regex](https://wiki.openstreetmap.org/wiki/Property:P13) ([example](https://wiki.openstreetmap.org/wiki/Item:Q574))
* [TagInfo Usage Database](https://taginfo.openstreetmap.org/download)
* [Deprecated Tags](https://wiki.openstreetmap.org/wiki/Deprecated_features)
* [Tagwatch Cleaner](https://wiki.openstreetmap.org/wiki/User:FrViPofm/TagwatchCleaner)

tsmock:
### Description:

This should fix #352 in a broad manner. This checks that the tag is documented. If not, and the tag is above a specific threshold, additional instructions are generated for wiki editors (a separate script still needs to be written to remove these from geojson files for Maproulette challenges -- the wiki data should be edited by people familiar with the wiki).

### Potential Impact:

No changes should be required by downstream libraries.

### Unit Test Approach:
* Unit tests check for both "incorrect" and "correct" results (mutation testing was used for this)

### Test Results:

|  SMR   |    80         |    76      |     95%       |   75 |   1 [1] | 27%   |
| AIA | 47 | 47 | 100% | 47 | 0[2] | 100% |
| BRA | 98487 | 516 | 0.52% || 2 | 0.4% |
| NZL | 30110 | 300 | 0.99% || 0 | 0% |
| PHL | 116277 | 300 | 0.25% || 3 | 1% |
| IND | 89515 | 500 | 0.55% || 1 | 0.2% |
| IDN | 42015 | 500 | 1.1% || 0 | 0% |

[1] In SMR, the only true FP was due to localization (`short_name:<LANG>` specifically). These should still be documented, but may not reasonably be documented due to the breadth of languages. There were many poorly documented tags (for WikiData). An example would be `public_transport:version=2`.
[2] All flags required either a tag change _or_ a wiki change/clarification.


tsmock:
I'm moving this from draft to ready for the following reasons:
1) It is getting large enough to be hard to review (4500 lines is probably a bit much)
2) Thus far, in testing, the only issues I've found were in WikiData. This means that the first few runs _must_ be completed by experienced wiki editors/users, and will require some coding to create new key/tag items. In order to help ensure that changed items can be checked, the get_wikidata.py script prints changed elements to a log file.

For example, there are some keys which are on boundary relations, but are prohibited on areas. Atlas considers a boundary to be an area (and so do I), and so the wiki data item should be clarified/fixed.

`description` was prohibited on relations. This has been fixed (the `description` key is from a pre-relation proposal).

Further notes:
1) Sometimes there are multiple issues with the same object, such that the autofix doesn't fix everything. In order to avoid having an "incomplete" fix, if the size of the instructions does not match the size of the fixes, the fixes are not appended. This should be changed in the event that MapRoulette understands "partial" fixes.
2) This is a generic tag check which looks for the following issues:
* Tag on object conflicts with other tag
* Key/Tag is undocumented or unpopular
* Deprecated/replaced/removable key/tags
* Checks the WikiData regex, if any, against the tag
* If the key values are well known, ensure that the current value is documented
* Check if the tag cannot be used on the element type
3) Future expansion (method stubs are present in current code)
* Relation and relation roles
* Check that a key/tag can be used or only can be used in a country
4) There is a new annotation so that the GenericTagCheck can skip tags that have explicit checks (e.g., `name`, `addr:street`, etc.)

sayas01:
Hi @tsmock, looks like there are 10+ code smells, mostly associated to the GenericTagCheck class. Please fix those. Also, it would be great if you could pull out some of the logic from the GenericTagCheck class to some utility classes to reduce the length of the class(1k+ lines) for readability purpose.

tsmock:
I'm moving this back to `draft` status due to discovering a bug with the external data change.

See:
https://github.com/osmlab/atlas/pull/727
https://github.com/osmlab/atlas-generator/pull/171 (depends upon Atlas #727)

atiannicelli:
@tsmock - first I want to say that this is a great check!!! and sorry just getting to review this! 
I have a couple of requests that I will put here so we can track them and their resolution better. 

1. The first issue I found was that when the feature being flagged is an edge then the proposed changes show nothing. After speaking ot you directly you found that the feature ID that was being fed to MR (and probably what is in the flag file) is incorrect. For example the OSM ID for this task (https://maproulette.org/challenge/16123/task/79523851) looks like this in the task: 
```
"id": "176504519000001176504519000002176504519000003176504519000004",
```
This may not be an issue with your check, but we need to find out the cause. If it is outside of your check then we can create a new issue and fix it in a new PR. If it is caused by this PR then we should fix it here.

2. The next issue was that when we find a deprecated tag we are proposing to remove it. See this task: https://maproulette.org/challenge/16123/task/79526432 where we suggest to remove the building=entrance. The [wiki](https://wiki.openstreetmap.org/wiki/Tag:building%3Dentrance) says that this can be converted to entrance=*. I think that if we want to keep the proposed change then we should suggest the most common change. (eg. "building=entrance" -> "entrance=yes")

3. I love that in the instructions you linked to the wiki page that describes the specific tag in some cases. I think that we should extend that to more of the different types of issues. For instance the example above with the deprecation would be great if it linked to the page that showed that it had been deprecated.

4. While talking directly you mentioned that the "created_by" tag key is "discardable" and probably isn't worth flagging. As it stands the configuration file lists discardable tags in its filter:
```
    "wikidata.tag_removal": ["abandoned", "deprecated", "discardable", "imported", "obsolete", "rejected"],
```
I think we should probably remove "discardable" from this filter in the config file. That will allow others to enable it if they really want those flagged, but if discardable tags are discarded by editors automatically then I agree with you that this should not be flagged by default. 

5. I'm also confused by the invalid relation role tasks I'm seeing. this task: 
https://maproulette.org/challenge/16123/task/79550601 says that "boundary=label is an unknown relation role for boundary", but if I read the wiki correctly it is valid to have a "label" role for boundaries. If this a bug? or am I reading this wrong?

tsmock:
1) I'll have to rerun with a recent atlas (you mentioned that it may have been fixed since I last ran the check).
2) I'll update the text such that there is a link to the data item with instructions for updating it, if there is a known replacement. I'll probably add a caveat that it will _not_ be immediately be reflected in the challenge.
3) Thanks. Since this depends upon wiki data to be useful, it was kind of necessary for me to do this. The only issue I've noticed is that MapRoulette doesn't like any of the links for data items (it stops at the `:` character for the link, IIRC). I should probably add links in all the instructions for the wiki data, at least.
4) Easy enough to change. And I've been meaning to do so. :)
5) Probably not a bug. It probably just wasn't in the wiki data (which is why I've tried to make it so that people can update/create wiki data with a JOSM plugin, and this particular challenge should be initially restricted to advanced users only).
In the sqlite db I used for this check, the following roles for boundary relations were available:

id | [P21](https://wiki.openstreetmap.org/wiki/Property:P21) | [P43](https://wiki.openstreetmap.org/wiki/Property:P43)
-- | -- | --
[Q16058](https://wiki.openstreetmap.org/wiki/Item:Q16058) | boundary=inner | [Q16019](https://wiki.openstreetmap.org/wiki/Item:Q16019)
[Q16059](https://wiki.openstreetmap.org/wiki/Item:Q16059) | boundary=outer | [Q16019](https://wiki.openstreetmap.org/wiki/Item:Q16019)
[Q16060](https://wiki.openstreetmap.org/wiki/Item:Q16060) | boundary=admin_centre | [Q16019](https://wiki.openstreetmap.org/wiki/Item:Q16019)
[Q16061](https://wiki.openstreetmap.org/wiki/Item:Q16061) | boundary=subarea | [Q16019](https://wiki.openstreetmap.org/wiki/Item:Q16019)
[Q16103](https://wiki.openstreetmap.org/wiki/Item:Q16103) | boundary= | [Q16019](https://wiki.openstreetmap.org/wiki/Item:Q16019)

Notably absent is `label`, so that would need to be added to the wiki data.

atiannicelli:
I am rerunning it now with the latest code merged in and will have output soon. So far all the "edge" feature issues don't have fix suggestions... looking for one that does. But here is an example of a flag file line that has edge data (no fix suggestion). The only weirdness I see is that the "id" in the "properties" section is a concat of all edges for this OSM way.
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [121.0551927, 14.2519563],
          [121.0551818, 14.2528253],
          [121.055155, 14.2529449],
          [121.0550853, 14.2529865],
          [121.0548814, 14.2530229]
        ]
      },
      "properties": {
        "last_edit_changeset": "64592616",
        "identifier": "90980214000001",
        "itemType": "Edge",
        "access": "permissive",
        "bicycle": "yes",
        "iso_country_code": "PHL",
        "osmIdentifier": "90980214",
        "source": "Google",
        "last_edit_user_name": "leven",
        "last_edit_time": "1542450771000",
        "last_edit_user_id": "1255860",
        "name": "Alsace St",
        "lanes": "2",
        "highway": "residential",
        "last_edit_version": "2",
        "foot": "yes"
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [121.0548814, 14.2530229],
          [121.0546078, 14.2530645]
        ]
      },
      "properties": {
        "last_edit_changeset": "64592616",
        "identifier": "90980214000002",
        "itemType": "Edge",
        "access": "permissive",
        "bicycle": "yes",
        "iso_country_code": "PHL",
        "osmIdentifier": "90980214",
        "source": "Google",
        "last_edit_user_name": "leven",
        "last_edit_time": "1542450771000",
        "last_edit_user_id": "1255860",
        "name": "Alsace St",
        "lanes": "2",
        "highway": "residential",
        "last_edit_version": "2",
        "foot": "yes"
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [121.0546078, 14.2530645],
          [121.0543932, 14.2531061],
          [121.0543181, 14.2530853]
        ]
      },
      "properties": {
        "last_edit_changeset": "64592616",
        "identifier": "90980214000003",
        "itemType": "Edge",
        "access": "permissive",
        "bicycle": "yes",
        "iso_country_code": "PHL",
        "osmIdentifier": "90980214",
        "source": "Google",
        "last_edit_user_name": "leven",
        "last_edit_time": "1542450771000",
        "last_edit_user_id": "1255860",
        "name": "Alsace St",
        "lanes": "2",
        "highway": "residential",
        "last_edit_version": "2",
        "foot": "yes"
      }
    }
  ],
  "properties": {
    "id": "909802140000019098021400000290980214000003",
    "instructions": "1. OSM feature 90980214 has invalid tags.\n2. The following element has an illegal source.",
    "identifiers": [
      "Edge90980214000001",
      "Edge90980214000003",
      "Edge90980214000002"
    ],
    "generator": "InvalidTagsCheck",
    "timestamp": "Wed Apr 07 14:08:03 UTC 2021"
  },
  "fix_suggestions": {}
}
```

config/configuration.json
```diff
@@ -4,7 +4,7 @@
       "org.openstreetmap.atlas.checks.validation"
     ],
     "type": "org.openstreetmap.atlas.checks.base.BaseCheck",
-    "enabled.value.default": true
+    "enabled.value.default": false
```
atiannicelli, line 7: careful - this needs to be reverted.

config/configuration.json
```diff
@@ -1270,14 +1288,15 @@
     }
   },
   "WaterWayCheck": {
+    "enabled": true,
```
atiannicelli, line 1291: needs to be removed.

scripts/taginfo/downloadTagInfo.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
```
atiannicelli, line 1: needs to be made executable.

gradle.properties
```diff
@@ -14,14 +14,20 @@ profile=local
 ## Local (default)
 checks.local.sharded=true
 checks.local.configFiles=file://@ROOTDIR@/config/configuration.json
-checks.local.input=file://@BUILDDIR@/example/data/atlas/
+checks.local.input=file://@BUILDDIR@/../atlas/
+#checks.local.input=hdfs://10.135.134.89:9000/tsmock/atlas/
```
atiannicelli, line 18: Guessing you want to take this out of the pr
```suggestion
checks.local.input=file://@BUILDDIR@/example/data/atlas/
```

gradle.properties
```diff
@@ -14,14 +14,20 @@ profile=local
...
+checks.local.countries=AIA,CYM,DMA,SMR,NZL
 #checks.local.pbfBoundingBox=
 checks.local.savePbfAtlas=true
 checks.local.compressOutput=false
 checks.local.startedFolder=file://@BUILDDIR@/example/tmp/
 checks.local.cluster=local
+#checks.local.cluster=spark://taylorrpi1:7077
 checks.local.multiAtlas=true
-checks.local.sparkOptions=spark.executor.memory->4g,spark.driver.memory->4g,spark.rdd.compress->true
+checks.local.sparkOptions=spark.executor.memory->2g,spark.driver.memory->1g,spark.rdd.compress->true,spark.default.parallelism->8
```
atiannicelli, line 33: I think you want to revert this entire file, right?

gradle/execute.gradle
```diff
@@ -50,7 +50,7 @@ task runChecks(type: JavaExec, dependsOn: 'assemble', description: 'Executes the
     classpath('./config/log4j')
 
     // uncomment to change jvm args
-    // jvmArgs(["-Xmx8g","-Xms8g"])
+    jvmArgs(["-Xmx2g","-Xms2g", "-XX:+HeapDumpOnOutOfMemoryError"])
```
atiannicelli, line 53: did you really want this?

settings.gradle
```diff
@@ -1 +1,3 @@
 rootProject.name = 'atlas-checks'
+//includeBuild '../atlas'
```
atiannicelli, line 2: Can you remove these comments?

atiannicelli (changes requested):
A few changes that leaked in to this pr I think that need to be reverted.

config/configuration.json
tsmock: Fixed

scripts/taginfo/downloadTagInfo.py
tsmock: Done

config/configuration.json
```diff
@@ -1290,6 +1308,9 @@
       "defaultPriority": "LOW"
     }
   },
+  "ElevationUtilities": {
+     "elevation.path": "extra/elevation"
+  },
```
atiannicelli, line 1313: I see that you removed elevation.path in the ElevationUtilities md file... but added it here. Should this be elevation.prefix?

atiannicelli:
@tsmock we definitely need some logic in here to prevent the proposed changes from removing some deprecated tags. The example below (propose removing "building=unclassified") would leave no tags on this area so it would become a line. Definitely need to either not propose the change and let the mapper figure out what the proper tag is OR propose the tag that the wiki suggests.

https://www.openstreetmap.org/way/629541343
Screen Shot 2021-04-12 at 8 36 32 AM

tsmock:
> The example below (propose removing "building=unclassified") would leave no tags on this area so it would become a line. [...]
> Definitely need to either not propose the change and let the mapper figure out what the proper tag is OR propose the tag that the wiki suggests

It is going to have to be the former. The latter requires a change in the wiki data (done for this specific item).

atiannicelli:
But... I think we _could_ propose changes for tags that are prohibited on a feature type. Can you add proposed changes to this type of issue?

Screen Shot 2021-04-12 at 8 57 13 AM

tsmock:
> But... I think we could propose changes for tags that are prohibited on a feature type. Can you add proposed changes to this type of issue?
Can you give me a link to the task? It _should_ be asking for removal.

config/configuration.json
```diff
@@ -315,6 +315,24 @@
...
+    "tag": {
+      "usage.min": 100,
+      "percentage_of_key_for_popular": 10
+    },
+    "wikidata.tag_removal": ["abandoned", "deprecated", "imported", "obsolete", "rejected"],
+    "challenge": {
+      "description": "Tasks containing features with tags containing missing, conflicting, incorrect or illegal values",
+      "blurb": "Features with invalid tags",
+      "instruction": "Open your favorite editor and check that the listed tags are correct.",
+      "difficulty": "Medium",
```
atiannicelli, line 332: Can you change this to "EXPERT"?
```suggestion
      "difficulty": "EXPERT",
```
I think that we want this to be noted as an expert challenge and if I am reading the code correctly the correct value is "EXPERT".
https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/maproulette/data/ChallengeDifficulty.java#L10

tsmock: And done. :)

config/configuration.json
```diff
@@ -315,6 +315,24 @@
...
+      "taginfo": "extra/taginfo-db.db",
+      "wikidata": "extra/wikidata.db"
+    },
+    "tag": {
+      "usage.min": 100,
+      "percentage_of_key_for_popular": 10
+    },
+    "wikidata.tag_removal": ["abandoned", "deprecated", "imported", "obsolete", "rejected"],
+    "challenge": {
+      "description": "Tasks containing features with tags containing missing, conflicting, incorrect or illegal values",
```
atiannicelli, line 329: Please expand this description. For example:

> This challenge contains tasks that identify features with OSM Tags (https://wiki.openstreetmap.org/wiki/Tags) that contain invalid tag keys or missing, conflicting, incorrect or illegal tag values. In many tasks in this challenge the task will identify an illegal or questionable tag key or value and provide a link to the Control Wiki Page (wiki page with identification string starting with Q, following with 3 to 4 digits, e.g. https://wiki.osm.org/Item:Q1189) that is used to define that tag’s characteristics in a computer readable format. That Control Wiki Page should contain a link to the Documentation Wiki Page for that tag that is a human readable page that discusses the proper use of the tag and other useful information. 
> For some tasks in this challenge the task may indicate a tag value that is valid or widely used that is not documented in the wiki. For these cases the mapper should not make any changes to the OSM data, but should instead consider adding documentation to the wiki to validate the tag value officially. 
> For more information on the Atlas Check that is used to create this challenge please go to https://github.com/osmlab/atlas-checks/tree/dev/docs/checks/genericTagCheck.md

atiannicelli (changes requested):
@tsmock, this is looking great! As discussed after QA results were reviewed could you add a bit more instruction to the "instructions" for these tasks? I have two examples below:

### More instruction in Tasks (For all places where we direct the user to go to the QXYZ site)

The current instruction for one of the tasks is:

> OSM feature 259288929 has invalid tags
> created_by=JOSM is prohibited on node (see https://wiki.osm.org/Item:Q1189 )

To better instruct the mapper as to what to do (instead of just identifying the issue) the following text would be proposed:

> OSM feature 259288929 has invalid tags
> created_by=JOSM is prohibited on a node. To determine whether to remove or change this tag please see this tag’s control wiki page, https://wiki.osm.org/Item:Q1189 or associated Documentation Wiki Page for this tag (found in the control wiki page).

### Undocumented popular Tag Values

The current instructions for undocumented tag values is:

> OSM feature 6135364595 has invalid tags
> cuisine=waffle is popular enough to merit wiki documentation (177 uses)

To instruct the mappers as to what to do here let’s change this to:

> OSM feature 6135364595 has invalid tags
> cuisine=waffle is not currently documented. Its global popularity (177 uses) may merit adding wiki documentation for this value. Please consider updating the wiki control page  https://wiki.osm.org/Item:QXYZ or adding a new page if necessary.
>

config/configuration.json
tsmock: Fixed in c3810a7a81523dda96ea4614be6b76da6ca108f3

atiannicelli (approved):
Looks awesome! Thanks for updating the instructions.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,939 @@
...
+                                tag.getValue(), relation.getOsmIdentifier(), checkInfo.getId()),
+                        Collections.emptyList());
+                return true;
+            }
+            final Collection<String> additionalInstructions = new TreeSet<>();
+            for (final RelationMember relationMember : relation.members())
+            {
+                // Empty roles should exist in WikiData
+                final String role = relationMember.getRole();
+                final WikiData roleInfo = this
```
atiannicelli, line 537: One more to convert to var
```suggestion
                final var roleInfo = this
```

tsmock: It is always fun when you miss _one_.

atiannicelli (dismissed):

src/main/java/org/openstreetmap/atlas/checks/utility/KeyFullyCheckedUtils.java
```diff
@@ -0,0 +1,181 @@
...
+        final List<TaggableFilter> filters = new ArrayList<>();
+        final List<Predicate<Taggable>> predicateFilters = new ArrayList<>();
+        final var classLoader = InvalidTagsCheck.class.getClassLoader();
+        Set<ClassInfo> classInfoList;
+        try
+        {
+            classInfoList = ClassPath.from(classLoader)
+                    .getTopLevelClassesRecursive(InvalidTagsCheck.class.getPackageName());
+        }
+        catch (final IOException e)
```
Bentleysb, line 54:
```suggestion
        catch (final IOException exception)
```

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/BigNodeBadDataCheckTest.java
```diff
@@ -73,6 +73,11 @@ private boolean runTest(final Atlas atlas, final Configuration config,
 
         bigNodes.forEach(bigNode ->
         {
+            if (bigNode.nodes().stream().mapToLong(node -> node.getOsmIdentifier())
+                    .anyMatch(number -> number == 812034701L))
+            {
+                logger.info("found something");
+            }
```
Bentleysb, line 80: I'm guessing this was to test something?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,940 @@
...
+            // different object types
+            this.markAsFlagged(this.getUniqueOSMIdentifier(object));
+
+            // Create a flag with generic instructions
+            final String instruction = this.getLocalizedInstruction(
+                    FALLBACK_INSTRUCTIONS.indexOf(INSTRUCTION_INVALID_TAGS),
+                    object.getOsmIdentifier());
+            // If the object is an edge add the edges with the same OSM id
+            final CheckFlag flag = (object instanceof Edge)
+                    ? this.createFlag(((Edge) object).getMainEdge(), instruction)
```
Bentleysb, line 253: The main edge does not encompass all edges with the same ID. It is just the main direction of a way setioned edge. 
 ```suggestion
                    ? this.createFlag(new OsmWayWalker((Edge) object).collectEdges(), instruction)
```

**Review by Bentleysb — COMMENTED** [review]
This is quite the check, and surrounding support system for it. Overall this is supper cool. I just had a few little comments.

File: docs/checks/genericTagCheck.md [hunk 15]
```diff
@@ -0,0 +1,89 @@
+# Generic Tag Check
+
+This check uses TagInfo and WikiData databases to look for invalid, deprecated, or otherwise invalid tags.
+In addition, there is a fallback check that uses TagInfo data and WikiData items to determine if a tag is valid.
+The fallback check can be enabled or disabled by setting `"fallback"` to `false`. The location of the databases are
+set in a `"db"` section with either `"wikidata"` or `"taginfo"` as the key, and then the path as the value. This
+_must_ be changed in Microsoft Windows environments (due to differences in file path separators). There is an additional
+key for the `wikidata` database, `"wikidata.tag_removal"` in main InvalidTagsCheck check configuration section. This
```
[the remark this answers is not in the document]
tsmock: Oops. I was originally modifying the `InvalidTagsCheck`, but later decided to move it into its own check.

scripts/taginfo/downloadTagInfo.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+import requests
+import argparse
+import bz2
+import sqlite3
+from tqdm import tqdm
+import pandas
```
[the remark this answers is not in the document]
tsmock: Yep. I'm going to do a minimal `requirements.txt` file (mostly to avoid having to manually update the `requirements.txt` all the time).

scripts/wikidata/requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+click>=7.1.2
+coverage>=5.3
+flake8>=3.8.3
+idna>=2.10
+importlib-metadata>=1.7.0
+lockfile>=0.12.2
+mccabe>=0.6.1
+msgpack>=1.0.0
+mypy-extensions>=0.4.3
+pathspec>=0.8.0
```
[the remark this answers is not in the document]
tsmock: Oops. I'm adding that now, and doing the same to the requirements.txt file as I'm doing for the `downloadTagInfo.py` requirements.txt file.

scripts/wikidata/get_wikidata.py
```diff
@@ -0,0 +1,479 @@
...
+    special = populate_special()
+    filecache = FileCache(".web_cache", forever=True)
+    session = CacheControl(requests.Session(), filecache)
+    session.headers.update({"User-Agent": user_agent})
+    db = DataBase("wikidata.db")
+    table = "wiki_data"
+    db.add_table(table)
+    skipped = 0
+    index = 1
+    progress = tqdm()
```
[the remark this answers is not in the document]
tsmock: I've gone ahead and added the `desc` param to the `tqdm` calls.

config/configuration.json
```diff
@@ -1336,6 +1354,9 @@
       "defaultPriority": "LOW"
     }
   },
+  "ElevationUtilities": {
```
sayas01, line 1357: Where is this being used? Wouldn't it cause a conflict with https://github.com/osmlab/atlas-checks/blob/fe33333d169d44c3bb679decd0c7f055d3ed27dd/config/configuration.json#L1373 ?

docs/checks/genericTagCheck.md
```diff
@@ -0,0 +1,89 @@
+# Generic Tag Check
```
sayas01, line 1: Thanks for adding a well written doc for the check!

docs/checks/invalidTagsCheck.md
```diff
@@ -78,16 +78,25 @@ In [Atlas](https://github.com/osmlab/atlas), OSM elements are represented as Edg
...
+If the tag does _not_ exist in OSM WikiData AND the key does exist and specifies that all values are `well-known`, an instruction is generated.
+If the tag does _not_ exist in OSM WikiData AND the key does exist AND the key specifies a value validation regex AND the value fails the regex, an instruction is generated.
+If the tag does exist in the OSM WikiData AND the key has been marked as unwanted, an instruction is generated.
+If the tag does _not_ exist in OSM WikiData AND the key does not exist THEN the TagInfo database is queried. If the uses of the tag is low (less than the `tag.usage.min` variable, or `100` if the variable is unset), an instruction for an editor is generated, otherwise an instruction for a wiki editor is generated.
+
+For each filter a feature passes an instruction is added to the flag. The tag keys from the TaggableFilter
+are pulled out and added to the instruction to give guidance on where the problem is.
+
+To learn more about the code, please look at the comments in the source code for the check:
```
sayas01, line 101: Looks like this part was added by mistake here?

docs/checks/invalidTagsCheck.md
```diff
@@ -1,17 +1,17 @@
 # Invalid Tags Check
 
 This check flags features for invalid tags based on a list of TaggableFilters and RegexTaggableFilters. The list of filters for this check are passed through the config and/or through resource files.
-The two configurable boolean values, "filters.resource.append" and "filters.resource.override", 
+The two configurable boolean values, "filters.resource.append" and "filters.resource.override",
```
sayas01, line 4: Are these changes to resolve some sonar issues?

docs/utilities/elevationUtilities.md
```diff
@@ -5,12 +5,14 @@ This is a general utilities class that allows checks to get elevation data. The
...
 
+The `elevation.prefix` is used to help filter files on Spark (all extra data files are put in the _same_ directory).
+
 ```json
 {
     "ElevationUtilities": {
         "elevation.srtm_extent": 1.0 (degree),
         "elevation.srtm_ext": "hgt" (file extension),
-        "elevation.path": "elevation"
+        "elevation.prefix": "elevation-"
```
sayas01, line 15: Are these changes associated with GenericTagCheck addition?

docs/checks/invalidTagsCheck.md
tsmock: No. My text editor (vim) is set up to autotrim trailing spaces on lines, which doesn't always make sense (specifically, for markdown files).

docs/checks/invalidTagsCheck.md
tsmock: It was left by mistake, yes. I was originally putting the code in the `InvalidTagsCheck` file, and later decided to move it out.

src/main/java/org/openstreetmap/atlas/checks/database/wikidata/WikiDataItem.java
```diff
@@ -0,0 +1,138 @@
...
+     * Get a well-known wiki data item from a value
+     *
+     * @param value
+     *            The value to look for
+     * @return A WikiDataItem or {@code null}
+     */
+    public static WikiDataItem fromValue(final String value)
+    {
+        WikiDataItem rItem = Stream.of(WikiDataItem.values())
+                .filter(p -> p.getId().equalsIgnoreCase(value)).findFirst().orElse(null);
```
sayas01, line 64: nit: please avoid single letter variable

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/BigNodeBadDataCheckTest.java
tsmock: Yes. I think I was having issues with a test failing somewhere when I was updating to newer `atlas`/`atlas-generator` versions.

config/configuration.json
tsmock: Yep. I probably missed it. Even though it is only 10 lines or so down.

docs/utilities/elevationUtilities.md
tsmock: No. I think it is a holdover from when I was fiddling with methods for transferring files between spark nodes.

tsmock:
Before merging this PR, can whoever intending to do the merge let me know in advance? I'd like to squash the commits so that there is a meaningful message (for example, `BigNodeBadDataCheckTest: Revert to dev` really doesn't need to be kept).

I'd do it right now, but @sayas01 or @Bentleysb might have the branch checked out for review.

Bentleysb:
@tsmock, we squash the commits as part of the merge, and the PR title becomes the commit message (see the dev branch commit log).

tsmock:
@Bentleysb : I am aware of that. But which makes more sense and is more readable?

```
    InvalidMultiPolygonRelationCheck: Extract AtlasToOsmType code
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    GenericTagCheck: Add generic test for tag checking (using OSM Wiki Data and TagInfo)
    
    AbbreviatedNameCheck: Improve validCheckForObject
    * Only objects with names are valid checks.
    
    FeatureChange: Add interfaces for remove/replace tag changes
    
    GenericTagCheck: Check for unpopular tags with taginfo
    * Implement Autocloseable and check for that in RunnableCheck (if
      implemented, close)
    * Add SQLiteUtils in an effort to allow the DB connection to be
      serialized (and significantly reduce time from reopening db)
    * Check for unwanted tags on objects using OSM Wiki Data
    * Use TagInfo information to expound upon issues detected by OSM Wiki
      Data.
    * OSM Wiki Data is used for regex checking, when the Wiki Data has a
      regex.
    * Account for P17 (redirect to) when a bad tag is encountered.
    
    SQLite:
    * Dependencies: Add SQLite
    * Cache gets for DB
      * Assumption: DB doesn't change during the run
    
        This brought the runtime on a MacBook 2017 (i7 w/ 16GB ram) from 14
        hours to 2 hours (and a bit). The runtime for the dev/master branch was
        also about 2 hours (and a bit).
    
        The countries run were AIA, CYM, DMA, NZL, and SMR. There was no real
        noticeable difference with BLZ.
    * Store a copy of the retrieved data without null values in SQLiteUtils
    * Store file fetcher for use on computation nodes
    
    TagInfo: Add download script
    WikiData: Add download script
    * Update older entries and write changes to a file
    
      This means that changes can be inspected and verified (future task:
      insert into an HTML page with links so people can more easily tell if
      the change is relevant or not).
    * When duplicate unique keys appear, print out what was being inserted
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    FIXUP: GenericTagCheck: Improve wording for some tasks
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    FIXUP: GenericTagCheck: Improve wording for some tasks
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    GenericTagCheck: Fix tests
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    FIXUP: Formatting
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    SonarLint: Use var where possible
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    Sonar: FIXUP: Use var where possible
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    GenericTagCheck: Account for keys that are not well-defined with tags
    
    This fixes and tests for an issue where a key would not have P9=Q8
    (i.e., not well-defined), and thus would not check for the tag itself.
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    Sonar: FIXUP: Use var where possible
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    GenericTagCheck Docs: Update wording
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    TagInfo/WikiData scripts: Add/update requirements.txt
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    KeyFullyCheckedUtils: Expand e -> exception
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    DownloadTagInfo: Add docstring for script, add desc to tqdm
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    Docs: Revert invalidTagsCheck to dev branch
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    WikiData: Modify get script
    
    * Add descriptions to tqdm
    * Print out duplicates, so that they are easier to remove
    * Faster updating after first download
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    BigNodeBadDataCheckTest: Revert to dev
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    GenericTagCheck: Use OsmWayWalker for edges in flags
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    WikiDataItem: p -> entry (variable length)
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    Configuration: Remove extraneous ElevationUtilities stanza
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>
    
    Docs: ElevationUtilities: Revert to dev
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>

```
or
```
    GenericTagCheck: Add generic test for tag checking (using OSM Wiki Data and TagInfo)
    
    AbbreviatedNameCheck: Improve validCheckForObject
    * Only objects with names are valid checks.
    
    FeatureChange: Add interfaces for remove/replace tag changes
    
    GenericTagCheck: Check for unpopular tags with taginfo
    * Implement Autocloseable and check for that in RunnableCheck (if
      implemented, close)
    * Add SQLiteUtils in an effort to allow the DB connection to be
      serialized (and significantly reduce time from reopening db)
    * Check for unwanted tags on objects using OSM Wiki Data
    * Use TagInfo information to expound upon issues detected by OSM Wiki
      Data.
    * OSM Wiki Data is used for regex checking, when the Wiki Data has a
      regex.
    * Account for P17 (redirect to) when a bad tag is encountered.
    
    SQLite:
    * Dependencies: Add SQLite
    * Cache gets for DB
      * Assumption: DB doesn't change during the run
    
        This brought the runtime on a MacBook 2017 (i7 w/ 16GB ram) from 14
        hours to 2 hours (and a bit). The runtime for the dev/master branch was
        also about 2 hours (and a bit).
    
        The countries run were AIA, CYM, DMA, NZL, and SMR. There was no real
        noticeable difference with BLZ.
    * Store a copy of the retrieved data without null values in SQLiteUtils
    * Store file fetcher for use on computation nodes
    
    TagInfo: Add download script
    WikiData: Add download script
    * Update older entries and write changes to a file
    
      This means that changes can be inspected and verified (future task:
      insert into an HTML page with links so people can more easily tell if
      the change is relevant or not).
    * When duplicate unique keys appear, print out what was being inserted
    
    Signed-off-by: Taylor Smock <tsmock@fb.com>

```

Bentleysb:
@tsmock, the later is defiantly better, but my point is that it is somewhat irrelevant. Once we merge any commit messages you made will not show up in the dev branch, at which point you can do what you like on your branch.

scripts/taginfo/downloadTagInfo.py
```diff
@@ -0,0 +1,102 @@
...
+def convert_to_parquet(database: str, save_directory: str="parquet"):
+    """
+    Convert an sqlite db to parquet files
+    """
+    with sqlite3.connect(database) as conn:
+        path = os.path.join(save_directory, "taginfo")
+        if not os.path.exists(path):
+            os.makedirs(path)
+        for table in conn.execute("SELECT name FROM sqlite_master WHERE type='table'").fetchall():
+            t = table[0]
```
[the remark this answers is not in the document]
Bentleysb: @tsmock

scripts/wikidata/get_wikidata.py
```diff
@@ -0,0 +1,479 @@
...
+    logging.error(f"Unsupported data value: {datavalue}")
+
+
+def parse_claim(claim: list, special: dict = populate_special()):
+    """
+    Parse a claim from a WikiData dict
+    >>> import json
+    >>> json_data = json.loads('[{"mainsnak":{"snaktype":"value","property":"P2","hash":"9c1b9f9b61faedefa272a9c8c980faba6cefe7d5","datavalue":{"value":{"entity-type":"item","numeric-id":7,"id":"Q7"},"type":"wikibase-entityid"},"datatype":"wikibase-item"},"type":"statement","id":"Q4108$D0688651-E4D8-41AF-B1C8-69C4FD0A9654","rank":"normal"}]')  # noqa: E501
+    >>> parse_claim(json_data)
+    'key'
```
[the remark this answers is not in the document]
Bentleysb: @tsmock

scripts/wikidata/get_wikidata.py
```diff
@@ -0,0 +1,479 @@
...
+    >>> parse_claim(json_data)
+    'Key:ref:EU:ENTSOE EIC'
+
+    >>> json_data = json.loads('[{"mainsnak":{"snaktype":"value","property":"P13","hash":"1352cfc4840bb96337bfc822ceb0255686e3e9e7","datavalue":{"value":"[1-6][0-9][ATVWXYZ][0-9A-Z-]{12}[0-9A-Z](;[1-6][0-9][ATVWXYZ][0-9A-Z-]{12}[0-9A-Z])*","type":"string"},"datatype":"external-id"},"type":"statement","id":"Q4108$a03623d5-4420-9b6c-9787-69a8bbb2c605","rank":"normal"}]')  # noqa: E501
+    >>> parse_claim(json_data)
+    '[1-6][0-9][ATVWXYZ][0-9A-Z-]{12}[0-9A-Z](;[1-6][0-9][ATVWXYZ][0-9A-Z-]{12}[0-9A-Z])*'
+
+    """
+    rrlist = []
+    for i in claim:
```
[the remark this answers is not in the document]
Bentleysb: @tsmock

scripts/wikidata/get_wikidata.py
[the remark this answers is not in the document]
tsmock: These are doctests ( https://docs.python.org/3/library/doctest.html ).

scripts/taginfo/downloadTagInfo.py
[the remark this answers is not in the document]
tsmock: Sorry about missing these -- GitHub had them hidden (it said "5 hidden conversations", and I wrongly assumed that you had created them, and then decided that they weren't needed).

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,941 @@
...
+        }
+    }
+
+    /**
+     * Check if a wiki data item for a tag redirects to another tag, and if so, create an
+     * instruction and change for that.
+     *
+     * @param instructions
+     *            The instructions collection to add to
+     * @param changes
```
sayas01, line 421: unused param

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,941 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Do relation specific checks
+     *
+     * @param instructions
+     *            Instructions to add to
+     * @param The
```
sayas01, line 507: @param object

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,941 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Check the status of a tag
+     *
+     * @param instructions
+     *            The instruction collection to append instructions to
+     * @param changes
```
sayas01, line 564: unused param

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,941 @@
...
+        return false;
+    }
+
+    /**
+     * Check for instances where a key is well-known but the value is not one of the well-known
+     * values.
+     *
+     * @param instructions
+     *            The instructions to add to
+     * @param changes
```
sayas01, line 598: unused param

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -0,0 +1,941 @@
...
+            }
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Check the tag against wiki data for cases where it should not appear on the object.
+     *
+     * @param changes
```
sayas01, line 644: unused param

src/main/java/org/openstreetmap/atlas/checks/utility/AtlasToOsmType.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.openstreetmap.atlas.checks.validation.relations.InvalidMultiPolygonRelationCheck;
+import org.openstreetmap.atlas.geography.atlas.items.ItemType;
+
+/**
+ * Convert Atlas ItemTypes into OSM types
+ *
+ * @author Taylor Smock, extracted from {@link InvalidMultiPolygonRelationCheck}.
+ */
+public final class AtlasToOsmType
```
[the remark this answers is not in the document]
tsmock: It probably could be. I think I didn't do that since I've got a few projects with various moving parts, and I didn't want to deal with yet another one at the time, and then forgot about it.

I've had too many PR's that depend upon other PR's that depend upon yet another PR...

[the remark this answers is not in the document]
Bentleysb: This can defiantly be something that is done later.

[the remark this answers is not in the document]
sayas01: Yp!!

scripts/taginfo/downloadTagInfo.py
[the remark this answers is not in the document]
Bentleysb: No worries. I guessed something like that might be the case.

scripts/wikidata/get_wikidata.py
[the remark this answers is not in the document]
Bentleysb: Interesting, I had not encountered that before. Cool!

Bentleysb (approved):
Looks good. Thanks for all the hard work on this one!

sayas01 (approved):
Great job on this one @tsmock ! This is well documented and clean. Thanks for addressing all the comments. 🚀

PR #360, opened 2020-09-14
Suddenhighwaychange (DNM)

### LINKED ISSUE #344: SuddenHighwayTypeChangeCheck (Osmose 1090)
**Description**
This checks for highway tag changes greater than one classification.

**Requirements**
Must be a car-navigable highway Edge 
Flag edges where the edge requires a link highway tag 

**Use Cases**
True Positive
Case 1:
The VALID case where an edge is connected to another edge of different classification NOT through a *_link edge 
Highway=primary Edge connects directly to a highway=tertiary Edge 
This should likely be set as highway=primary_link 
https://www.openstreetmap.org/way/698449634 

False Positive
 Case 1: 
Highway=primary Edge connects directly to a highway=residential Edge 
This is correct due to a gated military base causing a valid drop in classification along the path 
https://www.openstreetmap.org/way/290065445 

**Supported regions**
This should be fine in all countries

reichg:
### Description:

This check flags edges that should be tagged as a link.

### Test Results:

| Country | Total Objects Checked | False Positives | FP % |
|----------|:-------------------------:|:---------------:|:-------:|
|   AUS     |              134                  |           24           |    18%  |
|  ARG      |              100                  |           13           |    13%  | 
|  DEU      |              20                  |           2           |    10%  |
|  NZL      |              13                      |           1           |    7.5%  |
|  TUR     |                120                  |            23         |     19%   |

Some of the false positive rates are a bit high but there some scenarios that are hard to resolve without affecting the true positive counts. Would like to test on larger countries like CAN/BRA/USA etc but need ShardedIntegrityChecksSparkJob for that. Currently have issues running that job on windows. 

IDEAS WELCOME! 

Resolves #344


reichg:
closing this as we have reflected on this check and concluded we have strayed away from the osmose issue objective.

PR #368, opened 2020-09-18
Extend InvalidTagsCheck with regex filtering for illegal sources

mm-ciub:
### Description:

The InvalidTagsCheck has been extended to also accept RegexTaggableFilters, allowing for certain tags to be checked against specific regex patterns.
The main reason for this extension is to verify the source tag for illegal sources like "Google" or "Here".
The check can be extended with other filters containing different tag names and regex patterns to be checked against.
https://github.com/osmlab/atlas-checks/issues/334

### Potential Impact:

No impacts.

### Unit Test Approach:

Added unit tests that check the sources for nodes and edges with illegal tags such as maps link or Here maps.

### Test Results:

The check was run on shards from AIA, AUS, CYM, DMA, NZL, TUR, with results only for AUS, TUR and NZL for the illegal source filter.

|  AUS,TUR,NZL   |    47         |    38      |     80%       |   35 |   3 | 8%   |

There are still some open cases if certain values should be allowed or not, for example google drive links


danielduhh:
Is this ready for another review? @mm-ciub

danielduhh:
If so, do you mind fixing the build issues 🙏

Bentleysb:
This looks like an awesome addition. If you could fix the 2 code smells noted by sonar that would be great.

mm-ciub:
The smell has been dealt with.

danielduhh (approved):

src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidTagsCheck.java
```diff
@@ -34,50 +39,57 @@
...
- * "filters.resource.append.override" and "filters.resource.append". If the
- * "filters.resource.append.override" key is set to true, only the filters passed through config are
- * used to flag the atlas features. If the "filters.resource.append" is set to true, the filters
- * passed through the config are appended to the default filters that are in the "invalidTags.txt"
- * resource file.
+ * {@link AtlasEntity} classes to check and a {@link TaggableFilter} or a
+ * {@link RegexTaggableFilter} to test objects against. If a feature is of one of the given classes
+ * and passes the associated filter then it is flagged. In addition to the taggable filters, there
+ * are two configurable boolean values, "filters.resource.append.override" and
+ * "filters.resource.append". If the "filters.resource.append.override" key is set to true, only the
```
sayas01, line 46: Looks like the java docs were not updated when these params were modified in a previous PR. The current config just has "filters.resource.override". Since this check is going to be updated for https://github.com/osmlab/atlas-checks/issues/377, we can update these in the upcoming PR. @mm-ciub

sayas01 (approved):
Nice enhancement! 👍

PR #370, opened 2020-09-22
countryShardMapFromShardFiles Reads too Much

Bentleysb:
### Description:
This is to fix a bug with `IntegrityChecksCommandArguments.countryShardMapFromShardFiles()`. This method reads the input atlas files to assign shards to countries. For each country it was recursively reading all files in the input path. This fixes it so for each country it only reads the country's folder, and it only reads the file name instead of opening the file. 

### Potential Impact:
This should increase the pre-spark speed of the job.

### Unit Test Approach:

None

### Test Results:
Tested locally and observed that only the file names for a single country were read in for each country.
Tested with remote files and saw a 73% decrease in overall runtime.


danielduhh (approved):
💯

sayas01 (commented):
Thanks @Bentleysb ! Could you also fix the one code smell?

danielduhh (approved):

sayas01 (approved):

Issue #369, opened 2020-09-22
LevelCrossing Check - Requirements


atiannicelli:
Developers: Alex Iannicelli 
Analyst: Yunzhi Lin
Date: 09/22/2020

**Description**
Detect and flag objects that should not be tagged with the railway=level_crossing tag. Create missing railway/highway intersections. Tag railway/highway intersections that are not properly tagged as railway=level_crossing.

**Requirements**
There are three railway intersection scenarios that this check looks for:
1.  railway intersection node is missing.
 - Is a `Line`
 - Has `railway` tag, values include `rail`, `tram`, `disused`, `preserved`
 - Intersects with a car navigable highway on the same layer
 - No node exists at intersection
2) railway intersection node exists, but railway=level_crossing tag is missing on the intersection node.
 - Is an `Line`
 - Has `railway` tag, values include `rail`, `tram`, `disused`, `preserved`
 - Intersects with a car navigable highway on the same layer
 - node exists at location, but `railway` tag doesn't exist or is not equal to `level_crossing`
3) a railway=level_crossing tag exists, 
      a) and this tag is on a node, but is lacking of either highway or railway going through the node (osmose 7090) 
      - Is an `Node` or a `Point`
      - Has `railway` tag, values include `level_crossing`
      - Count number of Edge with `highway` tag that is car navigable.
      - Count number of Line with `railway` tag values including `rail`, `tram`, `disused`, or `preserved`
      - Either Edge or Line count is 0.
      b) but not on a node, instead, on the related way features (osmose 9015)
      - Is not an `Node` or a `Point`
      - Has `railway` tag, values include `level_crossing`

**Use Cases**
*Case 1: Bridge over Railway on same layer.*
 - Highway bridge goes over Railway, but layer missing on both railway and highway. The bridge should have a layer=1 tag
 - https://www.openstreetmap.org/way/80459517
![Screen Shot 2020-09-22 at 1 33 11 PM](https://user-images.githubusercontent.com/5851115/93916877-3032ea00-fcd8-11ea-923f-052989d313fd.png)

*Case 2: Railway crosses a highway with no level_crossing tag*
 - The intersection node (OSM ID: 273135212) is missing a railway=level_crossing tag. Add the appropriate tag to the node.
 - https://www.openstreetmap.org/node/273135212
[...]

PR #371, opened 2020-09-22
Add railway level_crossing check

atiannicelli:
### Description:

Add level_crossing atlas check and unit tests. Resolves Issue: https://github.com/osmlab/atlas-checks/issues/369 

### Potential Impact:

Other than the time it will take to run the test - None - just a new Check

### Unit Test Approach:

Unit test was designed to hit each of the possible flagged issues. Unit test "Rule" file was produced manually.

### Test Results:

|-----|------------|----------|-------------|----|----|------------------- |
| ARM | 128            | 128          | 100%           | 125| 2   | 2% |
| AFG | 4                 | 4             | 100%            | 4   | 0   | 0% |
| NZL | 193             | 193         | 100%           | 163 | 27 | 14% |
| THA | 213            | 213           | 100%          | 189  | 12 |  6% |
| VEN | 37              | 37           | 100%            | 35  | 0 |  0% |
| TOTAL | 575        | 575         | 100%            | 516 | 41 |  7% |

Map Roulette Challenge:
[AFG](https://maproulette.org/browse/challenges/14517)
[ARM](https://maproulette.org/browse/challenges/14518)
[THA](https://maproulette.org/browse/challenges/14525)
[VEN](https://maproulette.org/browse/challenges/14526)
[NZL - V1 With full QA](https://maproulette.org/browse/challenges/14523)
[NZL - V3 after fixes addressed issues](https://maproulette.org/browse/challenges/14710)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
```diff
@@ -0,0 +1,369 @@
...
+            final int issueType = this.isValidLevelCrossingNode(node);
+            if (Validators.isOfType(node, RailwayTag.class, RailwayTag.LEVEL_CROSSING)
+                    && issueType > 0)
+            {
+                // This is a node that is tagged with railway=level_crossing and is not a
+                // railway/highway intersection
+                return Optional.of(this
+                        .createFlag(object,
+                                this.getLocalizedInstruction(issueType, object.getOsmIdentifier()))
+                        .addFixSuggestion(FeatureChange.add(
```
Bentleysb, line 185: Nice!

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Flag all objects that are not nodes or points that are tagged with railway=level_crossing
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return an optional {@link CheckFlag} object that contains flagged issue details
+     */
+    private Optional<CheckFlag> flagNonNodeTagged(final AtlasObject object)
+    {
+        if (!(object instanceof Node || object instanceof Point)
```
Bentleysb, line 249:
```suggestion
        if (!(object instanceof LocationItem)
```

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheckTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.Test;
+import org.openstreetmap.atlas.checks.configuration.ConfigurationResolver;
+import org.openstreetmap.atlas.checks.validation.verifier.ConsumerBasedExpectedCheckVerifier;
+
+/**
+ * Unit tests for {@link LevelCrossingOnRailwayCheck}.
+ *
+ * @author atiannicelli
+ */
+public class LevelCrossingOnRailwayCheckTest
```
Bentleysb, line 14: Please add a some unit tests that use the ConfigurationResolver.inLineConfiguration() to check that the 2 configurables work as expected. And please test that the expected fix suggestions are created.

Bentleysb (commented):
The FP rate for this check is on the higher side, thought the average puts it within the acceptable range. Could you share some examples of the FPs? Just to have some record of what we can expect.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
```diff
@@ -0,0 +1,369 @@
...
+     * constructor
+     *
+     * @param configuration
+     *            the JSON configuration for this check
+     */
+    public LevelCrossingOnRailwayCheck(final Configuration configuration)
+    {
+
+        super(configuration);
+        this.layerDefault = this.configurationValue(configuration, "layer.default",
```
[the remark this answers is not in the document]
atiannicelli: Done

src/test/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheckTest.java
atiannicelli: Done. I added two new tests to specifically test the two config parameters. Also added verifications for fix suggestions.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
atiannicelli: Done

atiannicelli:
@Bentleysb Thanks for the review - I believe I have addressed all requested changes.

Bentleysb (dismissed):
LGTM

atiannicelli:
@Bentleysb you asked for some FP examples. There were some false positives with embedded trams in highways so we excluded that from the final version. The other false positives that we saw in this data set go into these groups:
1) level_crossing at miniature railroads: The daylight maps exclude miniature rails so these level crossings look like they shouldn't be there to the check, but actually they are valid because of the crossing of a miniature railroad in OSM data. So this issue should not show up when checking OSM base.
2) level_crossing at narrow_gauge railroads: The daylight maps exclude narrow_gauge rails so these level crossings look like they shouldn't be there to the check, but actually they are valid because of the crossing of a narrow_gauge railroad in OSM data. 
3) Our mappers identified a few places where roads run parallel to rails and do not cross, but in OSM the ways cross. These were flagged as issues in the check and identified as needing level_crossing nodes, but actually the ways just need to be adjusted not to cross.
4) Some false positives seemed to be due to a misunderstanding of the rules around abandoned rails. Some "false positives" flagged by our mappers were actually not false positives. Here is one example where the abandoned rail crosses a highway and there is a level crossing. Abandoned rails have no rails and therefor should not have level crossings so this flag is correct, but the mapper flagged it as a false positive. https://maproulette.org/admin/project/41568/challenge/14524/task/66476166/inspect
5) here is one where the railway correctly was given a "bridge" tag, but then they set the layer to 0. So the check flags that as a problem because it sees that it is on the same layer as the road it crosses. This is an issue, but just not exactly what was indicated in the flag: https://maproulette.org/admin/project/41568/challenge/14524/task/66475966/inspect
6) and some issues were flagged as a false positive by our mappers, but I think they are goot issues like this one: https://maproulette.org/admin/project/41568/challenge/14524/task/66476138/inspect

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
```diff
@@ -0,0 +1,369 @@
...
+     */
+    private Optional<CheckFlag> flagIncorrectlyTagged(final AtlasObject object)
+    {
+        if (object instanceof Node)
+        {
+            final Node node = (Node) object;
+
+            final int issueType = this.isValidLevelCrossingNode(node);
+            if (Validators.isOfType(node, RailwayTag.class, RailwayTag.LEVEL_CROSSING)
+                    && issueType > 0)
```
seancoulter, line 178: nit: `issueType` could be better served as an Enum to avoid referencing docs for explanation

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
```diff
@@ -0,0 +1,369 @@
...
+            {
+                // This is a node that is tagged with railway=level_crossing and is not a
+                // railway/highway intersection
+                return Optional.of(this
+                        .createFlag(object,
+                                this.getLocalizedInstruction(issueType, object.getOsmIdentifier()))
+                        .addFixSuggestion(FeatureChange.add(
+                                (AtlasEntity) ((CompleteEntity) CompleteEntity
+                                        .from((AtlasEntity) object)).withRemovedTag(RailwayTag.KEY),
```
seancoulter, line 187: just a thought. could the railway tag just have the wrong value, but the feature is still a legitimate "railway"? in which case we may want to not send up a fix suggestion

seancoulter (commented):
Nice, just a couple things I noticed
The Enum thing is totally optional as well

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
atiannicelli: In this check the feature is a node, so it is not a rail way. There are valid "railway" tags for nodes, but this specific check is determining that the "level_crossing" value is not correct. I guess there could be some other valid key value for this node, but I'm not sure how to determine that. Are you saying that you don't want a fix suggestion associated with this flagged issue?

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
atiannicelli: Addressed in latest version.

Bentleysb (approved):

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LevelCrossingOnRailwayCheck.java
seancoulter: Yeah after thinking about this for a bit we could still send up the fix suggestion. My only concern was that the editor might want to change the Railway tag value to something more appropriate if applicable, instead of deleting the tag,  but they have the ability to do that by just not applying the fix suggestion

seancoulter (approved):

PR #376, opened 2020-09-30
SCRIPTS: Add script to merge geojson files and format them in a MapRoulette tag fix friendly way

tsmock:
### Description:

This is a script I wrote awhile back to convert multiple geojson files into a single geojson file. To help with uploading to MapRoulette, there is also a command line flag that converts the geojson files into RFC7464 compliant line-by-line geojson. As a followup on #361 , I modified the script to also convert _basic_ `fix_suggestions` into MapRoulette cooperative challenges when outputting RFC7464 compliant geojson in order to help me validate the fix suggestion changes I made to the InvalidTagsCheck class.

There are a couple of `TODO` comments scattered throughout the file, mostly to do with investigating some edge cases (will Polygons _always_ be a way, or may they occasionally be a relation? -- looking at identifiers would probably work for this as well).

### Potential Impact:

N/A

### Unit Test Approach:

N/A

### Test Results:

N/A


scripts/merge_geojson/merge_geojson.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Get geojson files in a path for a flag
+    """
+    files = glob.glob(
+        os.path.join(country_path, "**", "*" + flag + "*"), recursive=True
+    )
+    return [g for g in files if re.search(r".*\.geojson(|\.gz)$", g)]
+
+
+def fix_suggestions_to_cooperativeWork(
```
danielduhh, line 30: This is great, thanks for adding it.
We're working on bringing in a pyatlas-checks library to house some MapRoulette related code integrated with the [maproulette-python-client](https://github.com/osmlab/maproulette-python-client).

Once we have that checked in, we can repurpose this code within that cli.

Also.. check out this ticket: https://github.com/osmlab/maproulette-python-client/issues/63. If you have time, it would be awesome if you could additionally add a Cooperative Challenge model for the python client.

danielduhh (commented):
This is nice!
Is there any reason why you're using the `geojson` output as opposed to the `.log` files that are already line-delimited? I know the log files unfortunately aren't separated by check and take some work to get the files organized that way.

tsmock:
>  Is there any reason why you're using the `geojson` output as opposed to the `.log` files that are already line-delimited? I know the log files unfortunately aren't separated by check and take some work to get the files organized that way.

Mostly because it was easier to work with organized checks to create MapRoulette tasks (when I first made the script, all I wanted to do was merge a specific check's files together -- I often run multiple checks instead of just the one I'm working on, to make certain I didn't accidentally break something). It eventually grew into what it is now, step by step. I added the `write_feature_rfc7464` function when I had some issues uploading rather large files to MapRoulette, and then I added the `fix_suggestions_to_cooperativeWork` function when I wanted to ensure that the fixes I was working on would "just work (tm)".

I didn't know we had a python client repository (but I didn't look either -- I wrongly assumed that everything related to atlas-checks would be in this repo).

scripts/merge_geojson/merge_geojson.py
```diff
@@ -0,0 +1,210 @@
...
+        fix_suggestions_to_cooperativeWork(feature)
+        if "fix_suggestions" in feature["properties"]
+        else geojson.FeatureCollection([feature])
+    )
+    with open(file_name, "ab") as fh:
+        # Line-by-line uses the 1E byte record separator
+        fh.write(ASCII_RECORD_SEPARATOR)
+
+    with open(file_name, "a") as fh:
+        string = geojson.dumps(fc)
```
Bentleysb, line 125: I am guessing this is just calling json.dumps, in which case we probably want to set `ensure_ascii=False` to maintain unicode characters.

scripts/merge_geojson/merge_geojson.py
```diff
@@ -0,0 +1,210 @@
...
+def main():
+    parser = argparse.ArgumentParser(
+        description="Create a line-by-line geojson file"
+    )
+    parser.add_argument("files", nargs="+")
+    parser.add_argument("-c", "--checks", nargs="?")
+    parser.add_argument("--line-by-line", action="store_true")
+    args = parser.parse_args()
+    for f in args.files:
+        logging.info(f)
```
Bentleysb, line 199: Nit: more context in this log would be good

scripts/merge_geojson/merge_geojson.py
```diff
@@ -0,0 +1,210 @@
...
+    with open(write_name, "w") as fp:
+        geojson.dump(featurecollection, fp)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Create a line-by-line geojson file"
+    )
+    parser.add_argument("files", nargs="+")
+    parser.add_argument("-c", "--checks", nargs="?")
```
Bentleysb, line 195: Based on the use of this argument it appears to be expecting a single check name. So I think it would make more sense to have the flag be --check.

Issue #364, opened 2020-09-18
[Feature Request] Create FeatureChange and ChangeDescription deserializers 


danielduhh:
**Is your feature request related to a problem? Please describe.**

- Create FeatureChange & ChangeDescription deserializer.

This deserializer will read CheckFlag GeoJSON outputs and convert the fix suggestions in the [ChangeDescription](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/change/description/ChangeDescription.java) model. This will allow us to leverage all that comes with the FeatureChange and ChangeDescription classes

Bentleysb:
Complete

PR #385, opened 2020-10-16
CheckFlag Deserializer and OpenStreetMapCheckFlagConverter

Bentleysb:
### Description:

This is [part two](https://github.com/osmlab/atlas-checks/issues/364) of changes to integrate fix suggestions into the atlas checks framework. [Part one](https://github.com/osmlab/atlas-checks/pull/361) added the ability to add fix suggestions to CheckFlags as FeatureChanges, and added the suggestions to the output logs in the form of feature change descriptors. This builds on that by adding utilities that lay the ground work for being able to export flags and suggestions into OSM like formats. 

There are two utilities that are the main focus of this PR. The first is the CheckFlagDeserializer, which is updated to fully deserialize a check flag, whereas previously it only rebuilt a minimal version of the flag. The second is the new OpenStreetMapCheckFlagConverter, which converts the flagged objects and fix suggestions of a CheckFlag into a more OSM compliant format. 

**CheckFlagDeserializer**
The CheckFlagDeserializer now full deserializes CheckFlags in line delimited geojson format (the flag output). The flag identifier, instructions, and challenge name (check name) are all simply re-added to the new flag. The flagged objects are reconstructed as CompleteEntities using the new GeoJsonFeatureToAtlasEntityConverter, and added to the new flag. The fix suggestions are rebuilt by using the CompleteEntities from the deserialized flagged objects as before views, and applying the change descriptors to that to create after views. This does mean that fix suggestions must have a matching flagged object in the flag to b deserialized. 

**OpenStreetMapCheckFlagConverter**
This is a new utility class that can can be used to make a CheckFlag more OSM compliant. It does this by removing Points that are also flagged as Nodes, and Merging Edges that share an OSM ID. This is done for both the flagged objects and the fix suggestions. When merging Edges it is assumed that everything other than the geometry is identical. It is expect this utility will be used as part of the process of converting CheckFlags into MapRoulette and OSM formats. 

**Other Changes**
CheckFlagEvent:
Made some constants public for use in both utilities.

CheckFlag:
Made the null identifiers constant public
Added an overloaded addObject method to allow direct adding of FlaggedObjects
Added the getRawInstructions method to allow access to the list of instructions without extra formating
Refactored a bit of code into the new FlaggedObject.getUniqueIdentifier method

FlaggedPoint/PolyLine/Relation:
Fixed a bug that was causing meta tags to get added to AtlasObjects instead of just Flagged Objects

### Potential Impact:

Anything that relies on the current version of the CheckFlagDeserializer may get more information that it was expecting. 

### Unit Test Approach:

Added new unit test for the CheckFlagDeserializer that cover all types of FlaggedObejcts and most varieties of fix suggestions. 
Added unit tests for the OpenStreetMapCheckFlagConverter that cover a variety of checkflags, especially those containing points, nodes, and edges to be merged.
Added a missing unit test for CheckFlagEvent to ensure REMOVE suggestions work properly. 

### Test Results:

Generated check results for a number of checks in a small country. Deserialized and reserialized the flags, and sampled through the results. Comparing the de/reserialized flags to the originals they appeared to all be the same.


src/main/java/org/openstreetmap/atlas/checks/utility/OpenStreetMapCheckFlagConverter.java
```diff
@@ -0,0 +1,209 @@
...
+            switch (entry.getValue().iterator().next().getBeforeView().getType())
+            {
+                // Merge edge FeatureChanges by concatenating the geometries of the before and after
+                // views and applying them to a copy of the first feature change
+                case EDGE:
+                    // Sort the before views and concatenate their geometries
+                    final Optional<PolyLine> concatenatedBeforePolyline = entry.getValue().stream()
+                            .map(FeatureChange::getBeforeView)
+                            .sorted(Comparator.comparing(AtlasEntity::getIdentifier))
+                            .map(entity -> ((Edge) entity).asPolyLine()).reduce(PolyLine::append);
```
seancoulter, line 119: "concatenate" when it comes to Polylines really means merging them together, correct?

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
```diff
@@ -51,15 +96,303 @@ public CheckFlag deserialize(final JsonElement json, final Type typeOfT,
...
+    {
+        // Copy the geometry to apply the changes to
+        final List<Location> newGeometry = Lists.newArrayList(afterView.getGeometry());
+
+        // Sort the geometry changes by position and type, as order of application matters
+        descriptors.stream()
+                .sorted(Comparator.comparingInt(descriptor -> Integer
+                        .valueOf(descriptor.get(POSITION).getAsString().split("/")[0])))
+                .sorted(Comparator.comparing(descriptor -> ChangeDescriptorType.valueOf(
+                        descriptor.get(GeoJsonConstants.TYPE).getAsString().toUpperCase())))
```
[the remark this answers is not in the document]
seancoulter: Also I believe the current sorting scheme is only sorting by type (as it's the terminal operation)-- if we wanted to achieve sorting by position then type, we could use `Comparator.thenComparing()`

seancoulter (commented):
Looking good. How did we arrive at the geometry change sorting mechanism?

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/GeoJsonFeatureToAtlasEntityConverter.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final Long DEFAULT_IDENTIFIER = 0L;
+
+    /**
+     * Attempt to convert a GeoJson {@link Feature} to an {@link AtlasEntity}.
+     *
+     * @param feature
+     *            {@link Feature}
+     * @return an {@link Optional} {@link AtlasEntity}
+     */
+    public static Optional<AtlasEntity> convert(final Feature feature)
```
seancoulter, line 51: Just a thought-- since the flagged object conversion is lossy in terms of relation identifiers and other feature-dependent data, we might want to (if not already) have some way of recovering that or at least merging it with the existing feature when we're applying fix suggestions in MR

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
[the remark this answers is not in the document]
Bentleysb: The effect here is that it is sorted by type and then position. However I had forgotten about `.thenComparing` so I have updated this to use that.

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
```diff
@@ -51,15 +96,303 @@ public CheckFlag deserialize(final JsonElement json, final Type typeOfT,
...
+                // Add the new member to the bean and add the bean to the relation
+                case ADD:
+                    bean.add(member);
+                    ((CompleteRelation) afterView).withMembers(bean,
+                            ((CompleteRelation) afterView).bounds());
+                    break;
+                // Remove the member from the bean and add the bean to the relation
+                case REMOVE:
+                    bean.removeItem(member);
+                    ((CompleteRelation) afterView).withMembers(bean,
```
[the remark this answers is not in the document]
Bentleysb: It could have some side effects to downstream processes if they assume that there aren't pointless relations with no members, but the check flag that gets created will still be valid.

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
```diff
@@ -51,15 +96,303 @@ public CheckFlag deserialize(final JsonElement json, final Type typeOfT,
...
+    private List<Location> convertWktToLocations(final String wkt)
+    {
+        try
+        {
+            return Lists.newArrayList(COORDINATE_ARRAY_CONVERTER.backwardConvert(
+                    new CoordinateArraySequence(WKT_READER.read(wkt).getCoordinates())));
+        }
+        catch (final ParseException parseException)
+        {
+            throw new CoreException("Cannot parse wkt : {}", wkt);
```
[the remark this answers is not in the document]
Bentleysb: Good call.

src/main/java/org/openstreetmap/atlas/checks/utility/OpenStreetMapCheckFlagConverter.java
Bentleysb: yes

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/GeoJsonFeatureToAtlasEntityConverter.java
Bentleysb: Initially this should not be a problem while we are focusing on tag change suggestions, but this will likely be something we need to look at for other types of suggestions.

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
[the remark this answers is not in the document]
Bentleysb: This sorting was arrived at through a bit of a trial and error process, as it is somewhat hard to decipher the exact system that is used to arrive at the change descriptions for this. The descriptors are created using a list diff form the git/github algorithms. I tested a bunch of geometry changes with combination of adds, removes and replaces in different places, and this appeared to work every time to rebuild from the diff.

seancoulter (commented):
@Bentleysb are the CheckFlagDeserializer changes backwards compatible? FlagDB uses the old version
Code looks good as well-- looks like just a few code smells having to do with generics

seancoulter (approved):

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
```diff
@@ -51,15 +96,306 @@ public CheckFlag deserialize(final JsonElement json, final Type typeOfT,
         final JsonObject full = json.getAsJsonObject();
         final JsonObject properties = full.get(PROPERTIES).getAsJsonObject();
         final String checkName = properties.get(GENERATOR).getAsString();
-        final String instruction = properties.get(INSTRUCTIONS).getAsString();
-        final String flagIdentifier = properties.get(IDENTIFIERS) == null
```
sayas01, line 55: Why is the null check removed?

sayas01 (commented):
Awesome work! Nice documentation and tests!

src/main/java/org/openstreetmap/atlas/checks/flag/serializer/CheckFlagDeserializer.java
Bentleysb: This is removed because it is no longer using the unique identifiers for the new flags identifier. The new flag now will have the same identifier as the old, and the unique identifiers can again be calculated from the flagged objects. Essentially, this was a bit of a hack because everything was not being deserialized previously.

sayas01 (approved):

PR #410, opened 2020-11-04
Copy S3 files by default in AWS script

atiannicelli:
### Description:

Add ability for AWS script to copy necessary files from S3 instead of using s3fs.

### Potential Impact:

None to atlas checks.
The AWS script will take longer to start but will most likely run smoother and faster once s3 sync completes.

### Unit Test Approach:

I ran the AWS script against my EC2 system and tested the aws s3 sync functionality manually.

### Test Results:

NA


Bentleysb (approved):

sayas01 (approved):
👍

PR #411, opened 2020-11-06
MapRouletteUploadCommand: support for Tag fixSuggestions

seancoulter:
### Description:

This is an update to MapRouletteUploadCommand (and related files) to send up tag-related fix suggestions. 

The new process is to upload line delimited geojson using the MR upload command is as follows: we read each line in the geojson file and deserialize it using the CheckFlagDeserializer, extract relevant data from the newly created CheckFlag to create Challenges, and use CheckFlag's built in `getMapRouletteTask()` to batch and upload Tasks to MR.

Notable points:
A new parameter for the MR upload command determines whether or not fix suggestions are uploaded. Supply true if you want all fix suggestions to be uploaded, or false if you want none. `-includeFixSuggestions=<true/false>`
TaskDeserializer is no longer used, its sole usage having been in the MR upload command and it being replaced by the CheckFlagDeserializer. I would argue TaskDeserializer is obsolete as it does not consider fix suggestions, while CheckFlagDeserializer does
If there are no issues with this reasoning I can delete TaskDeserializer.

### Potential Impact:

1) A `FlaggedPoint extends FlaggedObject` created by one of its constructors is now associated with a country, IF the supplied constructor data is derived from an Atlas. Another method to create FlaggedPoints is from a raw Location -- such FlaggedPoints are not associated with countries and behave as before
2) OpenStreetMapCheckFlagConverter uses the input CheckFlag's challenge name as its own challenge name if one exists

### Unit Test Approach:

TagChangeOperationTest, and some tweaks to existing tests

### Test Results:

Sample upload for tag update fix suggestion:
image

for tag remove:
image


src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
```diff
@@ -485,6 +485,10 @@ public Task getMapRouletteTask()
                     .map(flaggedRelation -> flaggedRelation.asGeoJsonFeature(this.identifier))
                     .forEach(features::add);
         }
+        if (!this.fixSuggestions.isEmpty())
+        {
+            task.setCooperativeWork(this.fixSuggestions);
+        }
```
Bentleysb, line 491: I think this should be optional. Can we add a parameter to this method to set whether we want to include the fix suggestions? This could also be accompanied by an overloaded version of this method that enables this by default, to keep things backwards compatible. We should then also have a command argument for the upload command to set this.

Bentleysb (commented):
Looks good. Just had one request to make the fix suggestion/cooperative challenge portion optional.

src/main/java/org/openstreetmap/atlas/checks/flag/CheckFlag.java
seancoulter: Yeah 100% agreed. Updates are in 1f1d8c1

Bentleysb (commented):
Looks good. Just one code smell for cognitive complexity that I would be fine with if you just wanted to set it to be ignored.

seancoulter:
@Bentleysb went with ignoring the cognitive complexity issue. Its biggest gripes were with chaining conditional statements using `||` and `&&` and having nested try/catches. I don't see a good reason to refactor based on the latter, and for the conditionals I think it's arguably easier to follow with the way it is now as opposed to breaking everything up into variables

Bentleysb (approved):

sayas01 (approved):
LGTM! 👍

PR #420, opened 2020-11-12
Adding PyAtlas-Checks scaffolding

mattmanley:
### Description:
This PR creates the basic structure for the PyAtlas-Checks repo. Essentially, this code will provide a central place to access tooling relevant to Atlas-Checks via a Python CLI. Some other key things to note:
* [Click](https://click.palletsprojects.com/en/7.x/) will be used for the CLI
* Some tools will be integrated with [Spark](https://spark.apache.org/)

### Potential Impact:
N/A

### Unit Test Approach:
N/A

### Test Results:
N/A


danielduhh:
winwinwinwin

@mattmanley

pyatlas_checks/pyatlas_checks/cli/cli.py
```diff
@@ -0,0 +1,20 @@
...
+def cli(ctx, verbose):
+    ctx.obj = {
+        'verbose': verbose
+    }
+
+
+@cli.command()
+@click.argument('test_string')
+@click.pass_context
+def test_command(ctx, test_string):
```
danielduhh, line 17: Do you have to declare a name here? How would you call this from the cli?

pyatlas_checks/setup.py
```diff
@@ -0,0 +1,21 @@
...
+packs = find_packages(exclude=exclude_dirs)
+
+setup(name='pyatlas_checks',
+      version="1.0",
+      packages=find_packages(),
+      include_package_data=True,
+      install_requires=["Click"],
+      entry_points='''
+            [console_scripts]
+            pyatlas=pyatlas_checks.cli.cli:cli
```
danielduhh, line 19: The cli.py file could live in the package root (instead of in its own module). That way you could avoid the `cli.cli:cli`

danielduhh (commented):
Love this! Just had a question about what it would look like to execute the test_command

pyatlas_checks/pyatlas_checks/cli/cli.py
mattmanley: @danielduhh To call this command, all you have to do is pass:
```
pyatlas test-command
```
Of course, calling the above will throw an error since we didn't pass the required parameter.
Click automatically converts all underscores in method names to dashes.

danielduhh: 💯

Bentleysb:
@mattmanley, perhaps it is just my unfamiliarity with click, but is there currently a way to run the cli?

mattmanley:
@Bentleysb To run the cli, simply execute `pyatlas` via the command line. Let me know if that doesn't make sense

Bentleysb (approved):
Looks good. Some next steps might be adding documentation and setting up a unit test framework.

seancoulter (commented):
Looks good -- I would only recommend that we have some sort of standardized input cleansing. Maybe this is doable with Click API, or we can just build off of BaseCommand's `__init__`. Just to make sure we're reading valid input from the command line

A couple other methods to override from BaseCommand that I can think of: getDescription(), printUsageError()

seancoulter (approved):

PR #421, opened 2020-11-12
InvalidTagsCheck Static Instructions Bug

Bentleysb:
### Description:
This fixes a bug in InvalidTagsCheck that was causing IndexOutOfBoundsExceptions in heavily distributed runs. Each time the check was instantiated it read in the custom instructions and added them to a static List of Strings. This caused the static list to grow each time, but spark had trouble communicating that to each executor causing the IndexOutOfBoundsExceptions.

### Potential Impact:

Only impacts the one check.

### Unit Test Approach:

None

### Test Results:

Tested a large distributed run with the changes and the exceptions were gone and flags for InvalidTagsCheck were up.


sayas01 (approved):
Thanks @Bentleysb !

seancoulter (approved):
🙏

PR #432, opened 2020-11-19
tollValidationCheck

reichg:
### Description:

This check attempts to validate toll tags based on the scenarios listed below. 

### Strategy
1. **Case 1:** Intersects toll feature but missing toll tag. 
** Toll tag should be added for all ways that intersect toll features. 

2. **Case 2:** Inconsistent toll tags 
** Out of place way that is between 2 ways with toll=yes. 
** Toll tag should be added to way in question if modeling surrounding way in question is correct. 

3. **Case 3:** Escapable toll 
** Checks for toll tag modeling issues by recursively checking possible routes and if they require a toll. Surrounding ways' toll tags and nearby toll features, if applicable, need to be investigated for proper modeling.  

### Test Results:

| ISO | Case 1 Sampled | Case 1 FP |  Case 1 FPR | Case 2 Sampled | Case 2 FP | Case 2 FPR | Case 3 Sampled | Case 3 FP | Case 3 FPR |
|--------|----------|-------|---------|----|---|----|----|----|----|
|  ARG  |    104     |    3     |  2.84% |  5  | 0 | 0% | 10 |  0  | 0% |
|  AUS  |    34     |    3     |  8.82% |  1  | 0 | 0% | 1 |  0  | 0% |
|  HUN  |    7     |    0     |  0% |  3  | 0 | 0% | 7 |  0  | 0% |
|  TUR  |    76     |    2     |  2.63% |  19  | 0 | 0% | 10 |  0  | 0% |
|  NOR  |    85     |    2     |  2.35% |  16  | 0 | 0% | 10 |  0  | 0% |


src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+    /**
+     * @param edge
+     *            some edge
+     * @return tag inconsistencies between 3 consecutive edges.
+     */
+    private boolean hasInconsistentTollTag(final Edge edge)
+    {
+        final Set<Edge> inEdges = edge.inEdges().stream()
+                .filter(inEdge -> inEdge.getOsmIdentifier() != edge.getOsmIdentifier()
+                        && inEdge.getIdentifier() > 0 && HighwayTag.isCarNavigableHighway(inEdge))
```
Bentleysb, line 455: nit
```suggestion
                        && inEdge.isMainEdge() && HighwayTag.isCarNavigableHighway(inEdge))
```

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+                .collect(Collectors.toSet());
+        final Set<Edge> outEdges = edge.outEdges().stream()
+                .filter(outEdge -> outEdge.getOsmIdentifier() != edge.getOsmIdentifier()
+                        && outEdge.getIdentifier() > 0 && HighwayTag.isCarNavigableHighway(outEdge))
```
Bentleysb, line 459: nit
```suggestion
                        && outEdge.isMainEdge() && HighwayTag.isCarNavigableHighway(outEdge))
```

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @param edgeInQuestion
```
Bentleysb, line 538: It would be good to give a description of what this case covers here.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+    private boolean isCaseOne(final Edge edgeInQuestion,
+            final Map<String, String> edgeInQuestionTags)
+    {
+        return !this.hasTollYesTag(edgeInQuestionTags)
+                && !this.markedIntersectingTollFeature.contains(edgeInQuestion.getOsmIdentifier())
+                && this.edgeIntersectsTollFeature(edgeInQuestion);
+    }
+
+    /**
+     * @param edgeInQuestion
```
Bentleysb, line 553: It would be good to give a description of what this case covers here.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+            final Edge escapableOutEdge)
+    {
+        return this.hasTollYesTag(edgeInQuestionTags)
+                && !this.edgeIntersectsTollFeature(edgeInQuestion)
+                && !this.hasInconsistentTollTag(escapableOutEdge)
+                && !this.hasInconsistentTollTag(escapableInEdge);
+    }
+
+    /**
+     * @param edgeInQuestion
```
Bentleysb, line 574: It would be good to give a description of what this case covers here.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.atlas.items.Node;
+import org.openstreetmap.atlas.tags.AccessTag;
+import org.openstreetmap.atlas.tags.BarrierTag;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.TollTag;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+
+/**
+ * Auto generated Check template
```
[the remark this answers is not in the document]
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+ *
+ * @author greichenberger
+ */
+public class TollValidationCheck extends BaseCheck<Long>
+{
+    // You can use serialver to regenerate the serial UID.
+    private static final long serialVersionUID = 1L;
```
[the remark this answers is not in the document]
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+        if (TypePredicates.IS_EDGE.test(object) && ((Edge) object).isMainEdge()
+                && HighwayTag.isCarNavigableHighway(object)
+                && !isFlagged(object.getOsmIdentifier()))
+        {
+            final Edge edgeInQuestion = ((Edge) object).getMainEdge();
+            final Map<String, String> keySet = edgeInQuestion.getOsmTags();
+            return !this.isPrivateAccess(keySet)
+                    && edgeInQuestion.highwayTag().isMoreImportantThan(this.minHighwayType);
+        }
+        return false;
```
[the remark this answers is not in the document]
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+     * @param object
+     *            object in question
+     * @return flag
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Edge edgeInQuestion = ((Edge) object).getMainEdge();
+        final Map<String, String> edgeInQuestionTags = edgeInQuestion.getOsmTags();
+        final Set<Long> abtNearbyTollEdges = new HashSet<>();
```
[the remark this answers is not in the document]
reichg: Addressed and fixed; Changed name to alreadyBeenCheckedNeafrbyTollEdges. abt meant "already been touched"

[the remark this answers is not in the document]
reichg: Updated all instances of abt

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,601 @@
...
+        final List<Segment> edge2Segments = edge2.asPolyLine().segments();
+        final Segment edge1EndSeg = edge1Segments.get(edge1Segments.size() - 1);
+        final Segment edge2StartSeg = edge2Segments.get(0);
+        final double aLength = edge1EndSeg.length().asMeters();
+        final double bLength = edge2StartSeg.length().asMeters();
+        final double cLength = new Segment(edge1EndSeg.start(), edge2StartSeg.end()).length()
+                .asMeters();
+        return Math.toDegrees(Math.acos(
+                (pow(aLength, 2) + pow(bLength, 2) - pow(cLength, 2)) / (2 * aLength * bLength)));
+    }
```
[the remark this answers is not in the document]
reichg: Addressed and fixed. One small difference is that this is just the difference between the angles and not the angle that the segments make by connecting so I had to update MIN_ANGLE_DEFAULT --> MAX_ANGLE_DIFF_DEFAULT and minAngleForContiguousWays --> maxAngleDiffForContiguousWays

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: Addressed and fixed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: Addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,609 @@
...
+                someEdge -> someEdge.isMainEdge() && HighwayTag.isCarNavigableHighway(someEdge))
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * @param edge
+     *            edge in question
+     * @return nearby toll feature id on the in edge side of the edge in question (upstream)
+     */
+    private Optional<Long> getNearbyTollFeatureInEdgeSide(final Edge edge,
```
Bentleysb, line 388: Is there any chance that this could end up taking a long time to complete, either because the closest toll feature is a long ways away or because this is a depth first search and it takes a while to get to the nearby toll feature.

docs/checks/tollValidationCheck
```diff
@@ -0,0 +1,41 @@
+# TollValidationCheck
+
+#### Description
+
+The purpose of this check is to identify ways that need to have their toll tags investigated/added/removed.
+
```
Bentleysb, line 6: It would be nice to have a description of the configurables here.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: I can just set a limit on how far the search will go. This is just complimentary information for the editor when they are investigating the toll tag modeling. I will limit to around 15 iterations in the search.

docs/checks/tollValidationCheck
reichg: addressed and fixed!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,624 @@
...
+        final Iterable<Node> edgeNodes = edge.connectedNodes();
+        for (final Area area : intersectingAreas)
+        {
+            final boolean areaContainsPolyline = area.asPolygon().overlaps(edge.asPolyLine());
+            final Map<String, String> areaTags = area.getOsmTags();
+            if (!areaContainsPolyline)
+            {
+                continue;
+            }
+            return this.containsBarrierTag(areaTags) && this.barrierTagContainsToll(areaTags);
```
seancoulter, line 237: Is it desired behavior to short circuit here with possibly returning false and not checking other areas/nodes?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,624 @@
...
+    {
+        final Set<Edge> inEdges = edge.inEdges().stream()
+                .filter(inEdge -> inEdge.getOsmIdentifier() != edge.getOsmIdentifier()
+                        && inEdge.isMainEdge() && HighwayTag.isCarNavigableHighway(inEdge))
+                .collect(Collectors.toSet());
+        final Set<Edge> outEdges = edge.outEdges().stream()
+                .filter(outEdge -> outEdge.getOsmIdentifier() != edge.getOsmIdentifier()
+                        && outEdge.isMainEdge() && HighwayTag.isCarNavigableHighway(outEdge))
+                .collect(Collectors.toSet());
+        if (inEdges.size() == this.minInAndOutEdges && outEdges.size() == this.minInAndOutEdges)
```
seancoulter, line 478: Not a huge deal but we could rename this variable (especially in the config) as it seems to be functioning as an exact constraint, not a min

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,624 @@
...
+            + "that might be helpful are: upstream {3, number, #} and downstream {4, number, #}.";
+    private static final String INCONSISTENT_TOLL_TAGS = "Way {0, number, #} has an inconsistent toll tag with its surrounding ways. Please check for proper toll tag modeling.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(INTERSECTS_TOLL_FEATURE,
+            ESCAPABLE_TOLL, INCONSISTENT_TOLL_TAGS);
+    private static final String HIGHWAY_MINIMUM_DEFAULT = HighwayTag.RESIDENTIAL.toString();
+    private static final Double MAX_ANGLE_DIFF_DEFAULT = 40.0;
+    private static final double MIN_IN_OUT_EDGES_DEFAULT = 1.0;
+    private static final double MAX_ITERATION_FOR_SEARCH_DEFAULT = 15.0;
+    private final Set<Long> markedInconsistentToll = new HashSet<>();
+    private final Set<Long> markedIntersectingTollFeature = new HashSet<>();
```
seancoulter, line 47: I think we can use BaseCheck's `markAsFlagged` here instead of these Sets, just because it's more conventional and the memory should be allocated already

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    /**
+     * @param edge
+     *            edge in question
+     * @param alreadyCheckedObjectIds
+     *            already been touched IDs
+     * @return edge that proves backwards is escapable
+     */
+    private Optional<Edge> edgeProvingBackwardsIsEscapable(final Edge edge,
```
seancoulter, line 259: Could we combine this and `edgeProvingForwardIsEscapable()` into a new method where the params would just be `Edge edge, Set<Edge> connectedEdges, Set<Long> alreadyCheckedObjectIds)`?

seancoulter: Actually I just noticed they call each other... if there's not already a stack overflow problem or anything I don't think my suggestion alone should cause that either

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
seancoulter: Same with my other comment about combing two similar methods into one -- I believe this and its outEdge counterpart can be combined

seancoulter (commented):
Neat check, had some requests to combine similar methods and address a short circuit in the code

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: For this specific instance I am just using the config value since it is 1. That variable is used other places as a min. I can just use a different variable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: "continue" won't short circuit the whole function, it will only move to the next area in the Iterable for loop. Nodes will only be checked if and only if there are no areas intersecting the edge. False will only be returned if there are no areas or nodes intersecting the edge.

reichg: This is a good point. It shouldn't. I will update this. I think I adjusted this to reduce the cognitive complexity.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: I think I originally was wanting to track these cases individually and not with markAsFlagged but now it should be okay to do so. Will update

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: I personally would like to keep them separate just for ease of access and understandability? Same with the nearby toll features request. If this is an issue maybe the PR could be accepted without combining the functions and I can see what I Can work out when I add in the autofix suggestions. What do you think?

Bentleysb (approved):

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
seancoulter: Yeah that sounds good. Usually sonar throws a duplication warning but looks like we're good here since inEdges is replaced by outEdges in a few lines

seancoulter (approved):
Thanks for addressing all the requests!

PR #434, opened 2020-11-20
Fix log4j Configuration

Bentleysb:
### Description:

The transient log4j dependency was not properly picking up the configuration from the log4j.properties file. This adds log4j as a direct dependency to resolve this. The unused config/log4j/log4j.properties file is also removed. The configuration is read from src/main/resources/log4j.properties. 

### Potential Impact:

Downstream projects may need to exclude the new dependency if they get it from another source. 

### Unit Test Approach:

None

### Test Results:
Ran `gradlew run` with the fixed configuration and verified that debug logs were no longer present. Also checked that changing the default output to a file worked as expected.


atiannicelli:
Worked for me! Thanks.

atiannicelli:
May I also suggest the following as the parser is very chatty when running the mapRouletteUploadCommand

```
diff --git a/src/main/resources/log4j.properties b/src/main/resources/log4j.properties
index 1a5ef30..28c8f91 100644
--- a/src/main/resources/log4j.properties
+++ b/src/main/resources/log4j.properties
@@ -9,6 +9,7 @@ log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1
 # Show our logs
 log4j.logger.org.openstreetmap.atlas=INFO
 log4j.logger.org.openstreetmap.atlas.geography.clipping=ERROR
+log4j.logger.org.openstreetmap.atlas.geography.geojson.parser=ERROR

 # Define the file appender
 log4j.appender.FILE=org.apache.log4j.RollingFileAppender
```

seancoulter:
@atiannicelli I will leave this up to @Bentleysb , but an alternative is to directly quiet `GeoJsonParserGsonImpl`'s logger from INFO to TRACE or DEBUG. This is one of the heavier sources of logging pollution

Bentleysb:
@atiannicelli, I agree that whole package is quite intense with its logging, and just a bit convoluted in general. I have added your suggestion.

atiannicelli:
Awesome! thanks!

seancoulter (approved):
🙌

sayas01 (approved):
👍

PR #441, opened 2020-12-08
Open boundary check

### LINKED ISSUE #439: OpenBoundaryCheck
### Description
This check attempts to check for Administrative relations that should be closed polygons but are not. 

##### Check Candidate
 - Must be a Relation
 - Must be of type Boundary
 - Must have admin_level tag

##### True Positives
 - [Yakıntaş](https://www.openstreetmap.org/relation/8284136) neighborhood boundary is not closed

##### False Positives
 - [Seattle](https://www.openstreetmap.org/relation/237385) city is a closed boundary 

**References**
[Osmose Issue](https://wiki.openstreetmap.org/wiki/Osmose/issues#6010)
[Osmose Analyzer](https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_relation_open.py)

reichg:
### Description:

This check attempts to check for relations that should be closed polygons but are not closed.  

### NOTES
* There is a max size of polygons that can bet checked based on how the Atlas' are loaded. Large countries and even some states/provinces will not be considered in this check simply due to boundary size.
* Country slicing issue causing some false positives that currently are unavoidable

### Test Results:
|   ISO | Sampled   | Sampling % | TP     | FP | False Positive Rate |
|-------|------------|---------------|------|-----|--------------------|
| ARG  |        46      |     100%       |   46  |   0   | 0%                       |
| AUS   |        2      |     100%          |   2 |   0  | 0%                         | 
| FIN    |        5       |     100%          |   4 |   1   | 20%                     |
| TUR |        3       |     100%          |   2 |   1   | 33%                        |

### IMPORTANT
***This check pretty much requires an exception for unavoidable country slicing issues. Synthetic tags, members vs known osm members, and is_in:country attributes taken into account in this check.***
* False positives above are due to country slicing issues in Atlas and cannot be avoided. 
* Few countries with empty results (no boundary issues detected): CHL, ECU, PER, VEN

Resolves #439


reichg:
#439

reichg:
This check pretty much requires an exception for unavoidable country slicing issues. Synthetic tags, members vs known osm members, and is_in:country attributes taken into account in this check.

Bentleysb (approved):
The code looks great. The large FP rates are a but of a concern, but it appears the aggregate FP rate from your testing is only 3.6%. So I am happy to include this check, with the hope that we can improve the atlas framework to catch those FPs in the future.

dcastrowa:
Is this a new check? If so does it need to be added to configuration.json

reichg:
> Is this a new check? If so does it need to be added to configuration.json

This is in the config!

seancoulter (approved):
Agreeing with the sentiment above, we can revisit the border issue later as it seems to be a weird corner case in only a 2/8 countries sampled

PR #444, opened 2020-12-09
ConstructionCheck Use Atlas Generation Date

Bentleysb:
### Description:

This alters the ConstructionCheck to use the Atlas generation date instead of the current date for comparing against feature dates. 

Using the current date was causing false positive cases because of the time differential between when the data was cut and when the check was being run. It also made it so that running the check with the same input parameters on different days would result in different flags. This causes issues for software validation. 

If the atlas generation date is not available the check defaults back to using the current date. 

### Potential Impact:

Should make the flags produced by ConstrucitonCheck more accurate and stable.

### Unit Test Approach:

none

### Test Results:

Ran the check on a year old atlas and compared the differences in results for pre and post enhancement. The only differences were flags dropped post enhancement, and sampling those showed that they were features that went out of date in between when the data was cut and when it was run.

### DIFF
--- a/docs/checks/constructionCheck.md
+++ b/docs/checks/constructionCheck.md
@@ -29,14 +29,16 @@ We first validate that the incoming object is:
 ##### Flagging the Object
 There are 3 ways we check if the object should be flagged:
 1. We get one of the valid date tags (see DATE_TAGS in ConstructionCheck.java)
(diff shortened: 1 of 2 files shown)


seancoulter (approved):

brianjor (approved):
Nice catch!

Issue #413, opened 2020-11-10
[New Check]SourceMaxSpeedCheck 


mm-ciub:
**Description**
This check should verify that the `source:maxspeed` tag is following the rules established here: https://wiki.openstreetmap.org/wiki/Key:source:maxspeed
The possible values:
- sign (where the speed limit is defined by a numeric sign.)
- markings (where the speed limit is defined by painted road markings.)
- <country_code>:\<context\> (where the speed limit is defined by a particular context, for example urban/rural/motorway/etc., and no maxspeed is signposted)

**Requirements**
1. Check Candidate 
 - Is an `Edge`|`Point`
 - Has `source:maxspeed` tag

**Use Cases**

*Case 1: Not following country:context rule (USA)*
 - `source:maxspeed=Unless otherwise posted speed limit 25 MPH in Oklahoma City;OKC code § 32-166.`
Should have been US:25
 - https://www.openstreetmap.org/way/858591689

*Case 2: URL as value (UY)*
- `source:maxspeed=http://monolitos.montevideo.gub.uy/resoluci.nsf/WEB/Numero/954-13` (10 774 cases)
- https://www.openstreetmap.org/way/78113805#map=15/-34.8557/-56.3414

*Case 3: incorrect value based on survey (USA)*
- `source:maxspeed=survey 2014-05-17`(257 cases)
- https://www.openstreetmap.org/way/9140656

**Further investigation**
Variations of `zone` as values are accepted as exceptions.

**Supported regions**
There are some exceptions that need to be taken into account: 
- UK uses `maxspeed:type=* ` for the purpose associated with `source:maxspeed`
- Belgium has relevant regions source:maxspeed=BE-VLG:..., source:maxspeed=BE-WAL:... and source:maxspeed=BE-BRU

**What's not supported (optional):**
This section explains what elements are not supported by this check, whether or not these elements should be included/addressed in the future, and if these cases are better handled in another (existing or new) check.

**References**
Official OSM wiki documentation:
https://wiki.openstreetmap.org/wiki/Key:source:maxspeed
Cases found using tag info: https://taginfo.openstreetmap.org/keys/source%3Amaxspeed

Bentleysb:
[...]

PR #445, opened 2020-12-11
New check source maxspeed

mm-ciub:
### Description:

The check verifies that the `source:maxspeed` tag follows the established mapping rules. All details can be found here: https://github.com/osmlab/atlas-checks/issues/413

### Potential Impact:

It is a new check so there should be no impact to other resources.

### Unit Test Approach:

Added checks for both valid and invalid cases and checked all the sub-cases that are verified for the tag value: general case, specific case with context, country iso code.

### Test Results:

The check was run on 4 countries with results in 3. Details in the table below

|  ROU   |    2176         |    2176      |     100%       |   2176 |   0 | 0%   |
|  TUR   |    0         |    -      |     -       |   - |   - | -   |
|  NZL   |    1834         |    1834      |     100%       |   1834 |   0 | 0%   |
|  AUS   |    13446         |    7038      |     52%       |   7038 |   0 | 0%   |


Bentleysb:
Please add this check to the [configuration](https://github.com/osmlab/atlas-checks/blob/dev/config/configuration.json) and [available checks](https://github.com/osmlab/atlas-checks/blob/dev/docs/available_checks.md), and add a docs page for it [here](https://github.com/osmlab/atlas-checks/tree/dev/docs/checks).

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+    private final List<String> exceptedCountries;
+
+    /**
+     * @param configuration
+     *            the JSON configuration for this check
+     */
+    public SourceMaxspeedCheck(final Configuration configuration)
+    {
+        super(configuration);
+        this.exceptedCountries = this.configurationValue(configuration, "excepted_countries",
```
Bentleysb, line 68: This can be done using the built in `countries.denylist` configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private static final long serialVersionUID = -7004341564141771203L;
+    private static final String GENERAL_INSTRUCTION = "The element with id {0,number,#} does not follow the source:maxspeed tagging rules";
+    private static final String WRONG_VALUE_INSTRUCTION = "The value must be 'sign', 'markings' or follow the country_code:context format.";
+    private static final String WRONG_COUNTRY_CODE_INSTRUCTION = "{0} is not a valid country code.";
+    private static final String WRONG_CONTEXT_INSTRUCTION = "{0} is not a valid context for the maxspeed source. (valid examples: urban, 30 etc.)";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(GENERAL_INSTRUCTION,
+            WRONG_VALUE_INSTRUCTION, WRONG_COUNTRY_CODE_INSTRUCTION, WRONG_CONTEXT_INSTRUCTION);
+    private static final String SOURCE_MAXSPEED = "source:maxspeed";
+    private static final List<String> POSSIBLE_VALUES = Arrays.asList("sign", "markings");
```
Bentleysb, line 40: This would be good as a configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+    private static final Pattern COUNTRY_CONTEXT_PATTERN = Pattern.compile("[a-zA-Z]{2}:.+");
+    private static final Set<String> EXPECTED_CONTEXT_VALUES = new HashSet<>(
```
Bentleysb, line 42: This would be good as a configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+            Arrays.asList("urban", "rural", "bicycle_road", "trunk", "motorway", "living_street",
+                    "school", "pedestrian_zone", "urban_motorway", "urban_trunk", "nsl", "express",
+                    "nsl_restricted", "nsl_dual", "nsl_single", "zone"));
+    // UK uses a different tag for this use case
+    private static final List<String> DEFAULT_EXCEPTIONS = Collections.singletonList("UK");
+    // Belgium has these 3 regions that are accepted because they have a different default rural or
+    // urban maxspeed
+    private static final List<String> COUNTRY_EXCEPTIONS = Arrays.asList("BE-VLG", "BE-WAL",
```
Bentleysb, line 50: This would be good as a configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+
+            }
+
+            else if (!POSSIBLE_VALUES.contains(sourceValue) && !sourceValue.contains(ZONE))
+            {
+                instructions.add(this.getLocalizedInstruction(VALUE_INSTRUCTION_INDEX));
+            }
+            if (!instructions.isEmpty())
+            {
+                final CheckFlag flag = this.createFlag(object, this.getLocalizedInstruction(
```
Bentleysb, line 128: It would be great if edges that share an osm id could be grouped into one flag like [here](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/tag/InvalidLanesTagCheck.java#L89). For this to work properly it is also important to mark items as flagged by there osm ids.

Bentleysb (commented):
This looks great, mostly had some comments about some configurables. Also it would be great if there are unit tests for any configurables.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private boolean hasSourceMaxspeed(final AtlasObject object)
+    {
+        return object.getOsmTags().keySet().stream().anyMatch(key -> key.contains(SOURCE_MAXSPEED));
+    }
+
+    private boolean isContextValid(final String context)
+    {
+        final boolean isNumber = context.matches("[0-9]].+");
```
seancoulter, line 150: there's an extra `]` here

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
seancoulter: Also what are your thoughts on embedding a fix suggestion here to convert the tag to `sign` or `markings`? @mm-ciub

seancoulter (commented):
This is looking good. Agree with @Bentleysb suggestions about making some fields configurable

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
mm-ciub: @Bentleysb I will start making the improvements, thank you.
@seancoulter, looking over taginfo I see that most cases have the value 'sign' at 21.30% while 'markings' is at 0.01%.  So perhaps a fix suggestion for sign could be added, but the user should assure that there is a sign present. Otherwise it falls under the zone source requiring the country:context format.

Bentleysb (commented):
Could you please add the new configurables in configuration.json, and a description of each in the checks MD file.

Bentleysb (approved):
Looks good!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SourceMaxspeedCheck.java
seancoulter: Ok thanks for the explanation. sounds like this might involve more research, and then we can open up an Issue if fix suggestions are found to be suitable

seancoulter (approved):

PR #447, opened 2020-12-14
InvalidMultiPolygonRelationCheck Max Shape Point Reduction

Bentleysb:
### Description:

This reduces the default for the max shape points configurable in InvalidMultiPolygonRelationCheck from 2,000,000 to 300,000. This configurable controls which features are processed through the expensive overlap operations in the check. At the previous default the check took over 3.5 hours to process [this relation](https://www.openstreetmap.org/relation/5430065). By reducing the cap, all shards are now once again processed in under an hour each. 

### Potential Impact:
Nothing major. A few less features may be processed through the overlap portion of the check. 

### Unit Test Approach:

None

### Test Results:

Tested about 200 countries. With the configuration value all shards took less than an hour to run for InvalidMultiPolygonRelationCheck. There were no differences in the features flagged between pre and post value change.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -416,7 +416,7 @@
   "InvalidMultiPolygonRelationCheck": {
     "members.one.ignore": true,
     "overlap.points.minimum": 0,
-    "overlap.points.maximum": 2000000,
+    "overlap.points.maximum": 300000,
     "challenge": {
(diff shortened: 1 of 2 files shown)


seancoulter (approved):
So simple, so effective

sayas01 (approved):
Nice fix!

Issue #387, opened 2020-10-19
[Fix Suggestions] Malformed Roundabout Check


ylinzhi:
**Which Atlas check is the fix suggestion related to?** 

Malformed Roundabout Check

Description: https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/malformedRoundaboutCheck.md
Script: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java

**Describe more details of suggested fix(es)**

*Suggested Fix 1*: 

- Type of fixing: Geometry fix. 

- Operation: Modify

- Suggested Action: Correct roundabout direction according to country driving rules

- **Restriction**: Auto fix should be only applied when flagged error is 1) Wrong-way-roundabouts, OR 2)Multi-directional roundabouts

**Is there any code enhancement needed if adding the fix suggestion component?**

No

seancoulter:
Looks good. We should be able to suggest auto fixes for roundabouts with invalid source edges (must be a main edge with an ISO code and junction=roundabout) in addition to the other 2 cases mentioned

vladlemberg:
The implementation of AutoFix turn out not a straightforward task. Proposed error checks occurs in different codebase: ComplexRoundabout (Atlas). MalformedRoundabout received combined error messages:
`WRONG_WAY_INVALIDATION = "This roundabout is going the wrong direction, or has been improperly tagged as a roundabout.";
    protected static final String INCOMPLETE_ROUTE_INVALIDATION = "This roundabout does not form a single, one-way, complete, car navigable route.";`
In order to isolate each particular case, ComplexRoundabout error messaging has to be more granular. Please refer: https://github.com/osmlab/atlas/issues/709. 

@ylinzhi  @seancoulter @Bentleysb @sayas01 @atiannicelli

vladlemberg:
@Bentleysb, I'd like to clarify AutoFix suggestion scenarios based on our discussion in PR.
**Wrong Roundabout Direction** 
Case 1: Simple Roundabout. Closed way. 4 Atlas Edges. [Example](https://www.openstreetmap.org/way/801814369)
Requirements: Create single AutoFix edit comprising 4 Edges

Case 2: Complex Roundabout. Unclosed Way. Different OSM ids. Example: need to find
Requirements: Create multiple AutoFix edits per each OSM id. Am I understand this correctly ? 

[...]

PR #449, opened 2020-12-16
Add AutoFix suggestion to "wrong direction" cases only.

vladlemberg:
### Description:
AutoFix suggestion to "wrong roundabout direction" cases only. Please refer https://github.com/osmlab/atlas-checks/issues/387

### Potential Impact:
Roundabouts with wrong direction will be suggested with AutoFix

### Unit Test Approach:
1. Modify all existent cases for precise error handling
2. Add AutoFix verifier to "wrong direction" test cases

### Test Results:
Passed

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
@@ -209,21 +209,8 @@ public class MalformedRoundaboutCheck extends BaseCheck<Long>
     }
 
     /**
-     * Checks if {@link ComplexRoundabout} is MultiWay Roundabout. MultiWay Roundabout formed of
-     * several unique OSM Ids. Example: https://www.openstreetmap.org/way/349400768
-     * 
-     * @param roundaboutEdges
-     *            the {@link Set} to check
-     * @return true if roundabout is MultiWay
-     */
-    private boolean isMultiWayRoundabout(final Set<Edge> roundaboutEdges)
-    {
-        return roundaboutEdges.stream().map(Edge::getOsmIdentifier).distinct().count() > 1;
-    }
-
-    /**
      * Checks if {@link AtlasObject} contains synthetic boundary Node
-     * 
+     *
      * @param object
      *            the {@link AtlasObject} to check
      * @return true if roundabout contains synthetic boundary Node.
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -99,17 +104,43 @@ public boolean validCheckForObject(final AtlasObject object)
...
+            // AutoFix candidate only for wrong direction case.
+            if (complexRoundabout.getAllInvalidations().size() == 1 && complexRoundabout
+                    .getAllInvalidations().get(0).getReason().equals(WRONG_WAY_INVALIDATION))
+            {
+                // Mark that the Edges have been processed
+                roundaboutEdgeSet.forEach(
+                        roundaboutEdge -> this.markAsFlagged(roundaboutEdge.getIdentifier()));
+
+                return Optional.of(this
+                        .createFlag(new OsmWayWalker((Edge) object).collectEdges(),
```
Bentleysb, line 125: A roundabout can be made up of multiple Ways, so to would be better to use the `roundaboutEdgeSet` here.

vladlemberg: Thanks @Bentleysb ! good suggestion

sayas01 (approved):
LGTM!

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -99,17 +103,43 @@ public boolean validCheckForObject(final AtlasObject object)
...
+                    .getAllInvalidations().get(0).getReason().equals(WRONG_WAY_INSTRUCTIONS))
+            {
+                // Mark that the Edges have been processed
+                roundaboutEdgeSet.forEach(
+                        roundaboutEdge -> this.markAsFlagged(roundaboutEdge.getIdentifier()));
+
+                return Optional.of(this
+                        .createFlag(roundaboutEdgeSet,
+                                this.getLocalizedInstruction(1, object.getOsmIdentifier()))
+                        .addFixSuggestion(FeatureChange.add(
```
Bentleysb, line 126: Sorry, I misread this slightly the first time. As well as flagging each edge in roundaboutEdgeSet, as you have now done, each of those edges also needs its own fix suggestion to reverse the geometry.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
vladlemberg: @Bentleysb, I moved back to new OsmWayWalker. 1) since we are dealing with single entity Roundabout only. 2) its sorted by way section id unlike roundaboutEdgeSet.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java
```diff
@@ -112,28 +114,32 @@ public boolean validCheckForObject(final AtlasObject object)
...
-                                        .from((AtlasEntity) object)).withGeometry(
-                                                ((Edge) object).asPolyLine().reversed()),
-                                object.getAtlas())));
+                        .addFixSuggestion(
+                                FeatureChange.add(
+                                        (AtlasEntity) ((CompleteEntity) CompleteEntity
+                                                .from((AtlasEntity) object)).withGeometry(
+                                                        CommonMethods.buildOriginalOsmWayGeometry(
+                                                                (Edge) object).reversed()),
+                                        object.getAtlas())));
```
Bentleysb, line 136: This is an interesting way to combine the edges into a single fix suggestion. It could result in some nondeterminism due to taking on the atlas id of the starting object, but I think that is ok in this instance.

Bentleysb (approved):
This looks like it should do the job.

seancoulter (approved):
Nicely done, I like the new CommonMethods method!

PR #452, opened 2020-12-18
Concerning angle building check

### LINKED ISSUE #450: ConcerningAngleBuildingCheck
**Description**
This check attempts to flag buildings who have angles between 80 and 89.9 degrees. (Acute)

**Requirements**
* Check Candidate:
  1. Is an area or relation AND is a building or is part of a building.
  2.  Angles between consecutive segments cannot be between 80 - 89.7 degrees or between 90.3 - 100 degrees.

**Use Cases**
Example:
[Building: 791199451](https://www.openstreetmap.org/way/791199451) fits the requirements and has angles that need to be squared.

**Supported regions**
All

reichg:
### Description:

This check attempts to flag building who have angles between 80 and 89.7 degrees or between 90.3 and 100 degrees that need to be squared.

### Test Results:

| ISO     | Sampled |    TP        | FP    | False Positive Rate |
|--------|----------|------------|------|----------------------|
|  ARG   |    300     |    300       |   0    |          0%                |
|  AUS   |    300     |    300       |   0    |          0%                |
|  CHL   |    300     |    300       |   0    |            0%               |
|  COL   |    300     |    299       |   1    |           0.003%           |
|  ECU   |    300     |    299       |   1    |            0.003%          |

Resolves #450

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/areas/ConcerningAngleBuildingCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/areas/ConcerningAngleBuildingCheck.java
@@ -1,0 +1,209 @@
+package org.openstreetmap.atlas.checks.validation.areas;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.Heading;
+import org.openstreetmap.atlas.geography.Polygon;
+import org.openstreetmap.atlas.geography.Segment;
+import org.openstreetmap.atlas.geography.atlas.items.Area;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+import org.openstreetmap.atlas.geography.atlas.items.ItemType;
+import org.openstreetmap.atlas.geography.atlas.items.Relation;
+import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
+import org.openstreetmap.atlas.tags.BuildingPartTag;
+import org.openstreetmap.atlas.tags.BuildingTag;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Auto generated Check template
+ *
+ * @author v-garei
+ */
+public class ConcerningAngleBuildingCheck extends BaseCheck<Long>
+{
+    private static final long serialVersionUID = 8586559001979110697L;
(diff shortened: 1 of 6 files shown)


src/test/java/org/openstreetmap/atlas/checks/validation/areas/ConcerningAngleBuildingCheckTest.java
```diff
@@ -0,0 +1,41 @@
...
+    public ConcerningAngleBuildingCheckTestRule setup = new ConcerningAngleBuildingCheckTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    private final ConcerningAngleBuildingCheck check = new ConcerningAngleBuildingCheck(
+            ConfigurationResolver.inlineConfiguration("{\"ConcerningAngleBuildingCheck\": {"
+                    + "\"angles\": " + "{\"minLowAngleDiff\": 80.0," + "\"maxLowAngleDiff\": 89.9,"
+                    + "\"minHighAngleDiff\": 90.1," + "\"maxHighAngleDiff\": 100.0" + "},"
+                    + "\"angleCounts\": {" + "\"min\": 4.0," + "\"max\": 16.0}}}"));
```
Bentleysb, line 25: It would be nice to have a unit test or two for the angleCounts configurable.

Bentleysb (commented):
This looks great, just had one comment.

src/test/java/org/openstreetmap/atlas/checks/validation/areas/ConcerningAngleBuildingCheckTest.java
reichg: Sure! I will add these!

reichg: Addressed and added!

Bentleysb (approved):
Looks good

src/main/java/org/openstreetmap/atlas/checks/validation/areas/ConcerningAngleBuildingCheck.java
```diff
@@ -0,0 +1,234 @@
...
+     */
+    private double getAngleDiff(final Segment segment1, final Segment segment2)
+    {
+        final Optional<Heading> segmentOneHeading = segment1.heading();
+        final Optional<Heading> segmentTwoHeading = segment2.heading();
+        if (segmentOneHeading.isPresent() && segmentTwoHeading.isPresent())
+        {
+            return segmentOneHeading.get().difference(segmentTwoHeading.get()).asDegrees();
+        }
+        return ANGLE_DEFAULT;
```
seancoulter, line 154: I think we may run into a situation where the vertex of 2 segments contains 2 nodes --1 node belonging to the meeting segments and another stray node which would be a Segment of length 0. So there's actually 3 segments all connected. We could log something so it's clear why we're skipping this kind of polygon

seancoulter (approved):
this looks good, just noticed an edge case that would be good to track

PR #457, opened 2020-12-29
Adding PyAtlas-Checks testing framework

mattmanley:
### Description:
This PR adds a unit testing framework for PyAtlas-Checks using [Tox](https://tox.readthedocs.io/en/latest/). It also adds a very basic log counter as the first command. With this command, you can count the number of lines in a log file. Along with this new command, I've added two unit tests to make sure the command is functioning as expected. I've provided some sample logs for the tests to use. 

### Potential Impact:
N/A

### Unit Test Approach:
I've created a unit test framework and two new tests. To run the tests, pass the `tox` command. 

### Test Results:
Tox test passes with zero failures.


Bentleysb (commented):
Is the log.gz file being used?

mattmanley:
@Bentleysb The log.gz file is not be used presently. We do support that type of file extension so I was initially thinking having tests for that sort of file. Let me know what you think.

Bentleysb:
@mattmanley, git can have a bit of trouble with binary files, they end up taking up extra space because they can't be diffed. So if we are not using the zipped file I am in favor of removing it.

Bentleysb (approved):
LGTM

danielduhh (approved):
👏🏿

Issue #460, opened 2021-01-05
[Bug] Properly handle "One Member Relation" check in MissingRelationTypeCheck & OneMemberRelationCheck.  


vladlemberg:
**Describe the bug**
Incorrect calculation of "One Member Relation" in utility.CommonMethods when artificial Relation Member (Point role) is being added during the Atlas Generation Step.

**To Reproduce**
Following "one member relations" have two members in Atlas.  
https://www.openstreetmap.org/relation/4580159
https://www.openstreetmap.org/relation/9587737
https://www.openstreetmap.org/relation/3641069
https://www.openstreetmap.org/relation/5686162
https://www.openstreetmap.org/relation/5383018
https://www.openstreetmap.org/relation/4492401

Atlas Example:
```
[Relation: id=3641069000000, [Members: 
				{Member: ID = 1957892426000000, Type = NODE, Role = }, 
				{Member: ID = 1957892426000000, Type = POINT, Role = } **Artificial Member**
			], [Tags: [last_edit_user_name => ], [last_edit_changeset => 0], [last_edit_time => 1397098835000], [last_edit_user_id => 0], [iso_country_code => CHN], [last_edit_version => 1]]]
			
[Relation: id=4580159000000, [Members: 
				{Member: ID = 3105460625000000, Type = NODE, Role = }, 
				{Member: ID = 3105460625000000, Type = POINT, Role = } **Artificial Member**
			], [Tags: [last_edit_user_name => ], [last_edit_changeset => 0], [last_edit_time => 1423652780000], [last_edit_user_id => 0], [iso_country_code => CHN], [last_edit_version => 1]]]

```			

**Expected behavior**
1) Count one member relation with size=1.
2) MissingRelationTypeCheck - will not flag such relations
3) OneMemberRelationCheck - will flag such relations

**Environment (please complete the following information):**
 - OS: Any

vladlemberg:
@sayas01, could you please assign this task to me ?

PR #462, opened 2021-01-07
Fix one member relation size

vladlemberg:
### Description:

Please refer to https://github.com/osmlab/atlas-checks/issues/460

### Potential Impact:
MissingRelationTypeCheck - to generate less flags
OneMemberRelationCheck - to generate more flags

### Unit Test Approach:
Create One Member Relation with Artificial Member (same OSM id)

### Test Results:
CommonMethods - passed
MissingRelationTypeCheck - passed
OneMemberRelationCheck - passed

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
@@ -6,7 +6,7 @@ import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
 /**
  * Hold common Methods (should be used in more than one check)
  *
- * @author Vladimir Lemnberg
+ * @author Vladimir Lemberg
  */
 public final class CommonMethods
 {
@@ -24,7 +24,7 @@ public final class CommonMethods
             // De-duplicating either Point or Node with same OSM Id
             if (entity.getType().toString().matches("POINT|NODE"))
             {
-                return String.valueOf("PointNone")
+                return String.valueOf("PointNode")
                         .concat(String.valueOf(entity.getOsmIdentifier()));
             }
             return entity.getType().toString().concat(String.valueOf(entity.getOsmIdentifier()));
--- a/src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
@@ -1,6 +1,5 @@
 package org.openstreetmap.atlas.checks.utility;
 
-import org.openstreetmap.atlas.geography.atlas.items.AtlasEntity;
 import org.openstreetmap.atlas.geography.atlas.items.Relation;
 import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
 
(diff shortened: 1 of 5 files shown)


atiannicelli (approved):

Bentleysb (approved):
Nice catch!

src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
```diff
@@ -19,10 +20,8 @@
      */
     public static long getOSMRelationMemberSize(final Relation relation)
     {
-        return relation
-                .members().stream().map(RelationMember::getEntity).map(entity -> entity.getType()
-                        .toString().concat(String.valueOf(entity.getOsmIdentifier())))
-                .distinct().count();
+        return relation.members().stream().map(RelationMember::getEntity)
```
seancoulter, line 23: Although unlikely, we might run into a case where 2 different feature types with the same OSM ID exist in the same relation. To avoid returning 1 in such cases, we could do something similar to the previous method but every time we see a "node" or "point" we convert it to the same string, then attach the OSM id

seancoulter (commented):
Interesting find -- I just noticed a minor edge case

src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
vladlemberg: @seancoulter, i'm not sure if its possible in original OSM. every element in OSM has unique and permanent ID. do you know any examples ? I ran a query in our local OSM replication and couldn't find any.

seancoulter: Here's a node and way with the same osm id. IDs should be unique by osm feature, but not necessarily across features: 
https://www.openstreetmap.org/way/152653228
https://www.openstreetmap.org/node/152653228

Since these aren't in the same relation there shouldn't be a problem in this particular example, but my comment was just pointing out the small probability that we'd encounter something like this in the same relation.

vladlemberg: thanks for the example! could we simply remove "POINT" type from the counting ? In my understanding all "POINT" type are artificial and added during the atlas generation. something like `.filter(entity -> !entity.getType().toString().equals("POINT"))`

seancoulter: I think actually the Nodes are the ones added later on, and Points are present after raw atlas generation (Points generally being the Atlas analogue to OSM Nodes). Atlas Nodes would be added during way sectioning at specific locations (see https://github.com/osmlab/atlas/tree/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning#detailed-steps step 2) . So, we might not want to remove points from the count as they might not have a corresponding node -- but I think we can get away with removing either a point or node from the count if they both exist and have the same OSM identifier.

vladlemberg: @seancoulter, thanks for the feedback. should be good now.

seancoulter (dismissed):
awesome, thanks for making that update!

src/main/java/org/openstreetmap/atlas/checks/utility/CommonMethods.java
```diff
@@ -19,10 +19,16 @@
...
-        return relation
-                .members().stream().map(RelationMember::getEntity).map(entity -> entity.getType()
-                        .toString().concat(String.valueOf(entity.getOsmIdentifier())))
-                .distinct().count();
+        return relation.members().stream().map(RelationMember::getEntity).map(entity ->
+        {
+            // De-duplicating either Point or Node with same OSM Id
+            if (entity.getType().toString().matches("POINT|NODE"))
+            {
+                return String.valueOf("PointNone")
```
Bentleysb, line 27: Nit: PointNone -> PointNode

Bentleysb (approved):

seancoulter (approved):

Bentleysb (approved):

PR #467, opened 2021-01-12
Fix Configuration Country Permitlist Bug

Bentleysb:
### Description:

This fixes a bug where a check with a `countries.permitlist` (or the shortcut paramter `countries`) would always be enabled in those countries regardless of the global or their local `enabled` value. Thus when UnwalkableWaysCheck was run in SGP, even if it was set with `"enabled": false` it would still run. 

### Potential Impact:

The correct checks will always be run. 

### Unit Test Approach:

None

### Test Results:

Tested setting the global enabled value to false and running checks in SGP and RUS. Before this fix UnwalkableWaysCheck and InvalidSignBoardRelationCheck would continue to respectively run in those countries. Ran again post fix and the runs failed in both countries with a no checks enabled error as expected.


seancoulter (approved):
Classic rogue && operator. Good catch

atiannicelli (approved):
nice

Issue #392, opened 2020-10-19
[Fix Suggestions] Building Road Intersections Check


ylinzhi:
**Which Atlas check is the fix suggestion related to?** 

Building Road Intersections Check
Description: https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/buildingRoadIntersectionCheck.md
Script: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/BuildingRoadIntersectionCheck.java

**Describe more details of suggested fix(es)**

*Suggested Fix 1*: 

- Type of fixing: Geometry fix

- Operation: Delete

- Suggested Action: Remove the building that intersect/overlap with a road

**Is there any code enhancement needed if adding the fix suggestion component?**

No

sayas01:
The check flags buildings that intersect the center lines of roads. I feel like the most appropriate fix is to update the geometry of the building to not intersect/overlap the road. Something like the lines that intersects the building could be updated to be so and so distance from the road. This will involve code change.

PR #472, opened 2021-01-14
AutoFix suggestions for Building Road Intersection check (#392)

ladwlo:
Adding auto-fix suggestions.


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/BuildingRoadIntersectionCheck.java
```diff
@@ -128,6 +131,7 @@ public boolean validCheckForObject(final AtlasObject object)
                 ignoreTags());
         final CheckFlag flag = new CheckFlag(getTaskIdentifier(building));
         flag.addObject(building);
+        flag.addFixSuggestion(FeatureChange.remove(CompleteEntity.shallowFrom(building)));
```
sayas01, line 134: @ladwlo Like we discussed offline, this may not be the best way to resolve buildings intersecting roads. Please see my comment on https://github.com/osmlab/atlas-checks/issues/392. I feel like this suggestion would not be acceptable in most cases. We may need a further investigation on this.

ameliaewatts:
@ladwlo Because this is still using the previous Travis framework, in order to pass, we need to close it and reopen this pr or push a new commit!

ladwlo:
Closing this PR according to the discussion.

Issue #390, opened 2020-10-19
[Fix Suggestions] Road Name Gap Check


ylinzhi:
**Which Atlas check is the fix suggestion related to?** 

Road Name Gap Check
Description: https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/RoadNameGapCheck.md
Script: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java

**Describe more details of suggested fix(es)**

*Suggested Fix 1*: 

- Type of fixing: Tag fix

- Operation: Add or Modify

- Suggested Action: 1) **Add** proper road segment name tag when it's missing. 2) **Modify** proper road segment name tag when it's inconsistent with its connected road segments

**Is there any code enhancement needed if adding the fix suggestion component?**

No

sayas01:
👍

mm-ciub:
Is this still open? I would be interested to work on it.

atiannicelli:
As far as I know it is. The road name gap check does not have auto fix suggestions yet. go for it!

mm-ciub:
I believe this issue could be closed now.

Bentleysb:
complete

PR #475, opened 2021-01-20
Added fix suggestion to RoadNameGapCheck

mm-ciub:
### Description:

Added fix suggestion for missing or wrong value for `name` tag flagged by RoadNameGapCheck. More details in original issue: https://github.com/osmlab/atlas-checks/issues/390

### Potential Impact:

No impact on other libraries.

### Unit Test Approach:

Updated existing unit tests to also verify that fix suggestions are issued as expected.

### Test Results:
For all the cases the check has correctly identified a road name gap the suggestion was valid. 

|  ROU   |    73         |    73      |     100%       |   72 |   1 | 1%   |
|  NZL   |    86         |    43      |     50%       |   43 |   0 | 0%   |


atiannicelli (approved):
Looks good to me. The only thing that makes me uneasy is the case where there is a name currently and it doesn't match the in and out edge names. If we suggest to rename it to the in and out edge name then the original name is lost. Wondering if it would make sense to add the original name into another tag (e.g. "old_name:<originalName>")

Probably in most cases the new name is correct- I've just seen a couple of these types where it looks like the road might have a valid second name.
Screen Shot 2021-01-20 at 9 01 04 AM

mm-ciub:
@atiannicelli I agree, adding the old name to the old_name tag would increase the quality of the fix suggestion. I'll get to it.

atiannicelli (commented):
1) I would like a second opinion as to whether the old_name tag is a good idea. I don't want to complicate things if we don't need to.
2) This change won't work if the old_name tag is already set. I would think that if the old_name tag is already set then we would just fall back to your original suggestion to just replace the name. But like I said above... not sure this is worth the complication.

Bentleysb:
I agree with @atiannicelli that storing the current name in another tag is a good idea. However, I would propose using the alt_name tag, for two reasons. One, I think it is safer to assume that the road has two names then to assume one is disused. Two, there is some convention for having multiple alt_name tags (alt_name_2, etc.), that could be used in case an alt name already exists. 

@mm-ciub, I am curious about your test results. It appears that the check is giving some false positives that are being passed on to the fix suggestions. Could you share the FPs that you found? Ideally we would like to see a FP rate of 1% or less for the fix suggestions. So I would like to see if we might be able to weed these FPs out, or if they are distinct enough that we could improve the rate by sampling more countries.

mm-ciub:
I looked over it again and one was added by mistake to the list of FP. The other two are:
- https://www.openstreetmap.org/edit?editor=id&way=168637804#map=20/46.93662/23.66136; It says that the edge name is empty but it is present, seems to have been added since 12.01.2020; perhaps this is just some data issue, the used shards are from 28.09.2020
- https://www.openstreetmap.org/edit?editor=id&way=659737340#map=20/47.73155/23.79189; The suggested name was not present on the connected edges; Looking in history, there has been an edit that removed the name tag from the connected edges in October 2020. 

I will take the latest shards from a different country and run the check on them to make sure that these were just some older data issues.

mm-ciub:
I noticed a lot of the cases where the name needs to be changed are bridges, so perhaps for this cases we should keep the old value in `bridge:name` tag?

atiannicelli:
That seems to make the most sense after reading this: https://wiki.openstreetmap.org/wiki/Key:bridge
I guess I would say if bridge=yes and name:bridge is not set then store the old name there.
If bridge!=yes or name:bridge is already set then use the alt_name tag to store the old name.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -184,6 +180,29 @@ private boolean findMatchingEdgeNameWithConnectedEdges(final Set<Edge> connected
...
+            final String originalName, final String nameSuggestion)
+    {
+        final CompleteEntity completeEntity = (CompleteEntity) CompleteEntity
+                .from((AtlasEntity) object);
+        completeEntity.withReplacedTag(NameTag.KEY, NameTag.KEY, nameSuggestion);
+        if (BridgeTag.isBridge(object) && object.getTag(BridgeNameTag.KEY).isEmpty())
+        {
+            completeEntity.withAddedTag(BridgeNameTag.KEY, originalName);
+        }
+        else
```
atiannicelli, line 193: I hate to say this, but for completeness should we also check for tunnels? In some parts of the world tunnels are named too...
```suggestion
        else if (TunnelTag.isTunnel(object) && object.getTag(TunnelNameTag.KEY).isEmpty())
        {
            completeEntity.withAddedTag(TunnelNameTag.KEY, originalName);
        }
        else
```

atiannicelli (dismissed):
Looks good to me. I just left one suggestion to add the same logic for tunnels.

mm-ciub:
That was a good point. I believe with this we have covered all major cases.

atiannicelli (dismissed):

ameliaewatts:
@mm-ciub Cane we get some other reviewers for this?

mm-ciub:
@ameliaewatts I'm not sure who else is available right now. @Bentleysb would you have time to take a look at the latest changes?

Bentleysb (approved):
Looks good to me!

sayas01:
@mm-ciub Could you also fix the one code smell? Thank you!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -162,6 +182,33 @@ private boolean findMatchingEdgeNameWithConnectedEdges(final Set<Edge> connected
...
+            completeEntity.withAddedTag(BridgeNameTag.KEY, originalName);
+        }
+        else if (TunnelTag.isTunnel(object) && object.getTag(TunnelNameTag.KEY).isEmpty())
+        {
+            completeEntity.withAddedTag(TunnelNameTag.KEY, originalName);
+        }
+        else
+        {
+            final long altNameCount = object.getOsmTags().keySet().stream()
+                    .filter(key -> key.startsWith(AlternativeNameTag.KEY)).count();
```
seancoulter, line 202: Nice!

src/main/java/org/openstreetmap/atlas/checks/validation/tag/RoadNameGapCheck.java
```diff
@@ -113,16 +123,23 @@ public boolean validCheckForObject(final AtlasObject object)
                 outEdges);
         if (matchingInAndOutEdgeNames.isEmpty())
         {
-            // There is no pair of inedge and out edge with same name.
+            // There is no pair of in edge and out edge with same name.
             return Optional.empty();
         }
-
+        final String nameSuggestion = matchingInAndOutEdgeNames.iterator().next();
```
seancoulter, line 129: Related to [this example](https://github.com/osmlab/atlas-checks/pull/475#pullrequestreview-572247523), I think we should be careful assigning an arbitrary name if there is more than one in `matchingInAndOutEdgeNames`. This can happen if pairs of in/out edges are within 9 degrees of heading. I see there is only 1 FP left so ill leave this up to you, but to fix such cases, we might be able to return an empty optional if `matchingInAndOutEdgeNames` is empty OR `edge` has a valence of > 1 at its start and end nodes

ameliaewatts: Do we want to go ahead and merge this, or should we address this before the release?

seancoulter (approved):

sayas01 (approved):
LGTM!

PR #477, opened 2021-01-21
WaterWayCheck: Add autofix suggestion (geometry)

### LINKED ISSUE #403: [Fix Suggestions] Water Way Check
**Water Way Check**
Description: https://github.com/tsmock/atlas-checks/blob/dev/docs/checks/waterWayChecks.md
Code: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/areas/WaterAreaCheck.java

**Describe more details of suggested fix(es)**

Suggested Fix 1:
  Type of fixing: Way fix
  Operation: Modify
  Suggested Action: reverse the way direction 

**Is there any code enhancement needed if adding the fix suggestion component?**

No

tsmock:
### Description:

Add fix suggestions for wrong way waterways, when elevation data suggests that it is going in the wrong direction.
This fixes #403 .

~This depends upon https://github.com/osmlab/atlas/pull/729 (without this PR, only \~8 points can actually be changed).~

~This depends upon https://github.com/osmlab/atlas-generator/pull/179 and https://github.com/osmlab/atlas/pull/750 for OSC upload functionality.~

### Potential Impact:

N/A

### Unit Test Approach:

All current unit tests now check to ensure that there is or is not a fix suggestion present. When there is a fix suggestion present, the before/after views are compared, and are expected to be different. Due to the change between the before/after views, it is expected that reversing a collection of points will cause the before/after view collections to be equal.

### Test Results:
MapRoulette testing mostly showcased issues with SRTM data accuracy -- this is not an issue, since (by default), the portion of the check with the fix will not run without higher quality data.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,6 +101,7 @@ dependencies
 
     // Support Junit 5 tests
     testImplementation packages.junit.api
+    testImplementation packages.junit.params
     testRuntimeOnly packages.junit.engine
     // Support JUnit 3/4 tests
     testCompileOnly packages.junit.junit4
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -30,6 +30,7 @@ project.ext.packages = [
         vintage: "org.junit.vintage:junit-vintage-engine:${versions.junit}",
         api: "org.junit.jupiter:junit-jupiter-api:${versions.junit}",
         engine: "org.junit.jupiter:junit-jupiter-engine:${versions.junit}",
+        params: "org.junit.jupiter:junit-jupiter-params:${versions.junit}",
     ],
     sqlite: "org.xerial:sqlite-jdbc:${versions.sqlite}",
     log4j: "log4j:log4j:${versions.log4j}"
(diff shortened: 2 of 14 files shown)


tsmock:
24 day ping....

tsmock:
> @tsmock, apologies for the lack of reviews. It was my understanding that there were some changes underway to reduce the false positive rate. If that is complete could you update the test results in the PR description?

I've gone through most of the false positives, and it largely relates to how I got the flags to actually show up. By default, the check does not look for uphill waterways unless the resolution of the elevation data is 1.0m or less. I modified it to 100m, as SRTM data is around 70-90m (depending upon latitude).

All that being said, the fix itself worked fairly well, with some minor issues.
MR Issues:
* Some tasks had a cooperative work section even if there was no cooperative work. This may be from how I added a field to prioritize tasks with cooperative work challenges. This may be unnecessary in the future.
OSC Issues:
* Getting the version number _just_ right. I believe it is currently correct, but the version number _matters_.

Bentleysb (approved):
This looks great! Thanks for all the hard work to make this possible. And thanks for all unit testing here.

rheatreena (approved):

PR #478, opened 2021-01-22
Add externalDataInput switch for Checks and make externalData configurable

seancoulter:
### Description:

This makes the framework's external data input path configurable by adding a new Checks switch `externalDataInput`. This input path is fed into the ExternalDataFetcher (currently only used by WaterwayCheck for elevation data). Previously, external data was fetched from the SparkJob input path.

There is also a new configurable, `externalData.enabled`, which lives in BaseCheck and can be used to specify whether externalData should be used on a Check-by-Check basis. So in the configuration, you can decide if you want a certain Check to use externalData. This switch is largely symbolic, as the Check code itself must decide what to do given the switch value.

### Potential Impact:

With an empty `externalDataInput`, we get the same behavior as before. With a user-supplied `externalDataInput`, we are just fetching external data from the user-supplied location

`externalData.enabled` can now turn off WaterWayCheck's external data fetching if desired

### Unit Test Approach:

Tested some local runs with a custom externalDataInput input path and with true/false values for `externalData.enabled` 

### Test Results:

Satisfactory

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -1247,5 +1247,10 @@
       "difficulty": "NORMAL",
       "defaultPriority": "LOW"
     }
+  },
+  "ElevationUtilities": {
+    "elevation.srtm_extent": 1.0,
+    "elevation.srtm_ext": "hgt.zip",
+    "elevation.path": "elevation"
   }
 }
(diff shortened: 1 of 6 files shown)


sayas01 (approved):
Awesome!

Bentleysb (approved):

sayas01 (approved):

PR #482, opened 2021-01-27
Bumping atlas version

ameliaewatts:
Bumping the atlas version from 6.3.4 to 6.4.0.


seancoulter (dismissed):

seancoulter:
Commit 78f421e includes an update to a unit test in InvalidMultiPolygonRelationCheckTest. Changes were made to RelationOrAreaToMultiPolygonConverter in https://github.com/osmlab/atlas/pull/717/ , which is brought into this PR. It appears that those changes led to a different instruction being assigned to the object in the test case in question. The fact that the outers in this test case are `Line`s is probably one reason why we get a different instruction. Previously we checked for geometric overlap between inner and outer (which is satisfied and resulted in instruction A), and now we check that the outer `containsProperly` the inner, which is a stricter containment criterion that is not satisfied and results in instruction B

Bentleysb (approved):

Issue #494, opened 2021-02-04
[New Check] IntersectionAtDifferentLayersCheck


vladlemberg:
**Description**
This check is to detect and flag a node when it is a non-terminal intersection node between two ways which have different layer tag values. Catching such cases will prevent “jump off the bridge” or “climb up the bridge” scenarios when routing.

**Requirements**
An Atlas Object to be flag must meet the following criteria to be considered:

1. Object must be node that is a member of at least two pedestiran-navigable ways and/or car-navigable ways.
2. Node must not be highway=crossing
3. Node must not be railway=level_crossing
4. Ways involved do not terminate at this node. (This may indicate that the node indicates a valid change from one layer to another)
5. Ways intersecting at this node must not have Area=yes
6. Ways intersecting at this node have different effective layer tags and include this node as a member of both ways.
    * Flag node.
    * Auto Fix Suggestion (Phase2):
        * *If* Node can be deleted without changing geometry of ways: Delete Node
        * *Else If* Node can be removed from one way without changing geometry of that way: remove node from that way.
        * *Else* Delete Node and then create a node on each way that is a short distance from intersection that does not alter geometry much.
        * *Note*: If node remains, then tags must be removed that indicate that this node is a crossing.

**Use Cases**
Example:

*Case 1:
True Positive
https://www.openstreetmap.org/way/595435280
<img width="785" alt="image" src="https://user-images.githubusercontent.com/62575608/106534404-da9df380-64a8-11eb-931c-e0e0a33de061.png">

**Further investigation**
N/A

**Supported regions**
All

**What's not supported (optional):**
N/A

vladlemberg:
@sayas01, please assign this to me.

vladlemberg:
@sayas01 please close this issue.

PR #495, opened 2021-02-04
[New Check] IntersectionAtDifferentLayers

vladlemberg:
### Description:

This check is to detect and flag a node when it is a non-terminal intersection node between two ways which have different layer tag values. Catching such cases will prevent “jump off the bridge” or “climb up the bridge” scenarios when routing.

Please refer to requirements: https://github.com/osmlab/atlas-checks/issues/494
 
### Potential Impact:
N/A

### Unit Test Approach:

Simulate true positive and false positive cases.  

### Test Results:

ISO | Sampled | TP | FP | False Positive Rate
-- | -- | -- | -- | --
USA | 7 | 7 | 18 | 0%
GBR | 8| 7 | 0 | 0%
BRA | 2 | 2 | 0 | 0%
IDN | 2| 2| 0 | 0%

### False Positives:
All false positive cases that was identified during the QA validation belong to indoor mapping: level, indoor, steps tags. All these cases handled by applying indoorMappingFilter:` indoor->*|highway->corridor,steps|level->* `


docs/checks/intersectionAtDifferentLevelsCheck.md
```diff
@@ -0,0 +1,35 @@
...
+* Must not be a Railway Level Crossing [OSM-wiki:railway=crossing](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dcrossing).  
+
+Our second goal is to identify connected Edges. 
+* Must be a Main Edge
+* Must be a Car-navigable or Pedestrian-navigable highway type
+* Must not be an Area.
+* Must match Great Separation highway tag filter. Default is empty in configurations.json
+
+#### Great Separation Filter Example 
+Flag Nodes that connect only matching highway types: "bridge->yes|tunnel->yes|embankment->yes|cutting->yes|ford->yes";
```
atiannicelli, line 32: The default, if not defined in the config files, is: "bridge->yes|tunnel->yes|embankment->yes|cutting->yes|ford->yes"
I think that we either need to specify that here or remove that default from the code so that the default is empty. I would also explain here what an empty filter would do.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectionAtDifferentLevelsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @author Vladimir Lemberg
+ */
+public class IntersectionAtDifferentLevelsCheck extends BaseCheck<Long>
+{
+    // Instructions
+    private static final String INSTRUCTION_FORMAT = "The Node id {0,number,#} connects two Ways at different levels.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Collections
+            .singletonList(INSTRUCTION_FORMAT);
+    private static final String GREAT_SEPARATION_FILTER_DEFAULT = "bridge->yes|tunnel->yes|embankment->yes|cutting->yes|ford->yes";
```
atiannicelli, line 35: I think it best to make the default for this an empty string. Then add this string as a recommended filter to reduce the flags to high priority ways if wanted.

atiannicelli (changes requested):
Looks good overall - just a couple things:
1) It looks like your patch made changes to the config file that you did not intend.
2) I personally think that the Great Separation filter should default to "". Then the user can set it to whatever or remove it from the config file to get no filter. As it is not, if they remove it from the config file then they will get the default filter as defined internally and that seems not intuitive to me.
3) Also hoping to add another instruction line to instruct the user as to instruct the mapper to perform a fix.

config/configuration.json
```diff
@@ -4,7 +4,7 @@
       "org.openstreetmap.atlas.checks.validation"
     ],
     "type": "org.openstreetmap.atlas.checks.base.BaseCheck",
-    "enabled.value.default": true
+    "enabled.value.default": false
```
atiannicelli, line 7: careful. we don't want to switch this.

config/configuration.json
```diff
@@ -392,6 +392,17 @@
       "tags":"building"
     }
   },
+  "IntersectionAtDifferentLevelsCheck":
+  {
+    "enabled": true,
```
atiannicelli, line 397: in final version we should take this line out.

atiannicelli (approved):
Assuming you remove the lines with "enabled" in the configuration file the rest looks great!

atiannicelli:
@vladlemberg as discussed can you please change the name of this check and PR to "IntersectionAtDifferentLayers"? This will indicate that this check is specific to outdoor features that have to do with "layers" and NOT "levels".

atiannicelli (dismissed):

Bentleysb (approved):

Bentleysb (approved):

vladlemberg:
@sayas01, @Bentleysb, I think it's the final version of the check. Please also check the code smell report. Whats the guidance of using **var**? I have no problem to change to `final var node = (Node) object;`

Bentleysb (dismissed):
I've never seen sonar flag that before. I wonder if something has updated with sonar. I think it is fine to leave that with the stricter typing for now.

sayas01:
@vladlemberg @Bentleysb The local variable type inference was introduced in java 10 and  available in sonarqube rules since Available Since Apr 19, 2021. I think I have seen other devs use that in some of the recent codes pushed to this repo (https://github.com/osmlab/atlas-checks/blob/2b1b24008e485e3c0a96ebe3ce2024a3ae288990/src/main/java/org/openstreetmap/atlas/checks/validation/points/LoneNodeCheck.java#L101). Since sonar has it as a code smell, I would prefer using var here.

sayas01 (approved):
LGTM! 👍

Bentleysb (approved):

config/configuration.json
```diff
@@ -392,7 +392,7 @@
       "tags":"building"
     }
   },
-  "IntersectionAtDifferentLayerCheck":
+  "IntersectionAtDifferentLayersCheck":
```
sayas01, line 395: Good catch!

sayas01 (approved):

PR #498, opened 2021-02-05
ShardedIntegrityChecksSparkJob Deduplication Revamp

Bentleysb:
### Description:

This changes the way flag deduplication is done in the ShardedIntegrityChecksSparkJob to make it more distributable. 

Currently deduplicaiton is done using a reduce function to combine UniqueCheckFlagContainers. This results in all the flags for a country ending up on a single executor and thread.

This new system rework the UniqueCheckFlagContainer to contain only a single CheckFlag instead of many, and gives it a custom equals method that only compares the check name and unique identifiers of the flag. This allows for the use of sparks built in distinct method to remove duplicate containers in a distributed manor. By distributing this, and the following output step, each executor can be assigned less memory, as the maximum amount of data that any executor may need to store is drastically lowered. 

In making the deduplication step more distributed the final output step had to become more distributed as well, in order to maintain the distribution. This has resulted in minor changes to the naming of the output files to allow for multiple simultaneous writes to occur. Appended to the end of each time stamp in the flag, geojson, and tippecanoe outputs there is a `P` followed by a partition number. The one other result of the output distribution is that slightly more output files are created. This is due to each spark partition creating outputs independently, so the 2500 flag per file cap is reached less often. 

### Potential Impact:

There is no known impact. However anything that uses the UniqueCheckFlagContainer or the timestamp in the output file names will be impacted.

### Unit Test Approach:

Rewrote the UniqueCheckFlagContainer unit tests to reflect the new function.

### Test Results:

Local test runs were performed on CYP, and all output types were verified to be identical in content to runs with the old deduplication system. 
2 Large scale tests were performed. One with most countries, but excluding high data density countries like GRB, FRA, DEU, USA, and CAN. The other was a full world run. Flags from these runs were verified to be identical to the old system, excluding a few differences in checks with known nondeterministic behavior. The smaller run required only 8gb or memory per executor to run, down from 48gb (200 executors with 8 threads each). The world run required 24gb to run effectively, down from 64gb+ to run at all.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
@@ -2,6 +2,7 @@ package org.openstreetmap.atlas.checks.event;
 
 import java.util.Date;
 
+import org.apache.spark.TaskContext;
 import org.openstreetmap.atlas.checks.distributed.GeoJsonPathFilter;
 import org.openstreetmap.atlas.checks.vectortiles.TippecanoeCheckSettings;
 import org.openstreetmap.atlas.generator.tools.spark.utilities.SparkFileHelper;
@@ -64,7 +65,7 @@ public class CheckFlagTippecanoeProcessor extends FileProcessor<CheckFlagEvent>
     @Override
     protected String getFilename()
     {
-        return String.format("%s-%s%s", new Date().getTime(), getCount(),
+        return String.format("%sP%s-%s%s", new Date().getTime(), TaskContext.getPartitionId(), getCount(),
                 new GeoJsonPathFilter(doesCompressOutput()).getExtension());
     }
 }
--- a/src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
@@ -65,7 +65,7 @@ public class CheckFlagTippecanoeProcessor extends FileProcessor<CheckFlagEvent>
     @Override
     protected String getFilename()
     {
-        return String.format("%sP%s-%s%s", new Date().getTime(), TaskContext.getPartitionId(), getCount(),
-                new GeoJsonPathFilter(doesCompressOutput()).getExtension());
+        return String.format("%sP%s-%s%s", new Date().getTime(), TaskContext.getPartitionId(),
+                getCount(), new GeoJsonPathFilter(doesCompressOutput()).getExtension());
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/checks/event/FileProcessor.java
```diff
@@ -159,8 +160,8 @@ public void setBatchSize(final int batchSize)
      */
     protected String getFilename()
     {
-        return String.format("%s-%s%s", new Date().getTime(), this.getCount(),
-                new LogFilePathFilter(this.compressOutput).getExtension());
+        return String.format("%sP%s-%s%s", new Date().getTime(), TaskContext.getPartitionId(),
```
sayas01, line 163: Is there any chance this may cause issues with MR upload?

Bentleysb: All of the utilities I am aware of only care about the folder structure and the file extension of the flag outputs. The MR upload command should be unaffected by this change.

sayas01 (dismissed):
Great work @Bentleysb! The test results are promising. LGTM! 🚢

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
```diff
@@ -64,7 +65,7 @@ public void process(final org.openstreetmap.atlas.event.ShutdownEvent event)
     @Override
     protected String getFilename()
     {
-        return String.format("%s-%s%s", new Date().getTime(), getCount(),
-                new GeoJsonPathFilter(doesCompressOutput()).getExtension());
+        return String.format("%sP%s-%s%s", new Date().getTime(), TaskContext.getPartitionId(),
```
seancoulter, line 68: what is an example of an output file name with the new partition string?

src/main/java/org/openstreetmap/atlas/checks/utility/UniqueCheckFlagContainer.java
```diff
@@ -1,109 +1,90 @@
...
+        {
+            return true;
+        }
+        if (other == null || getClass() != other.getClass())
+        {
+            return false;
+        }
+        final UniqueCheckFlagContainer that = (UniqueCheckFlagContainer) other;
+        return Objects.equals(this.checkName, that.checkName)
+                && Objects.equals(this.uniqueIdentifiers, that.uniqueIdentifiers);
```
seancoulter, line 62: I think this is a good way to do it, but I have run into odd cases where checkflags do not have any atlas objects associated with them. It would be the check's fault, however, something to note if we run into it again and are wondering where all the flags went

seancoulter (commented):
This is exciting! Thanks @Bentleysb

src/main/java/org/openstreetmap/atlas/checks/event/CheckFlagTippecanoeProcessor.java
Bentleysb: Thanks for checking this. It made me realize I had forgotten to change this for the geojson output. 

flag: 1612311851465P5-52.log.gz
geojson: SelfIntersectingPolylineCheck-1612989473907P2-87.geojson.gz
tippecannoe: 1612311851558P0-52.geojson.gz

seancoulter (approved):

Issue #485, opened 2021-01-28
[New Check] SimilarTagValueCheck


brianjor:
**Description**
This check looks for similar values in a tag that are separated by a ‘;’.  

**Requirements**
An Atlas object must meet the following criteria to be considered:
1. Has a tag with multiple values

**Use Cases**
*Case 1: [FP] Destination names are similar but correct*
 - https://www.openstreetmap.org/way/6140462  
 - ![2021-01-27 10 51 48](https://user-images.githubusercontent.com/22751140/106181518-8b775c00-6152-11eb-89fb-09dbae04dd8d.png)

*Case 2: [FP] Ethnonyms are similar but correct*
 - https://www.openstreetmap.org/way/696057937 
 - ![2021-01-27 10 46 55](https://user-images.githubusercontent.com/22751140/106181726-c7122600-6152-11eb-8239-13c20032095c.png)
 
 *Case 3: [TP] Values are similar with only a space being added as the difference*
 - https://www.openstreetmap.org/way/316757724  
 - ![2021-01-27 10 50 17](https://user-images.githubusercontent.com/22751140/106181817-e741e500-6152-11eb-9622-cbc5b2f35e7a.png)

**Further investigation**
There are a lot of FPs from tiger import data on the name_base tag, due to alternative spellings of locations, eg. Annabelle vs Anabelle. Ethnonyms also create a lot of FPs.

**Supported regions**
All countries

**What's not supported (optional):**
This might only support scripts that use an Alphabet type writing system, such as Latin and Cyrillic. So for any tag values typed out in any other script there might be some issues.

**References**
- https://wiki.openstreetmap.org/wiki/Osmose/issues#3060

Bentleysb:
This looks like it could end up being very heavy on FPs. It might be worth trying to get an estimate on the rate before getting to far into development.

brianjor:
Looking at the results that Osmose has (https://osmose.openstreetmap.fr/en/errors?country=usa_washington&item=3060&level=1%2C2%2C3) I'm thinking of having a list of tags to ignore, and possibly a list of values to ignore (ethnonyms, and what ever else pops up), to reduce FP rates.

brianjor:
From a preliminary run just looking for 100% matching tag values in NZL.
All the FPs are from the tags:
- destination -> https://www.openstreetmap.org/way/520976518
- turn:lanes -> https://www.openstreetmap.org/way/627477493
[...]

PR #500, opened 2021-02-05
[New Check] SimilarTagValueCheck

brianjor:
### Description:

This check looks for similar values in a tag that are separated by a ‘;’.  

### Potential Impact:

No impact

### Unit Test Approach:

Added tests for duplicate tags and similar tags

### Test Results:

|-----|-------------|------------|--------------|----|----| -------------------- |
|AUS|      491        |      491     |    100%        | 489 | 2  |         0.004%          |
|ARG|      24         |        24      |    100%        |  24 | 0   |             0%            |
|NZL|      28         |        28      |    100%        |  28 | 0   |             0%            |
|SWE|      44         |        44      |    100%        | 44 | 0   |             0%            |
|TUR|      79         |        79      |    100%        |  79 | 0   |             0%            |

The two FPs (both on guide_text tag):
https://www.openstreetmap.org/node/8965935
https://www.openstreetmap.org/node/4092708898

Resolves: #485

### DIFF
--- a/docs/checks/similarTagValueCheck.md
+++ b/docs/checks/similarTagValueCheck.md
@@ -6,7 +6,11 @@ The purpose of this check is to identify tags whose values are either duplicates
 enough to warrant someone to look at them.
 
 #### Live Examples
+Similar tag values
+1. The node [5142510561](https://www.openstreetmap.org/way/5142510561) has the similar values: "crayfish" and "Crayfish"
 
+Duplicate tag values
+1. The way [173171120](https://www.openstreetmap.org/way/173171120) has multiple duplicate values in the "source" tag
 
 #### Code Review
 
--- a/docs/checks/similarTagValueCheck.md
+++ b/docs/checks/similarTagValueCheck.md
@@ -5,6 +5,17 @@
 The purpose of this check is to identify tags whose values are either duplicates or similar
 enough to warrant someone to look at them.
 
+Configurables: 
+* "value.length.min": Minimum length an individual value must be to be considered for inspection, value.length >= min. 
+* "similarity.threshold.min": Minimum edit distance between two values to be added to the flag where a value of 0 is 
+   used to include duplicates, value >= min.
+* "similarity.threshold.max": Maximum edit distance between two values to be added to the flag, value <= max.
+* "filter.commonSimilars": values that can commonly be found together validly on a tag that are similar but with no 
+   action needed to be taken.
+* "filter.tags": tags that commonly have values that are duplicates/similars that are valid.
+* "filter.tagsWithSubCategories": tags that contain one or many sub-categories that commonly have valid 
+   duplicate/similar values.
+
 #### Live Examples
 Similar tag values
 1. The node [5142510561](https://www.openstreetmap.org/way/5142510561) has the similar values: "crayfish" and "Crayfish"
--- a/docs/checks/similarTagValueCheck.md
+++ b/docs/checks/similarTagValueCheck.md
@@ -35,3 +35,7 @@ similarity threshold.
 From there we split the gathered pairs between those that are duplicate values, and those that are similar.
 The duplicates are added to the instructions and used to create a fix suggestion.
 The similars are just added to the instructions.
+
+#### Fix Suggestion
+We create fix suggestions only on duplicate values, as similar values are difficult to determine which one (if not both)
+should be kept. The fix, for duplicates, is to remove all but one occurrence of the duplicate value from the tag.
--- a/docs/checks/similarTagValueCheck.md
+++ b/docs/checks/similarTagValueCheck.md
@@ -1,0 +1,33 @@
+# SimilarTagValueCheck
+
+#### Description
+
+The purpose of this check is to identify tags whose values are either duplicates or similar
+enough to warrant someone to look at them.
+
+#### Live Examples
+
+
+#### Code Review
+
+This check evaluates all atlas objects that can hold OSM tags.
+Any duplicate tags are removed in a feature change, while similars are flagged for user review.
+
+#### Validating the object
+The incoming object must:
+* have at least one tag with multiple values (contains a ";")
+
+#### Flagging the object
+We filter out all tags that:
+* are tags that commonly contain valid duplicate/similar values
+* values that are similar to others that commonly occur on the same tag
+* values that either contain: length shorter than the defined min length, a number, non-latin characters
+* the last filtering step we remove any tags that do not contain multiple values
+
+We then take the valid tags and compare each value computing similarity between each, using the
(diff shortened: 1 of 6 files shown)


docs/checks/similarTagValueCheck.md
```diff
@@ -0,0 +1,37 @@
+# SimilarTagValueCheck
+
+#### Description
+
+The purpose of this check is to identify tags whose values are either duplicates or similar
+enough to warrant someone to look at them.
+
```
Bentleysb, line 7: It would be great to have a description of each of the configurables here.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
```diff
@@ -0,0 +1,332 @@
...
+            // Gender
+            Set.of("male", "female"), Set.of("woman", "man"), Set.of("women", "men"),
+            Set.of("male_toilet", "female_toilet"),
+            // Medical
+            Set.of("radiology", "cardiology"),
+            // Sports
+            Set.of("baseball", "basketball"), Set.of("bowls", "boules"), Set.of("padel", "paddel"),
+            // Other
+            Set.of("formal", "informal"), Set.of("hotel", "hostel"), Set.of("hump", "bump"),
+            Set.of("seed", "feed"));
```
Bentleysb, line 68: It would be great to have these as configurables.

docs/checks/similarTagValueCheck.md
```diff
@@ -0,0 +1,37 @@
...
+* values that are similar to others that commonly occur on the same tag
+* values that either contain: length shorter than the defined min length, a number, non-latin characters
+* the last filtering step we remove any tags that do not contain multiple values
+
+We then take the valid tags and compare each value computing similarity between each, using the
+Levenshtein Edit Distance algorithm. We keep value pairs with a similarity that falls within our 
+similarity threshold.
+
+From there we split the gathered pairs between those that are duplicate values, and those that are similar.
+The duplicates are added to the instructions and used to create a fix suggestion.
```
Bentleysb, line 36: Please add a description of the fix suggestion under a `#### Fix Suggestion` section.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * Map a tag entry to an entry where their value is a list of tuples containing the two similar
+     * tags and their edit distance.
+     */
+    private Entry<String, List<Similar>> findSimilars(final Entry<String, String> entry)
+    {
+        final List<String> values = Arrays.asList(entry.getValue().split(SEMICOLON));
+        final List<Similar> similars = new ArrayList<>();
+        for (int i = 0; i < values.size() - 1; i++)
```
Bentleysb, line 232: Please don't use single character variable names.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+import org.openstreetmap.atlas.checks.configuration.ConfigurationResolver;
+import org.openstreetmap.atlas.checks.validation.verifier.ConsumerBasedExpectedCheckVerifier;
+
+/**
+ * Tests for {@link SimilarTagValueCheck}
+ *
+ * @author v-brjor
+ */
+public class SimilarTagValueCheckTest
```
Bentleysb, line 14: It would be great to have a test or two for true negative cases.

Bentleysb (commented):
This looks great. I just had a few little comments. Also it looks like there is one code smell.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
```diff
@@ -0,0 +1,346 @@
...
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // Only want objects with tags that contain multiple values.
+        return object.getOsmTags().values().stream().anyMatch(value -> value.contains(SEMICOLON));
```
seancoulter, line 117: Noting that it might be worth ensuring the semicolon is not at the end of a string, in which case we could use regex here

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
```diff
@@ -0,0 +1,346 @@
...
+     */
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        final Map<String, String> tagsWithMultipleValues = object.getOsmTags().entrySet().stream()
+                .filter(entry -> !this.tagsToIgnore.contains(entry.getKey()))
+                .filter(Predicate.not(this::isTagWithSubCategoriesToIgnore))
+                .map(this::removeCommonFalsePositiveValues)
+                // Only keep tags that still have multiple values
+                .filter(entry -> entry.getValue().contains(SEMICOLON))
```
seancoulter, line 135: could be here too if you prefer, to drive down the # of expensive regex checks

seancoulter (commented):
Cool check, noticed something that could be cleaned up earlier in the code

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
brianjor: Just did a comparison run against AUS data with this change
`return object.getOsmTags().values().stream().anyMatch(value -> value.contains(SEMICOLON) && (value.indexOf(SEMICOLON) != value.length() - 1));`
and the times are ~equal (59s vs 60s, both producing 1030 flags)

src/main/java/org/openstreetmap/atlas/checks/validation/tag/SimilarTagValueCheck.java
brianjor: When a tag gets to that the last filter its value will not have a semicolon as it's last character. While it wasn't something I was considering when I wrote it, any tag with a semicolon as its last character will get it removed in the 'removeCommonFalsePositiveValues'.

Bentleysb (dismissed):

Bentleysb (approved):

seancoulter (approved):
Thanks for testing the regex change, this looks good

PR #505, opened 2021-02-10
Add --force param to aws script to force a rerun of script

atiannicelli:
### Description:

this patch changes the semantics of the aws script a bit. Once this is merged the script will execute the check on the AWS instance once! If rerun then it will not rerun the check unless you add the --force param. This makes it so that if you are disconnected and the script finishes while you are disconnected then it will simply sync the output and finish without re-executing the checks

### Potential Impact:

Since the semantics of the script are changing a bit this could impact automation scripts that use this script. This will make it easier to pick up where the developer left off if disconnected.

### Unit Test Approach:

Ran this a few times to test --force parameter and behavior when script was still running on the remote instance and when the script had finished.

### Test Results:

NA


atiannicelli:
@vladlemberg and @reichg would you mind reviewing this?

reichg (approved):
Seems to work as expected. I ran a couple tests and they worked. Nice! I don't use the sharded spark job. I have found that I can still run larger countries with the normal spark job and not the sharded job. Seems to run a bit faster too.

Bentleysb (approved):

scripts/cloud-check-control/cloudAtlasCheckControl.py
```diff
@@ -216,22 +225,21 @@ def atlasCheck(self):
...
-            )
+        if self.is_process_running("SparkSubmit"):
+            logger.info("About to wait for remote script to complete. If "
+                        "disconnected before the script completes then execute the "
+                        "following command to continue waiting:\n {} --id={}"
+                        .format(" ".join(sys.argv), self.instanceId))
+            # wait for script to complete
+            if self.wait_for_process_to_complete():
+                self.finish(
+                    "Timeout waiting for script to complete. TODO - instructions to reconnect.",
```
seancoulter, line 240: Can --force be used if there was a timeout?

atiannicelli: yes. If there was a timeout and the internal script completed then the user would need to use force to rerun the script... or if force is not used then it would simply let the user know that the script completed and then attempt to sync the data produced to the s3 output.

seancoulter (approved):

PR #512, opened 2021-02-22
OverlappingEdgeCheck: One task per way

### LINKED ISSUE #501: [Feature Request] OverlappingEdgeCheck one task per set of contiguous overlapping segments
**Is your feature request related to a problem? Please describe.**
Each OverlappingEdgeCheck flag indicates a single segment of two ways that are duplicated. It would be better to group all contiguous segments that are duplicates into one flag.

**Describe the solution you'd like**
Currently the instructions for this check when uploaded to Map Roulette state "Duplicate Highway found for way id 829552865. Overlapping way id(s) 829571967." But if multiple segments are flagged for these two ways each task has the same instructions. It would be nice to have one task that lists all the contiguous segments that overlap for the two ways. Having a task for each segment that is overlapped is too noisy. Once we have an auto-fix suggestion for this check that suggests to delete one of the duplicates we will want a single task for the set of segments instead of one task per segment.
Also, Since OverlappingEdgeCheck has replaced duplicateWaysCheck I would propose removing the documentation docs/checks/duplicateWaysCheck.md 

**Describe alternatives you've considered**
NA

**Additional context**
Good Map Roulette example of two segments on the same duplicated ways
https://maproulette.org/challenge/16317/task/80047048
https://maproulette.org/challenge/16317/task/80047049
Same instructions, but different segment - wouldn't it be nice if it was one task with a list of segments that were all duplicates?

brianjor:
### Description:

Resolves #501. For each edge we walk the way and gather all other edges. Looping over those edges we mark them as flagged and group all overlaps into one instruction.

### Potential Impact:

Reduces the number of tasks for this check by grouping up flagged edges that are part of the same OSM way into the same task.

### Unit Test Approach:

None added.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/OverlappingEdgeCheck.java
@@ -6,7 +6,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 import org.openstreetmap.atlas.checks.base.BaseCheck;
 import org.openstreetmap.atlas.checks.flag.CheckFlag;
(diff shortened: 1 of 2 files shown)


brianjor:
I don't understand what is going on with the tests.
The tests testOverlappingHeadAtlas and testOverlappingTailAtlas are having some weird behavior.
The original tests expect 1 flag each but the actual is 2.
When I print the size of the flags, for each test, I get 1. If I change the expected size to 2 the actual then becomes 1.

Bentleysb (approved):

seancoulter (approved):
Nice update

Issue #520, opened 2021-02-26
[Bug] RoadNameGapCheck suggestion when name exists almost never the right thing to do


atiannicelli:
**Describe the bug**
The roadNameGapCheck suggestion when the gap segment has an existing name today suggests to add that name to the alternatives and then use the in and out segments' names. Seems like a good way to go, but in practice it is almost never the right thing to do.

**To Reproduce**
Execute RoadNameGapCheck and then create a MapRoulette challenge from the results. look through the tasks until you find a task where the gap segment has a current name.

**Expected behavior**
I think, considering the feedback from the mappers, that the suggestion for when the gap segment has a name should be removed. For these cases most of the time the name is almost correct and just needs to be edited a bit instead of saving the current name in the alternate name field.

**Screenshots & Stack traces**
NA

**Environment (please complete the following information):**
 - OS: [All]
 - Version [6.1.13]

**Additional context**

mm-ciub:
I believe this was fixed in https://github.com/osmlab/atlas-checks/pull/521

Bentleysb:
Looks complete.

Issue #507, opened 2021-02-11
[Feature Request] RoadNameGapCheck instructions and bridge enhancement


atiannicelli:
**Is your feature request related to a problem? Please describe.**
The instructions for RoadNameGapCheck use the word "edge" when it should use the word "way" or "road". The feature ID of the way in question is also not included in the instructions.

**Describe the solution you'd like**
The current instruction for this check looks like this
```
"Edge name Puente Mecano is different from in edge name and out edge name."
```
This should be changed something like this:
```
"The name, 'Puente Mecano', for way id 157600955 is different from the ways it is connected to, "Cruce Longitudinal (Coquiao) - Pumanzano". Consider setting the name so that it is consistent with the other ways."
```
Also, this check needs to be extended so that if the road name gap happens on a bridge (like in this example) that the instructions are altered to instruct the mapper to assign the name of the bridge using the " bridge:name=*" notation.
```
"The name, 'Puente Mecano', for bridge with id 157600955 is different from the ways it is connected to, 'Cruce Longitudinal (Coquiao) - Pumanzano'. Consider setting the name of this way to 'Cruce Longitudinal (Coquiao) - Pumanzano' so that it is consistent with the other ways setting 'bridge:name=Puente Mecano' to save the original name of the bridge."
```
https://maproulette.org/challenge/14920/task/69444772

**Describe alternatives you've considered**

**Additional context**
This request came from a test team that was working on resolving these issues.

mm-ciub:
I believe this can be closed (https://github.com/osmlab/atlas-checks/pull/521)

Bentleysb:
Looks complete.

PR #521, opened 2021-03-01
#520 roadnamegapcheck update suggestions

mm-ciub:
### Description:

https://github.com/osmlab/atlas-checks/issues/520 
Removed fix suggestion when the flagged road already has a name tag.

### Potential Impact:

Anybody expecting fix suggestions for road gaps with existing name tag will no longer receive them.

### Unit Test Approach:

Updated existing unit tests based on the changes.

### Manual testing
Looked over ROU and NZL flags and all cases only offer suggestions when the name tag is missing. 

ISO | Total Flags | Sampled | Sampling % | TP | FP | False Positive Rate
-- | -- | -- | -- | -- | -- | --
ROU | 73 | 73 | 100% | 73 | 0 | 0%
NZL | 86 | 86 | 100% | 86 | 0 | 0%


atiannicelli:
@mm-ciub while you are editing this check, do you want to address the instruction changes that were requested in this Issue: https://github.com/osmlab/atlas-checks/issues/507 ?

mm-ciub:
> @mm-ciub while you are editing this check, do you want to address the instruction changes that were requested in this Issue: #507 ?

Yes, of course. I just have one question related to this: In this PR I removed all fix suggestions if the flagged road already has a name tag, regardless if it is a bridge or not. In #507 I see an instruction for this case as well. Should I add this bridge case back as a suggestion with the new instruction? Or should I just ignore this instruction?

atiannicelli:
I think that the feedback from our mappers what that the fis suggestions when there was a conflict of names (not just the absence) were never enough to fix the issue completely. I personally would just create an instruction for them to try to clarify what might be a good solution, but do not create a fix suggestion.

atiannicelli (approved):

Bentleysb (approved):

PR #523, opened 2021-03-01
Toll validation check fix

### LINKED ISSUE #412: TollValidationCheck
## ***TollEscapeCheck - Requirements***

Developers: Gabe Reichenberger 
Date: 10/09/2020 

### ***Description*** 

 This check attempts identify 3 cases to validate toll tag modeling.

### ***Requirements***  
***An Atlas Object must meet the following criteria to be considered***

Atlas Object
- Is an Edge 
- Is car navigable 
- Is not private access  
- Minimum tertiary classification

### ***Cases***
***Case 1 - Way intersects toll feature and needs toll=yes tag added to it.***
- Check if way contains node with toll feature related tags or intersects a an area with toll related tags. (examples: barrier=toll_booth and highway=toll_gantry)

***Case 2 - Way is completely surrounded by toll=yes tags and needs toll=yes tag added to it.***
- Check if every connected way has toll=yes tag, flag if true.

***Case 3 - Way does not require a toll because there is a possible route that avoids paying a toll.***
- Checks if there are upstream AND downstream routes that avoid a toll. 
- This is a more complicated case and typically requires investigation on the way in question and edges proving escapable to see if there is a modeling issue.

### ***Live Examples***
***Case 1***
[Way:712334447](https://www.openstreetmap.org/way/712334447) intersects toll feature (Start node is a toll feature)

image

***Case 2***
***Case 3***
 
### ***Supported Countries***
All

reichg:
### Description:

**After re-evaluating the FPs since there was a discrepancy in modeling policy for toll=yes on roads, case 1 results prove to be quality enough for auto-fix. Re-added.**

replacing Sets with Lists, improving logic, and removing auto-fix suggestion from case 1.

### Test Results:

Case 1 - Way intersecting toll feature, add toll=yes tag 
Weighted Average -> 0.0078%
| ISO | Sampled  | TP | FP | False Positive Rate |
|-----|------------|----|----| ------------------- |
|  AUS |    85       | 85 |  0 | 0%   |
|  CAN |    127     | 127 |  0 | 0%   |
|  FRA |    301       | 301 |  0 | 0%   |
|  GBR |    83       | 83 |  0 | 0%   |
|  NOR |    228       | 227 |  1 | 0.4%   |
|  PER |  165        | 165 |  0 | 0%   |
|  USA |    296       | 296 |  0 | 0%   |

Case 2 - Inconsistent toll tag - Way is surrounded by toll=yes tags and needs it added to itself.
Weighted Average -> 0%
| ISO | Sampled  | TP | FP | False Positive Rate |
|-----|------------|----|----| ------------------- |
|  AUS |    23       | 23 |  0 | 0%   |
|  CAN |    17     | 17 |  0 | 0%   |
|  FRA |    26       | 26 |  0 | 0%   |
|  GBR |    0       | 0 |  0 | 0%   |
|  NOR |    63       | 63 |  0 | 0%   |
|  PER |  2        | 2 |  0 | 0%   |
|  USA |    37       | 37 |  0 | 0%   |

Case 3 - Way proven "escapable" - investigation needed to prove toll tag needs to be removed from way or if there are modeling issues nearby way in question.
Weighted Average -> 0%
| ISO | Sampled  | TP | FP | False Positive Rate |
|-----|------------|----|----| ------------------- |
|  AUS |    65       | 65 |  0 | 0%   |
|  CAN |    32     | 32 |  0 | 0%   |
|  FRA |    0       | 0 |  0 | 0%   |
|  GBR |    38      | 38 |  0 | 0%   |
|  NOR |    40      | 40 |  0 | 0%   |
|  PER |  0          | 0 |  0 | 0%   |
|  USA |    79       | 79 |  0 | 0%   |

Resolves #412


config/configuration.json
```diff
@@ -4,7 +4,7 @@
       "org.openstreetmap.atlas.checks.validation"
     ],
     "type": "org.openstreetmap.atlas.checks.base.BaseCheck",
-    "enabled.value.default": true
+    "enabled.value.default": false
```
Bentleysb, line 7: This needs to be reverted.

config/configuration.json
```diff
@@ -1067,6 +1067,7 @@
     "minHighwayType": "tertiary"
   },
   "TollValidationCheck": {
+    "enabled": true,
```
Bentleysb, line 1070: This needs to be reverted.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
```diff
@@ -132,19 +134,46 @@ public boolean validCheckForObject(final AtlasObject object)
...
-            markAsFlagged(object.getOsmIdentifier());
+            markAsFlagged(edgeInQuestion.getOsmIdentifier());
+            if (!isFlagged(escapableInEdge.getOsmIdentifier())
+                    && this.hasInconsistentTollTag(escapableInEdge))
+            {
+                markAsFlagged(escapableInEdge.getOsmIdentifier());
+                return Optional.of(this
+                        .createFlag(new OsmWayWalker(escapableInEdge).collectEdges(),
+                                this.getLocalizedInstruction(2, edgeInQuestion.getOsmIdentifier()))
+                        .addFixSuggestion(FeatureChange.add(
```
Bentleysb, line 150: I thought there wasn't going to be a fix suggestion for this case. Has that changed?

Bentleysb (commented):
I had a question and a couple of comments, otherwise this looks good. Also, are there any unit tests that need to be updated to capture what is being fixed in this PR?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
reichg: This has been re-evaluated. At the beginning of this PR at the top I have the explanation in bold

reichg: this one checks the edges that prove the edge in question is "escapable" but they are being passed through the "inconsistent" logic. That means that if they are found to be "inconsistent" we should flag them as that which means adding a fix suggestion.

config/configuration.json
reichg: addressed

config/configuration.json
reichg: addressed

reichg: > I had a question and a couple of comments, otherwise this looks good. Also, are there any unit tests that need to be updated to capture what is being fixed in this PR?

I could technically add some unit tests for the process where edge proving escapable is being put through the "inconsistent" tag logic prior to continuing case 3 logic.

reichg: at the moment I think the tests are fine but I will see if anymore are necessary

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
Bentleysb: Ok, in that case can we add that to the documentation. Right now it only says cases 1 and 2 will have fix suggestions.

reichg: that is still the case.. The edges proving escapable are put through case 2 before the edge in question can be deemed "escapable" in case 3. This is the discussion we had where the proving edges had to be put through the "inconsistent" test (case 2) first.

Bentleysb (approved):

src/main/java/org/openstreetmap/atlas/checks/validation/tag/TollValidationCheck.java
Bentleysb: Oh, ok. Sorry, I see how this works now. It is just a bit confusing because it is inside the case 3 if block.

seancoulter (approved):
👍

PR #529, opened 2021-03-09
LongNameCheck

Bentleysb:
### Description:

This adds a new LongNameCheck

This check flags features with names longer than a configurable length. 
It's expected use is to detect descriptions in name tags or name tags containing multiple names. 

### Potential Impact:

None just a new check.

### Unit Test Approach:

Unit tests cover true positive and negative cases for a variety of name tags. 

### Test Results:

The check was tested in 8 countries with a total of 57 flags. It has a 0% false positive rate.


docs/checks/longNameCheck.md
```diff
@@ -0,0 +1,33 @@
+# Long Name Check
+
+This check flags features with names longer than a configurable length. 
+It's expected use is to detect descriptions in name tags or name tags containing multiple names.
```
sayas01, line 4: typo: its

config/configuration.json
```diff
@@ -573,6 +573,16 @@
...
+    "name.max": 40,
+    "challenge": {
+      "description": "Tasks containing features with unusually long names.",
+      "blurb": "Unusually long names.",
+      "instruction": "Open your favorite editor validate that the marked names of the features are correct.",
```
sayas01, line 582: Open your favorite editor  **and** validate

src/main/java/org/openstreetmap/atlas/checks/validation/tag/LongNameCheck.java
```diff
@@ -0,0 +1,111 @@
...
+                .filter(entry -> entry.getValue().length() >= this.nameMax).map(Map.Entry::getKey)
+                .collect(Collectors.toSet());
+
+        if (!invalidNameTags.isEmpty())
+        {
+            this.markAsFlagged(object.getOsmIdentifier());
+            final String instruction = this.getLocalizedInstruction(0, object.getOsmIdentifier(),
+                    this.nameMax, String.join(", ", invalidNameTags));
+            return Optional.of(object instanceof Edge
+                    ? this.createFlag(new OsmWayWalker((Edge) object).collectEdges(), instruction)
```
sayas01, line 86: 👍

src/main/java/org/openstreetmap/atlas/checks/validation/tag/LongNameCheck.java
```diff
@@ -0,0 +1,111 @@
...
+     * This function will validate if the supplied atlas object is valid for the check.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        return !this.isFlagged(object.getOsmIdentifier())
```
sayas01, line 61: Since the check looks at all AtlasObjects, if two atlas objects of different types, say relation and way has long names, wouldn't marking the osmIdentifier instead of atlasIdentifier , skip one of the objects?

Bentleysb: Good catch! The atlas identifier has the same issue, but we can create a unique osm id.

sayas01: oh yeah! Thanks!

sayas01 (approved):
LGTM! 💯

seancoulter (approved):
Clean 👍

PR #538, opened 2021-03-28
HighwayMissingNameAndRefTagCheck

### LINKED ISSUE #537: HighwayMissingNameAndRefTagCheck
**Description**
This check detects highways that are missing a name and ref tag. At least one of them is required.

**Requirements**
Check Candidate
 - Is an `Edge`
 - Is MainEdge
 - HighwayTag greater than or equal to 'tertiary'
 - Not a roundabout
 - Not circular

**Use Cases**
* Primary road [way:174686797](https://www.openstreetmap.org/way/174686797) is missing both name and ref tag.

image

* Missing name and ref tag on [way:913183089](https://www.openstreetmap.org/way/913183089)

image

* Missing Name and Ref tag on [Way:29221580](https://www.openstreetmap.org/way/29221580)

image

**Supported regions**
All Countries

reichg:
### Description:

This check detects highways that are missing a name and ref tag. At least one of them is required.
Resolves #537 

### Test Results:

| ISO |Sampled | TP | FP | False Positive Rate |
|AUS |    123    |  122 |  1  |                0.8%   |
|BRA |    120      | 119|  1  |                0.8%   |
|COL|    100      |  100|  0  |                0%   |
|FRA |    119      |  119|  0  |                 0%   |
|NOR |    103     |  102|  1  |                 0.9%   |
|PER |    101      |  101|  0  |                   0%   |
|USA|    105      |  104|  1  |                    0.9%   |


src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,241 @@
...
+import org.openstreetmap.atlas.geography.atlas.walker.OsmWayWalker;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.JunctionTag;
+import org.openstreetmap.atlas.tags.names.NameTag;
+import org.openstreetmap.atlas.tags.names.ReferenceTag;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Angle;
+
+/**
+ * Auto generated Check template
```
Bentleysb, line 26: This needs to be updated.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+
+        // by default we will assume all objects as valid
```
Bentleysb, line 65: This template comment can be removed.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,241 @@
...
+        return object instanceof Edge && ((Edge) object).isMainEdge()
+                && !isFlagged(object.getOsmIdentifier())
+                && ((Edge) object).highwayTag().isMoreImportantThanOrEqualTo(HighwayTag.TERTIARY)
```
Bentleysb, line 68: It would be nice to have this highway tag value be configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Function to determine if edge is a small connector edge which doesn't require a name or ref
+     * tag. Will be ignored.
+     * 
+     * @param edge
+     *            edge in question
+     * @return boolean if edge is a small connector road that doesn't require a name or ref tag.
+     */
+    private boolean isConnectorWayToIgnore(final Edge edge)
```
Bentleysb, line 182: This feels like it may be creating a lot of false negative cases. I would be curious to see some examples of the type of Edge this is supposed to be filtering out.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheckTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Test
+    public void hasNameTagFalsePositive()
+    {
+        this.verifier.actual(this.setup.hasNameTagFalsePositive(), this.check);
+        this.verifier.verifyEmpty();
+    }
+
+    @Test
+    public void missingNameAndRefTag()
```
Bentleysb, line 32: It would be great to have tests for name and ref separately as well.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameOrRefTagCheckTestRule.java
```diff
@@ -0,0 +1,63 @@
...
+                            @TestAtlas.Loc(value = WAY1_NODE2) }, tags = { "highway=motorway",
+                                    "name=highwayname" }),
+                    @TestAtlas.Edge(id = "4000001", coordinates = {
+                            @TestAtlas.Loc(value = WAY1_NODE2),
+                            @TestAtlas.Loc(value = WAY2_NODE2) }, tags = { "highway=motorway",
+                                    "name=highwayname" }),
+                    @TestAtlas.Edge(id = "5000001", coordinates = {
+                            @TestAtlas.Loc(value = WAY2_NODE2),
+                            @TestAtlas.Loc(value = WAY3_NODE2) }, tags = { "highway=motorway",
+                                    "name=highwayname" }) })
```
Bentleysb, line 35: It would be good to have one of these edges have a ref tag instead of a name.

Bentleysb (commented):
I had a few comments and questions. I am most interested in the small connectors, and why they don't need names. I would also be interested to see an example of the false positives encountered.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: A connector road is a road that connects a split road like the following (Thought name/ref was optional on these OR they take higher priority name?):

image

reichg: Some have names and some don't, I will research proper modeling for these and get back.

Bentleysb: I see. This makes a decent bit of sense now. In a case like this I can see not needing a name, though I don't know the proper modeling either. The sort of case I would be worried about would be if the crossing road continued through the other side of the split road. In that case I think it should defiantly have a name or ref.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameOrRefTagCheckTestRule.java
reichg: addressed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: addressed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: addressed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: addressed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,376 @@
...
+    @Override
+    protected Optional<CheckFlag> flag(final AtlasObject object)
+    {
+        markAsFlagged(object.getOsmIdentifier());
+        final Map<String, String> tags = object.getTags();
+        if (this.highwayMissingBothNameAndRefTag(tags)
+                && !this.isConnectorWayToIgnore(((Edge) object).getMainEdge()))
+        {
+            return Optional.of(
+                    createFlag(object, this.getLocalizedInstruction(0, object.getOsmIdentifier())));
```
Bentleysb, line 92: We need to flag all Edges with the same OSM id together. Otherwise it will be nondeterministic which Edge from a Way gets flagged.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
```diff
@@ -0,0 +1,376 @@
...
+     * @param edge
+     *            edge in question
+     * @param inEdges
+     *            edge in question's inEdges
+     * @param inEdgeNames
+     *            Set of inEdges' names
+     * @param inEdgeRefs
+     *            Set of inEdges' refs
+     */
+    private void inEdgeLogic(final Edge edge, final Set<Edge> inEdges,
```
Bentleysb, line 213: Couldn't this be combined with outEdgeLogic?

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: addressed

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayMissingNameAndRefTagCheck.java
reichg: I originally had them together but the cognitive complexity was 23

reichg: yeah I will put this into one function, just need to account for inEdge vs outEdge for the angleDifference between edges.

Bentleysb (approved):

seancoulter (approved):
I think this looks good, thanks @reichg

PR #551, opened 2021-04-28
Add purgeIncompleteTasks parameter to MapRouletteUploadCommand.

atiannicelli:
### Description:

This patch will add a new parameter (defaults to false) that if set to true will instruct the MapRouletteUploadCommand to purge all "Incomplete tasks" from any challenge that is being uploaded before uploading tasks to the challenge. Incomplete tasks are any task with a status of "created" or "skipped". These types of tasks have either not been processed or skipped for some reason. This purge parameter will delete them from the challenge before uploading a new task list so that any stale tasks that might have been fixed since the last update will be removed from the challenge. This prevents mappers from processing old tasks that have already been fixed and will (obviously) reduce the number of "already-fixed" task statuses.

### Potential Impact:

This is a new parameter that if not specified will act exactly the way things used to. If the user wants to purge old tasks then the parameter can be specified when calling the MapRouletteUploadComand. This could potentially extend the amount of time it takes to upload a challenge to Map Roulette because the delete command can take minutes if there are a lot of tasks in the challenge. 

### Unit Test Approach:

I tested this code by uploading a challenge using maproulette upload command. Then setting one task to each of the different statuses. I then ran the upload command again with the purgeIncompleteTasks parameter set and verified that the single "skipped" task and all the "created" tasks were deleted, and then verified that the upload command uploaded a new set of tasks into the challenge. 
I also tested that I could specify the parameter when uploading to a brand new (non existent) challenge.
I also tested that not specifying the parameter did NOT delete old tasks and continue to work as usual.

### Test Results:

All good.

### DIFF
--- a/scripts/cloud-check-control/cloudAtlasCheckControl.py
+++ b/scripts/cloud-check-control/cloudAtlasCheckControl.py
@@ -304,6 +304,7 @@ class CloudAtlasChecksControl:
                     + f" -countries='{c}'"
                     + f" -checks='{check}'"
                     +  " -includeFixSuggestions=true"
+                    +  " -purgeIncompleteTasks=true"
                     + f" > {self.atlasCheckLog}"
                 )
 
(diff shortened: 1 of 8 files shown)


atiannicelli:
Working on "code smells" and "bug"

Bentleysb (approved):
Looks good. Thanks for this great addition.

sayas01 (approved):
Nice addition!

PR #560, opened 2021-05-24
DuplicateLocationInPolyLineCheck: Fix #553, do not flag terminating loops

### LINKED ISSUE #553: [Bug] DuplicateLocationInPolyLineCheck should not flag terminating loops.
**Describe the bug**
The current check flags roads that loop around and end on themselves. Parking lots that loop around and come back to the same entry or turning loops that are mapped with a road that goes around the circle and meets back up with itself should not be flagged. 

**To Reproduce**
Run the DuplicateLocationInPolyLineCheck check on the USA and you well find that it flags many turning loops or parking lots or service roads that simply go around a structure (like in the example below).

**Expected behavior**
The expected behavior would be to NOT flag service roads that end on one of the nodes included in the way. I think you could make a case that if there are no consecutively duplicated nodes in the way then we should ignore the first and last node of the way for detecting when the way crosses itself. A way that loops around and then continues through itself (created an "x" with a loop should still be flagged as an issue.

**Screenshots & Stack traces**
https://www.openstreetmap.org/way/361312322

**Environment (please complete the following information):**
 - OS: All
 - Version 6.1.19

**Additional context**
I found this in a challenge that was created from the check for USA.

tsmock:
* Stop flagging closed loops (Edges and LineItems)
* Continue flagging closed loops (Areas -- the last location in an area line
  _should not_ be the same as the first location in the area line)

### Description:

This fixes #553 by not flagging terminating loops (Edges and LineItems). It continues to flag closed Area loops.

### Potential Impact:

N/A

### Unit Test Approach:

Added unit tests check that the last node does not flag the object when the last node already exists on the line.

### Test Results:
Country | Flags | Sampled | False Positives
-|-|-|-
GBR|138|100%|0

Note: The atlas used for the test samples was from sometime between 2020-12-11 and 2020-12-13 -- some that I initially thought were FP turned out to have been fixed already.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicateLocationInPolyLineCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/DuplicateLocationInPolyLineCheck.java
@@ -66,8 +66,10 @@ public class DuplicateLocationInPolyLineCheck extends BaseCheck<Long>
                             || object instanceof Area || firstHit > 1))
             {
                 this.markAsFlagged(object.getOsmIdentifier());
-                return Optional.of(this.createFlag(object, this.getLocalizedInstruction(0,
-                        currentLocation.toString(), object.getOsmIdentifier())));
+                return Optional.of(this.createFlag(object,
+                        this.getLocalizedInstruction(0, currentLocation.toString(),
+                                object.getOsmIdentifier()),
+                        Collections.singletonList(currentLocation)));
             }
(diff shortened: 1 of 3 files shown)


atiannicelli (approved):
Looks great!

atiannicelli (approved):

Bentleysb (approved):

sayas01 (approved):
LGTM!

PR #565, opened 2021-06-03
MapRoulette Challenge New API Key

Bentleysb:
### Description:

This updates the Challenge object used in the MapRouletteUploadCommand to reflect the change here: https://github.com/maproulette/maproulette2/pull/894

Also bumps a log from debug to error, to save a couple hours of debugging next time. 

### Potential Impact:

None

### Unit Test Approach:

None

### Test Results:
Got a 400 error before the change when creating new challenges. After the change we are back to 200s.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteConnection.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteConnection.java
@@ -124,7 +124,7 @@ public class MapRouletteConnection implements TaskLoader, Serializable
                     logger.debug(logSuccessMessage, responseID);
                     return responseID;
                 default:
-                    logger.debug("{} - {}", createUpdate.getStatusCode(),
+                    logger.error("{} - {}", createUpdate.getStatusCode(),
                             createUpdate.getRequestBodyAsString());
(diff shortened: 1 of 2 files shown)


Bentleysb:
Sonar is unhappy because it thinks the added field is unused, when it is used in converting the class to json, and the logging statement isn't covered by a unit test. I would suggest that both cases are in fact valid, and sonar should be ignored.

atiannicelli (approved):
Thank you!!! Looks good to me!

src/main/java/org/openstreetmap/atlas/checks/maproulette/MapRouletteConnection.java
```diff
@@ -124,7 +124,7 @@ public long create(final String getURI, final String postURI, final String putUR
                     logger.debug(logSuccessMessage, responseID);
                     return responseID;
                 default:
-                    logger.debug("{} - {}", createUpdate.getStatusCode(),
+                    logger.error("{} - {}", createUpdate.getStatusCode(),
```
sayas01, line 127: nice!

sayas01 (approved):
@Bentleysb Thank you for the quick update!

Issue #557, opened 2021-05-20
[Bug] ShortSegmentCheck flags are mostly False Positives 


atiannicelli:
**Describe the bug**
The ShortSegmentCheck seems to want to flag edge/way segments that are too short, but in practice I see that OSM ways that don't have any short segments are being flagged.  

**To Reproduce**
Execute the ShortSegmentCheck

**Expected behavior**
I think that this check wants to find way segments that are very short for no reason. It may be that this check needs to be converted to consider way segments instead of edge segments??? Anyway, I don't think that most of the flagged issues that I have seen are actually short segments. I have added a couple MapRoulette examples of ShortSegmentCheck flags that are not short segments. The issue MAY be that some of these are near boundaries. Is it possible that the edge segment is being segmented internally in atlas when a boundary is involved? 

Also, the instructions should be altered to say "way" instead of "edge" and use the OSMID instead of the edge ID. 
"This edge 41895328001000 is short (length < 1 m) and it is connected to node 432327349000000 that has less than 3 connections."
should become:
"A segment of way 41895328 is short (length < 1 m) and it is connected to node 432327349 that has less than 3 connections."

**Screenshots & Stack traces**
https://maproulette.org/admin/project/43639/challenge/18416/task/95410708/inspect
<img width="1750" alt="Screen Shot 2021-05-20 at 4 34 02 PM" src="https://user-images.githubusercontent.com/5851115/119045260-34a7c380-b989-11eb-9a0a-eef1b155062d.png">
`{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [
                    34.933394,
                    31.343115
                ]
            },
            "properties": {}
        },
        {
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [
                    34.933394,
                    31.343115
                ]
            },
            "properties": {
                "flag:check": "Israel - ShortSegmentCheck",
                "flag:generator": "Atlas Checks",
                "synthetic_highlight_point": "yes"
            }
        },
        {
            "type": "Feature",
            "geometry": {
                "type": "LineString",
                "coordinates": [
                    [
                        34.933394,
                        31.343115
                    ],
                    [
                        34.9333939,
                        31.3431133
                    ]
                ]
            },
            "properties": {
                "ref": "60",
                "oneway": "yes",
                "highway": "trunk",
                "itemType": "Edge",
                "maxspeed": "90",
[...]

PR #568, opened 2021-06-08
Fix bug #557 to exclude atlas internal segments

xiaohongtang2:
### Description:

1. Exclude atlas internal synthetic segments
2. Change instructions to use OSMId, and include startNode, endNode, like below:
      "instructions": "1. This segment from startNode 268179597 to endNode 7392913442 on way 24673494 is short (length < 1 m) and node 268179597 has less than 3 connections."
3. Add a new unit test for internal synthetic segment.
  
I have run the fix for ISR,ISL,MLT,LUX,CPY,PER,NZL,MAR, the synthetic segments/nodes are removed successfully

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ShortSegmentCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ShortSegmentCheck.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.tuple.Triple;
 import org.openstreetmap.atlas.checks.base.BaseCheck;
 import org.openstreetmap.atlas.checks.flag.CheckFlag;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
@@ -13,6 +14,7 @@ import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Node;
 import org.openstreetmap.atlas.tags.BarrierTag;
 import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.SyntheticBoundaryNodeTag;
 import org.openstreetmap.atlas.tags.annotations.validation.Validators;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
 import org.openstreetmap.atlas.utilities.scalars.Distance;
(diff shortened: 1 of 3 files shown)


atiannicelli (approved):
LGTM! Thanks

Bentleysb:
@xiaohongtang2, please correct me if I'm wrong, but I don't think this resolves https://github.com/osmlab/atlas-checks/issues/557. It does make the instructions clearer, but the segments being flagged are still Edges, not Ways. I think we want to reform Ways from there Edges and check the length of those. There is a handy class in atlas for collecting the Edges that make up a Way: https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/walker/OsmWayWalker.java

Bentleysb (commented):
@xiaohongtang2, please correct me if I'm wrong, but I don't think this resolves #557. It does make the instructions clearer, but the segments being flagged are still Edges, not Ways. I think we want to reform Ways from there Edges and check the length of those. There is a handy class in atlas for collecting the Edges that make up a Way: https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/walker/OsmWayWalker.java

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/ShortSegmentCheck.java
```diff
@@ -111,15 +116,17 @@ public boolean validCheckForObject(final AtlasObject object)
     protected Optional<CheckFlag> flag(final AtlasObject object)
     {
         final Edge edge = (Edge) object;
-        final Optional<Node> lowValenceNodes = getConnectedNodesWithValenceLessThan(edge,
-                this.minimumValence);
-        if (lowValenceNodes.isPresent() && !this.isGateLike((Edge) object))
+        final Optional<Triple<Node, Node, Node>> lowValenceNodes = getConnectedNodesWithValenceLessThan(
+                edge, this.minimumValence);
+        if (lowValenceNodes.isPresent() && !this.isGateLike((Edge) object)
+                && !this.isEdgeWithSyntheticBoundaryNode(edge))
```
atiannicelli, line 122: @Bentleysb I think this line is the crux of this patch. This will cause the check to ignore edges that connect to a boundary... correct?

Bentleysb: Oh, I reread the issue and it seems I thought there was more to the issue than I though. Yes, you are correct, and this makes sense now.

Bentleysb (approved):
Sorry, misread the issue. This looks great!

sayas01 (approved):
LGTM!

Issue #561, opened 2021-05-28
lineCrossingWaterBodyCheck leading to bad edits


andrewharvey:
I'm not sure if this is the correct place to report but the check at https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/lineCrossingWaterBodyCheck.md is being used by Facebook mappers to update in Australia, however I'm finding a number of these changes not correct. So I'll post them here?

1. https://osmcha.org/changesets/105459547 the path and the water object were never intersecting to begin with

andrewharvey:
2. https://osmcha.org/changesets/105460554 it was changed to the water and path don't connect, but in this case the paths go to the waters edge so should connect

atiannicelli:
1. I'm not sure why the mapper added a node to that way indicated in the changeset (566314737). The issues as flagged in task https://maproulette.org/task/94126725 was trying to flag the fact that the last node of way 48731943 ends on the edge of the water body 531128566. The check looks for ways that terminate at the water's edge that don't have a tag on that intersecting node to indicate what the connection to the water body is. (e.g. slipway, ferry, etc)
2. This changeset was trying to fix https://maproulette.org/task/94128928 which indicates that both 332749248 and 332750251 end at the waters edge without a tag on that node to indicate what the connection is (e.g. slipway, dock, etc)

The specific filter we use to indicate that a node in a way is allowed to connect to a water body edge is:
"ferry->!no&ferry->*|ford->!no&ford->*|leisure->slipway|amenity->ferry_terminal"
Generally, if a path ends at a body of water then we have been ending the way just before it gets to the water body feature.

It sounds like you would rather allow the ways to connect to the water's edge without a special tag to indicate what the connection is?

I have sent a note directly to the mapper that was working on the challenge to ask her to pause edits for this challenge until we resolve this issue - we by no means want to make edits that cause issues or that are inaccurate.

andrewharvey:
> I'm not sure why the mapper added a node to that way indicated in the changeset (566314737). The issues as flagged in task https://maproulette.org/task/94126725 was trying to flag the fact that the last node of way 48731943 ends on the edge of the water body 531128566. The check looks for ways that terminate at the water's edge that don't have a tag on that intersecting node to indicate what the connection to the water body is. (e.g. slipway, ferry, etc)

Oh that makes more sense, I should have checked the actual MapRoulette task.

So these are both the same issue then, a way terminating at the waters edge.

[...]

PR #571, opened 2021-06-09
lineCrossingWaterBodyCheck adding handling for footway, bridleway, steps, corridor, path

vladlemberg:
### Description:

Please refer to: https://github.com/osmlab/atlas-checks/issues/561

mapping rules:
1. ways that end on the edge of a water body with the following tags should not be flagged as issues:
highway=(footway|bridleway|steps|corridor|path)
2. ways that cross into a water body (and end there) with the highway=steps should also not be flagged.

### Potential Impact:

better handling for footway, bridleway, steps, corridor, path 

### Unit Test Approach:

created unit tests to validate logic according to mapping rules 

### Test Results:

passed

### DIFF
--- a/docs/checks/lineCrossingWaterBodyCheck.md
+++ b/docs/checks/lineCrossingWaterBodyCheck.md
@@ -8,6 +8,8 @@ An Edge should be flagged if it has the appropriate highway tag (specified in "h
 
 A building should be flagged if the "buildings.flag" is set to true, and the building is not a public_transport->station,aerialway=station, is on the same level as the waterbody, and intersects a valid part of the waterbody.
 
+Footpaths highway=(footway|bridleway|steps|corridor|path) should not be flagged if they are ending at waterbody. Special handling of highway=steps, which may go inside a waterbody. 
+
 #### Live Examples
 
 1. Street [id:32845241](https://www.openstreetmap.org/way/32845241) and building [id:753774494](https://www.openstreetmap.org/way/753774494) intersect waterbody feature [id:10886166](https://www.openstreetmap.org/relation/10886166) invalidly. Both the street and waterbody share a location (node) at the point of intersection, but this node is not tagged with something from "nodes.intersecting.non_offending".
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
@@ -86,6 +86,7 @@ public class LineCrossingWaterBodyCheck extends BaseCheck<Long>
             + "embankment->yes|location->underwater,underground|power->line,minor_line|"
             + "man_made->pier,breakwater,embankment,groyne,dyke,pipeline|route->ferry|highway->proposed,construction|ice_road->yes|winter_road->yes|snowmobile->yes|ski->yes|"
             + "ford->!no&ford->*";
+    private static final String DEFAULT_LINE_ITEMS_OFFENDING_CROSSING_ONLY = "highway->footway,bridleway,steps,corridor,path";
     private static final String DEFAULT_HIGHWAY_MINIMUM = "TOLL_GANTRY";
     private static final List<String> DEFAULT_HIGHWAYS_EXCLUDE = Collections.emptyList();
     private static final String BUILDING_TAGS_DO_NOT_FLAG = "waterway->dam|public_transport->station,platform|aerialway->station";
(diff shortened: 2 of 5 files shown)


atiannicelli (approved):
Looks great to me. Vlad, would you mind updating the documentation to reflect the new parameter?
atlas-checks/docs/checks/lineCrossingWaterBodyCheck.md

ameliaewatts:
@vladlemberg May we get one more reviewer on this and have you addressed @atiannicelli's comment?

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -412,12 +425,18 @@ private boolean canCrossWaterBody(final AtlasItem crossingItem,
                 final Set<Tuple<PolyLine, Set<Location>>> interactionsPerWaterbodyComponent = this
                         .getInteractionsPerWaterbodyComponent(waterbody, object,
                                 lineItem.asPolyLine(), flag);
-                final AtomicBoolean validCrossWaterbody = new AtomicBoolean(true);
+                this.validCrossWaterbody.getAndSet(true);
```
Bentleysb, line 428: What was the reason for this change?

vladlemberg: I moved validCrossWaterbody to global variable because I need an access from handlingFootPath method.

Bentleysb: I see, ok.

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -412,12 +425,18 @@ private boolean canCrossWaterBody(final AtlasItem crossingItem,
...
                 // Just need to see if the intersection points are allowed in OSM; if not flag them
                 if (interactionsPerWaterbodyComponent.isEmpty())
                 {
                     return false;
                 }
+                // refer to: https://github.com/osmlab/atlas-checks/issues/561
+                else if (interactionsPerWaterbodyComponent.size() == 1
+                        && this.lineItemsOffendingCrossingOnly.test(lineItem))
+                {
+                    this.handlingFootPath(interactionsPerWaterbodyComponent, lineItem);
```
Bentleysb, line 438: This seems redundant at the moment, as handlingFootPath only ever sets validCrossWaterbody to be true or does nothing and I believe validCrossWaterbody is always already true at this point.

vladlemberg: ones this condition met the logic is not following to else statement where the most of the cases processing. the confusion because of inverted return: !this.validCrossWaterbody.get();

Bentleysb: I get that the the final return inverted, and that if this else if is met then the else is bypassed. My issue is that when this else if is true then handlingFootPath is called, and handlingFootPath can only set validCrossWaterbody to true or do nothing, but validCrossWaterbody will always be true at the point handlingFootPath is called from this else if body.

Bentleysb (approved):

Bentleysb (approved):

atiannicelli (approved):

docs/checks/lineCrossingWaterBodyCheck.md
```diff
@@ -8,6 +8,8 @@ An Edge should be flagged if it has the appropriate highway tag (specified in "h
 
 A building should be flagged if the "buildings.flag" is set to true, and the building is not a public_transport->station,aerialway=station, is on the same level as the waterbody, and intersects a valid part of the waterbody.
 
+Footpaths highway=(footway|bridleway|steps|corridor|path) should not be flagged if they are ending at waterbody. Special handling of highway=steps, which may go inside a waterbody.
```
sayas01, line 11: Thanks for updating the doc! 🙌

src/main/java/org/openstreetmap/atlas/checks/validation/intersections/LineCrossingWaterBodyCheck.java
```diff
@@ -507,6 +521,41 @@ private boolean canCrossWaterBody(final AtlasItem crossingItem,
...
+    {
+        final AtomicBoolean allowedFootPath = new AtomicBoolean(false);
+        // refer to: https://github.com/osmlab/atlas-checks/issues/561
+        if (interactionsPerWaterbodyComponent.size() == 1
+                && this.lineItemsOffendingCrossingOnly.test(lineItem))
+        {
+            interactionsPerWaterbodyComponent.forEach(tuple ->
+            {
+                final Set<Location> interactionLocations = tuple.getSecond();
+                // crossing only ones or touching.
```
sayas01, line 545: typo: crossing only ~ones~ -> once

sayas01 (commented):
Looks good!

sayas01 (approved):

PR #581, opened 2021-07-14
[new] StreetNameCheck

nelkapelmenka:
### Description:

This check flags all the objects (Nodes, Ways and Relations) with "ss" or "ß" found in objects. In addition, it flags 
Relations with deprecated tags

Austria: Finding invalid values ("strasse") in the name tag or street_name tag
Switzerland: Finding invalid values ("straße") in the name tag or street_name tag
Germany Case 1: Finding invalid values ("strasse") in the name tag or street_name tag
Germany Case 2: Finding deprecated tag: Relation with a type tag "associatedStreet"
Liechtenstein: Finding invalid values ("straße") in the name tag or street_name tag

### Test Results:

| ISO  | Sampled  | TP | FP | False Positive Rate |
|  AUT |    103|    99 |   4 | 4%   |
|  CHE|    98|    96|   2| 2%   |
|  DEU Case 1|    105|    102|   3| 3%   |
|  DEU Case 2|    1|    1|   0 | 0%   |
|  LIE|    40|    40 |   0 | 0%   |


src/main/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameCheck.java
```diff
@@ -0,0 +1,339 @@
...
+    private static final String CONTAINS_VALUE_INSTRUCTION = "The object contains flagged tags: {0}";
+    private static final String CONTAINS_DEPRECATED_VALUE_INSTRUCTION = "The type tag {0} is deprecated.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays
+            .asList(CONTAINS_VALUE_INSTRUCTION, CONTAINS_DEPRECATED_VALUE_INSTRUCTION);
+
+    private final List<String> allCountriesIsoConfig;
+    private final List<List<String>> allValuesToFlagConfig;
+    private final List<List<String>> allValuesNotFlagConfig;
+    private final List<List<String>> allDeprecatedValuesToFlagConfig;
+    private final List<List<String>> allItemsToFlagConfig;
```
Bentleysb, line 94: This never seems to be functionally used, even though these values are supposed to be the main reason for flagging objects.

docs/checks/streetNameCheck.md
```diff
@@ -0,0 +1,46 @@
+# StreetNameCheck
+
+#### Description
+This check flags all the objects (Nodes, Ways and Relations) with "ss" or "ß" found in objects. In addition, it flags 
+Relations with deprecated tags as seen on  [OSMOSE](https://osmose.openstreetmap.fr/en/errors?item=9009).
```
Bentleysb, line 5: I understand that this is all part of the same osmose check, but it seems a bit odd to flag deprecated tags in the same check as flagging name/address_street tag issues.

Bentleysb (changes requested):
I don't understand the main issue this check is trying to solve. To me it appears to just be flagging legitimate spellings of the word road in german. If that is what it is doing, then why is this something we would want to flag?

nelkapelmenka:
@Bentleysb I have made all the small changes mentioned above. Additionally, I removed anything that has to-do with the deprecated tags from the code, the tests and the docs. I also added another config variable (correctTags) to use in the instructions message to clarify which items needs to be changed into what. Thank you!

Bentleysb (dismissed):
LGTM. Thanks for all the back and forth with me on this one @nelkapelmenka.

src/test/java/org/openstreetmap/atlas/checks/validation/tag/StreetNameCheckTest.java
```diff
@@ -0,0 +1,191 @@
...
+ */
+public class StreetNameCheckTest
+{
+    @Rule
+    public StreetNameCheckTestRule setup = new StreetNameCheckTestRule();
+
+    @Rule
+    public ConsumerBasedExpectedCheckVerifier verifier = new ConsumerBasedExpectedCheckVerifier();
+
+    private final StreetNameCheck check = new StreetNameCheck(
```
sayas01, line 21: Love the detailed unit tests. Could you also add a unit test to test the config values using "ConfigurationResolver.inlineConfiguration" ?

nelkapelmenka: Thank you for looking over this. Would you like me to add just a couple of them, or try to get most possible scenarios?

sayas01: I think just one would do.

nelkapelmenka: All done! Thank you

sayas01 (approved):
LGTM!

Bentleysb (approved):

Issue #577, opened 2021-06-23
[Feature Request] LevelCrossingOnRailwayCheck refinement request


atiannicelli:
**Is your feature request related to a problem? Please describe.**
Many of the issues that this check is flagging are intersections between a road and disused rail. I believe that the best tag to add to these nodes is disused:railway=level_crossing
Also, there is probably no real need to flag level crossing on service ways or lower.

**Describe the solution you'd like**
Add a refinement to the LevelCrossingOnRailwayCheck to recognize that the railway is disused and if so then recommend to add `disused:railway=level_crossing` tag to that intersection. I think this can be a auto suggestions case.
Also, we should add a parameter to this check to be able to control what road types we care about. I think that service roads and lower can probably be ignored as they seem to happen in parking lots and undefined spaces that don't really need a crossing node.

**Describe alternatives you've considered**
The way it is makes it tedious to fix these issues...

**Additional context**
see this disused rail to see all the intersections that could be fixed with this suggestion.
https://www.openstreetmap.org/way/771448302

atiannicelli:
@Bentleysb @sayas01 - if this sounds good to you then I will work on this one.

vladlemberg:
@sayas01, please assign this to me

vladlemberg:
Hi @sayas01, please close this issue. Its merged.

PR #583, opened 2021-07-15
LevelCrossingOnRailwayCheck refinement

vladlemberg:
### Description:

Please refer to: https://github.com/osmlab/atlas-checks/issues/577

### Potential Impact:
1) assigning "disuse" to a Node if railway is disused.
2) adding minimum highway type for configuration.json

### Unit Test Approach:

created unit test for following scenarios:
- minimum highway type
- railway=disuse case 

### Test Results:
passed

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -560,6 +560,7 @@
     }
   },
   "LevelCrossingOnRailwayCheck": {
+    "minimum.highway.type": "residential",
     "layer.default": 0,
     "railway.filter": "railway->rail,tram,disused,preserved,miniature,light_rail,subway,narrow_gauge",
     "challenge": {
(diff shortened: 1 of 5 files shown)


vladlemberg:
@sayas01, @Bentleysb, could you please check the sonar bug ? this is strange... do I need to refactor the code to satisfy sonar ?

Bentleysb:
@vladlemberg, as this bug is covered in a different way by the API you can suppress the sonar check for that method. See an example of this here: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/InvalidPiersCheck.java#L129

Bentleysb (approved):
The overall flow of how the railwayTag fields get used is a bit odd, but it works. It might be nice to have some comments just to point out how those are getting used.

vladlemberg:
> The overall flow of how the railwayTag fields get used is a bit odd, but it works. It might be nice to have some comments just to point out how those are getting used.

@Bentleysb, I believe there is no usage of disused tag. The purpose of this change is assigning correct tag with Autofix

sayas01 (approved):
LGTM!

PR #584, opened 2021-07-16
Add Duplicate Relation Check

xiaohongtang2:
### Description:

Add Duplicate Relation Check, this check flags multiple members Relations in OSM that have the same OSM tags and the same members with same roles.
As discussed,  this check excludes one member relations and leave that for the OneMemberRelationCheck for avoiding one member relations overlap. 

### Test Results:
| ISO | Sampled | TP | FP | False Positive Rate |
|  ARG |   99 |    97     |   2 |  2%   |
|  AUS |   100 |    100     |   0 |  0%   |
|  ISR |   56 |    56     |   0 |  0%   |
|  MLT |   7 |    7     |   0 |  0%   |
|  NZL |   49 |    49     |   0 |  0%   |
|  SWE |   61 |    61     |   0 |  0%   |
|  TUR |   51 |    45     |   6 |  11%   |

### Note: 
The False Positive flagged relations in ARG and TUR are actually caused by those relations atlas objects generated from atlas having different members from the Relations in OSM, I filed [atlas issue#763 ](https://github.com/osmlab/atlas/issues/763) for it.

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -268,6 +268,14 @@
       "difficulty": "EASY"
     }
   },
+  "DuplicateRelationCheck": {
+    "challenge": {
+      "description": "Tasks contain Relations where duplicate relations are found.",
+      "blurb": "Duplicate Relations",
+      "instruction": "Open your favorite editor and  remove one or more of the duplicate relations until only one remains.",
+      "difficulty": "EASY"
+    }
+  },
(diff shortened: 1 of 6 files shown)


Bentleysb (approved):

sayas01 (approved):
LGTM!

PR #588, opened 2021-07-21
[new] Highway access check

nelkapelmenka:
### Description:

The check flags the objects that contain the proper access and highway tags.

### Test Results:

| ISO |  Sampled  | TP | FP | False Positive Rate |
|  CUB   |     113      |   113 |   0 | 0%   |
|  NPL   |     113      |   107 |   6 | 5.5%   |
|  FJI   |     8      |   8 |   0 | 0%   |
|  NZL   |     112      |   112 |   0 | 0%   |
|  JOR   |     120     |   120 |   0 | 0%   |
|  PER   |     111      |   111 |   0 | 0%   |
|  LIE   |    112      |   110 |   2 | 1.8%   |


config/configuration.json
```diff
@@ -1073,6 +1081,20 @@
       "tags":"highway"
     }
   },
+  "StreetNameCheck": {
```
Bentleysb, line 1084: Looks like your street name check commits ended up in this PR as well. Please clean that up so that this PR only contains changes for the highway access check.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayAccessCheck.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @author v-naydinyan
+ */
+public class HighwayAccessCheck extends BaseCheck<Long>
+{
+
+    private static final long serialVersionUID = -5533238262833368666L;
+    private static final List<String> AccessTagsToFlag = Arrays.asList("yes", "permissive");
+    private static final List<String> MotorwayHighwayTags = Arrays.asList("motorway", "trunk");
+    private static final List<String> FootwayHighwayTags = Arrays.asList("footway", "bridleway",
```
Bentleysb, line 29: It would be great to have these all be configurable.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayAccessCheck.java
```diff
@@ -0,0 +1,114 @@
...
+     * This function will validate if the supplied atlas object is valid for the check.
+     *
+     * @param object
+     *            the atlas object supplied by the Atlas-Checks framework for evaluation
+     * @return {@code true} if this object should be checked
+     */
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
+        // by default we will assume all objects as valid
```
Bentleysb, line 61: This default comment is no longer true.

Bentleysb (commented):
Out of curiosity, what are some examples of the false positives that were found?

nelkapelmenka:
Hi @Bentleysb and @atiannicelli, thank you for your comments. I have made the modifications mentioned above.

atiannicelli (approved):
LGTM

src/main/java/org/openstreetmap/atlas/checks/validation/tag/HighwayAccessCheck.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private static final long serialVersionUID = -5533238262833368666L;
+    private static final List<String> ACCESS_TAGS_TO_FLAG_DEFAULT = Arrays.asList("yes",
+            "permissive");
+    private static final List<String> HIGHWAY_TAGS_TO_FLAG_DEFAULT = Arrays.asList("motorway",
+            "trunk", "footway", "bridleway", "steps", "path", "cycleway", "pedestrian", "track",
+            "bus_guideway", "busway", "raceway");
+
+    private static final List<String> FALLBACK_INSTRUCTIONS = Arrays.asList(
+            "The access tag value is probably too generic for this way. A tag value of \"yes\" or \"permissive\" allows access to all types of traffic. If special access is granted or restricted on this way then please specify it or remove the access tag. See https://wiki.openstreetmap.org/wiki/Key:access?uselang=en#Transport_mode_restrictions for more information.");
```
Bentleysb, line 34: Should we be checking if there are already special restrictions paired with the access tag? I would guess we don't wan to flag features with something like access=yes and motor_car=no.

nelkapelmenka: This is what I found on OSM key access, "To avoid ambiguity, you may therefore want to avoid general tags access=yes and access=permissive, and use more specific transport modes where appropriate. For example, to distinguish a footway with open access from one with private access, use tags like foot=yes instead of access=yes." 
The assumption was that for each mode there shouldn't be an access=yes, unless it means that there is full access on the road. Hence, when we add the highway tag, we know that it is something specific. This is the logic that was used in the backend code, here is the link: https://github.com/osm-fr/osmose-backend/blob/c6c0e8a4598495c3b21c308081dbda92be0d1717/analysers/analyser_osmosis_highway_motorway.py

Bentleysb: Right, but just below that bit in the wiki it also has the following as a valid example:
`access=yes, motor_vehicle=no means that all transport modes except motor vehicles can use the element`

nelkapelmenka: I think the idea here is that we are combining both the access tag and the highway tag as described in this part, "In theory, adding access=yes to highway=footway could be read as changing default restrictions (which usually are foot=yes and vehicle=no for highway=footway) to yes, highway=footway + access=yes means "road, which is open for all pedestrians and vehicles". In practice, this combination is often used by mappers to modify (rather than enlarge) default values: for example, access=permissive with highway=steps is very unlikely to be traversable by a truck, whatever the tags may say." So, we are not just looking at access tag = yes but specifically a combination of that and the highway tag. Here is an Osmose example flagged for this check, where the tags are: [access:yes, highway:cycleway, motor_vehicle:no] https://www.openstreetmap.org/way/705630999 
I will ask around to see if this makes sense.

Bentleysb: Ah, right. That makes sense now. I did not fully think about the impact of the highway tag filter that is being used. Thanks for the clarification!

Bentleysb (dismissed):

rheatreena (approved):

rheatreena:
Please resolve the conflicts.

nelkapelmenka:
All done with conflicts, thank you

Bentleysb (approved):

sayas01 (approved):
LGTM!

PR #594, opened 2021-08-02
ConnectivityCheck - extracted highway values

mm-ciub:
### Description:

The check only flagged navigable roads. This update extracts highway values to be checked in a configurable in order to
expand the type of flagged edges when required. Default remains navigable roads.

### Potential Impact:

This could increase/decrease the quantity of flags based on configuration. No other changes have been made.

### Unit Test Approach:

Added 2 unit tests, one does not output flags because the highway configured does not match and one which matches and outputs a flag. 

### Test Results:

No country tests have been run since there was no change to logic.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheck.java
@@ -55,6 +55,7 @@ public class ConnectivityCheck extends BaseCheck<Long>
     private static final long serialVersionUID = -380675222726130708L;
     private final TaggableFilter denylistedHighwaysTaggableFilter;
     private final Distance threshold;
+    private final List<String> checkedHighwayValues;
 
     public ConnectivityCheck(final Configuration configuration)
     {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheck.java
```diff
@@ -64,6 +65,8 @@ public ConnectivityCheck(final Configuration configuration)
         this.denylistedHighwaysTaggableFilter = TaggableFilter
                 .forDefinition(this.configurationValue(configuration, "denylisted.highway.filter",
                         DEFAULT_DENYLISTED_HIGHWAYS_TAG_FILTER));
+        this.checkedHighwayValues = this.configurationValue(configuration, "checked.highway.values",
```
Bentleysb, line 68: Could you add a note in the md file for this check describing this configurable?

sayas01 (commented):
LGTM! Will approve after resolving Bentley's comment.

Bentleysb (approved):

PR #599, opened 2021-08-10
added custom names for 5 checks

jo-taylo:
added custom names for 5 checks


Bentleysb:
@jo-taylo, custom names should not be added to the default configuration.

PR #606, opened 2021-08-12
Change Calls to isMultiPolygon to isGeometric

Bentleysb:
### Description:
https://github.com/osmlab/atlas/pull/754 changed the logic of Relation.isMultiPolygon, and moved the previous logic to Relation.isGeometric. This introduced a few false positive flags across a couple of checks, as well as some changes in flag geometries.
The changes in this PR alter the effected checks and classes by changing to use isGeometric. 3 checks are effected, OneMemberRelationCheck, LineCrossingWaterBodyCheck, and InvalidAccessTagCheck. FlaggedRelation is also changed to allow all geometric relations to once again be output with multipolygon geometry, instead of just bounding boxes. 

### Potential Impact:
None

### Unit Test Approach:

Passes existing tests. 

### Test Results:
Ran in over 200 countries and calculated diffs. 

5 flags for InvalidAccessTagCheck were dropped, and all were in found to be intersecting military boundary relations. This is what is expected from this check.
Dropped flag features:
https://www.openstreetmap.org/way/202404790
https://www.openstreetmap.org/way/5205062
https://www.openstreetmap.org/way/721778383
https://www.openstreetmap.org/way/721778382
https://www.openstreetmap.org/way/910938149

3 flags for LineCrossingWaterBodyCheck were dropped. All were false positives, and where dropped due to being part of boundary relations.
Dropped flag features:
https://www.openstreetmap.org/way/112077775
https://www.openstreetmap.org/way/647501394
https://www.openstreetmap.org/way/192168453

The OneMemberRelationCheck changes just effected the instruction used.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedRelation.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/flag/FlaggedRelation.java
@@ -198,7 +198,7 @@ public class FlaggedRelation extends FlaggedObject
      */
     private MultiPolygon relationGeometry(final Relation relation)
     {
-        if (relation.isMultiPolygon())
+        if (relation.isGeometric())
         {
             try
             {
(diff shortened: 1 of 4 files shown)


sayas01 (approved):
LGTM!

Issue #504, opened 2021-02-09
[Bug] ConnectivityCheck flags parallel way as connectable


atiannicelli:
**Describe the bug**
The ConnectivityCheck flags the connection points between parallel ways that connect to parallel ways that contain bridge=yes tags

**To Reproduce**
Execute the Connectivity Check on any country with dual highways that have bridges.

**Expected behavior**
Nodes that are the connecting nodes between two ways should not be considered for connectivity check. 
Here is a Map Roulette task specifying that the node at the end of a bridge should be connected to the endpoint of the parallel bridge.
https://maproulette.org/challenge/14988/task/69492326
Here is the RapiD view of it: https://mapwith.ai/rapid#background=Bing&disable_features=boundaries&id=w314513554&map=22.89/-7.67993/109.05203
These types of nodes (ones that connect a bridge to a way) should not be flagged to be connected to the parallel ways.
Probably the connectivity check needs an additional requirement that:
if node is (NOT the terminating node of a way) and (NOT a connecting node between two ways) then
the angle of the one of the ways that include that node to a possible connecting way must be greater than 30 to consider it as a flagged node. 
In other words, if two ways are parallel (or close to parallel) then they should not be flagged.

**Screenshots & Stack traces**
![Screen Shot 2021-02-09 at 4 00 26 PM](https://user-images.githubusercontent.com/5851115/107427940-efedd000-6aef-11eb-8816-f9d4d18014f7.png)
![Screen Shot 2021-02-09 at 4 00 49 PM](https://user-images.githubusercontent.com/5851115/107427974-fe3bec00-6aef-11eb-8283-56595709363e.png)
Here is another parallel street that just gets close to the other road...
![Screen Shot 2021-02-09 at 4 03 16 PM](https://user-images.githubusercontent.com/5851115/107428233-570b8480-6af0-11eb-869c-29550cf51e2f.png)

**Environment (please complete the following information):**
 - OS: [All]
 - Version [6.1.12]

**Additional context**

PR #608, opened 2021-08-13
ConnectivityCheck flags parallel way as connectable

vladlemberg:
### Description:

Please refer to https://github.com/osmlab/atlas-checks/issues/504

ConnectivityCheck scope is reduced by removing duplicated logic of crossing edges to EdgeCrossingEdge check.  

### Potential Impact:

Nodes that too close but connected to two+ edges won't be flagged. 

### Unit Test Approach:

Corresponded unit tests: invalidDisconnectedNodesCrossingTest, invalidDisconnectedEdgeCrossingTest  from ConnectivityCheck moved to EdgeCrossingEdge. 

### Test Results:
passed

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheckTest.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheckTest.java
@@ -56,6 +56,14 @@ public class ConnectivityCheckTest
     }
 
     @Test
+    public void invalidConnectedNodesTestNavigableDeadEnd()
+    {
+        this.verifier.actual(this.setup.invalidConnectedNodesAtlasNavigableDeadEnd(),
+                new ConnectivityCheck(ConfigurationResolver.emptyConfiguration()));
+        this.verifier.globallyVerify(flags -> Assert.assertEquals(1, flags.size()));
+    }
+
+    @Test
     public void invalidDisconnectedEdgesTest()
     {
         this.verifier.actual(this.setup.invalidDisconnectedEdgesAtlas(),
--- a/src/test/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheckTest.java
+++ b/src/test/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheckTest.java
@@ -55,13 +55,12 @@ public class ConnectivityCheckTest
         this.verifier.globallyVerify(flags -> Assert.assertEquals(1, flags.size()));
     }
 
-    @Test
-    public void invalidDisconnectedEdgeCrossingTest()
-    {
-        this.verifier.actual(this.setup.invalidDisconnectedEdgeCrossingAtlas(),
-                new ConnectivityCheck(ConfigurationResolver.emptyConfiguration()));
-        this.verifier.globallyVerify(flags -> Assert.assertEquals(1, flags.size()));
-    }
+    /*
+     * @Test public void invalidDisconnectedEdgeCrossingTest() {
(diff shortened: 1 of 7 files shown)


src/test/java/org/openstreetmap/atlas/checks/validation/points/ConnectivityCheckTest.java
```diff
@@ -55,13 +55,12 @@ public void invalidConnectedNodesTest()
...
     }
 
-    @Test
-    public void invalidDisconnectedEdgeCrossingTest()
-    {
-        this.verifier.actual(this.setup.invalidDisconnectedEdgeCrossingAtlas(),
-                new ConnectivityCheck(ConfigurationResolver.emptyConfiguration()));
-        this.verifier.globallyVerify(flags -> Assert.assertEquals(1, flags.size()));
-    }
+    /*
```
atiannicelli, line 58: I think we should probably remove this (remove commented section) to simplify and clear up confusion. If we leave it in there people may be confused as to why they are there.

atiannicelli (dismissed):
Looks good to me. I would remove the commented out sections, though. 

Also, just to be clear, there is one type of connectivity that we will no longer be detected if we make this change:
- When a node in a road comes close to another node in another road but the two ways don't intersect.

I think that this is GOOD and is specifically why we started making this change because we noticed that parallel bridges were being flagged incorrectly, but I just wanted to call it out.

atiannicelli:
Also, @vladlemberg do you have any stats about how many issues the new version of the check finds compared to the original?
Looks like currently there are 4253 flags in USA. How many after the change? (You can pick any country - just looking for stats.)

vladlemberg:
@atiannicelli, @Bentleysb, @sayas01, 
Number of cases in USA have been reduced from 4K+ to 44! with the new logic. I have created private MR challenge. Please let me know if you want to take a look. Not sure whether or not that massive drop is good. 

My plan is to create MR with the reversed logic to ensure that it only comprises:
1) parallel ways.
2) very closed ways that should not be connected.
3) crossings (covered by edgeCrossingEdge)  

Please let me know if you have other thoughts.

atiannicelli (approved):
Looks good to me! Nice to see this get a bit more focussed.

Bentleysb:
The unit test move is great, but has any analysis been done to check that the dropped flags are picked up my EdgeCrossingEdgeCheck?

atiannicelli:
@Bentleysb we were thinking that if the tests that we moved over were caught by the edgecrossingedgecheck then that would be good enough. It is difficult to compare the two checks since they might flag a different OSM ID for each issue. For instance if there is an issue between way X and Way Y then the old connectivity check may have flagged way X and the edgeCrossingEdgecheck may have flagged Way Y. Do you have any suggestions about how to determine if the ECE Check is covering the issues that used to be flagged by the old connectivity check?

Bentleysb:
@atiannicelli, I would suggest using the [Log Diff Command](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/commands/AtlasChecksLogDiffSubCommand.java) to compare a run with and without the changes. Once you have the diffs from the runs compare the CC and ECE diffs geographically to see if they are ruffly in the same spots. Then spot check flags to see if the nodes flagged by CC have an Edge they're connected to flagged in ECE. This doesn't need to be be comprehensive, but it would be nice have a once over the diffs of real world data. There are just too many weird scenarios in OSM for unit tests to cover everything.

Bentleysb:
Closing as a newer PR was merged.

PR #617, opened 2021-08-30
GenericTagCheck: Fail fast when databases cannot be read

tsmock:
### Description:

GenericTagCheck: Fail fast when databases cannot be read

In the interest of allowing people to not use the TagInfo db, i.e., if they
want to find tags without entries in wikidata, there are some new
*undocumented* options.

### Potential Impact:

People who depend upon GenericTagsCheck to always run will have to add an undocumented config option to the configuration json.

### Unit Test Approach:

N/A

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
@@ -178,9 +178,10 @@ public class GenericTagCheck extends BaseCheck<String>
         this.wikiTable = DEFAULT_WIKI_TABLE;
         this.tagInfoTagTable = DEFAULT_TAGINFO_TAG_TABLE;
         this.tagInfoKeyTable = DEFAULT_TAGINFO_KEY_TABLE;
-        this.tagInfoDB = this.configurationValue(configuration, "db.taginfo", DEFAULT_TAGINFO_DB);
-        this.wikiDataDB = this.configurationValue(configuration, "db.wikidata",
-                DEFAULT_WIKIDATA_DB);
+        this.tagInfoDB = this.configurationValue(configuration, "database.taginfo",
+                this.configurationValue(configuration, "db.taginfo", DEFAULT_TAGINFO_DB));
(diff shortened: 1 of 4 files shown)


atiannicelli (dismissed):

atiannicelli:
@Bentleysb this PR will prevent the issue that we had over the weekend (when a misconfiguration of the genericTagCheck caused hundreds of millions of atlas checks flags to be produced) from happening again.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
```diff
@@ -201,6 +205,23 @@ public GenericTagCheck(final Configuration configuration, final ExternalDataFetc
...
+
+        /*
+         * Prevent this check from running, unless the user has indicated that not all databases are
+         * required Specifically leave these configuration values undocumented -- if people want to
+         * use only one or the other, they should understand all the issues involved. Namely, they
+         * may have many issues where they are told "key is not documented or popular" even if it is
+         * popular (i.e., they ran with wiki data but no tag info). Regardless, wiki data is
+         * *always* required.
+         */
+        final boolean errorIfDbNotFound = this.configurationValue(configuration, "db.require_all",
```
Bentleysb, line 217:
```suggestion
        final boolean errorIfDatabaseNotFound = this.configurationValue(configuration, "database.require_all",
```

Bentleysb (commented):
Looks good. Just one nit on naming.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/GenericTagCheck.java
tsmock: The first one is done.
The second one I'm not doing (the configuration file already uses `db` as a key), at least until I get feedback from users.

@atiannicelli : Are you OK with updating the `db` key in the configuration file to `database`?

atiannicelli (approved):

Bentleysb (approved):

sayas01 (approved):

rheatreena (approved):

PR #624, opened 2021-09-21
[Check Improvement] UnusualLayerTagsCheck

### LINKED ISSUE #611: [Check Improvement] UnusualLayerTagsCheck
**Description**
The improvement of the UnusualLayerTagsCheck. Everything below is the addition to the check. Checking if the object with key:layer less or greater than 0 has incorrect tags. 

**Requirements**
 - Is a Way or Node
 - Layer tag exists and is not 0

**Use Cases**
*Case 1: [Landuse feature not on ground.]*
 - If the object has a landuse tag. Landuse is used to describe "the primary use of land by humans."
 - [Way: 94449954](https://www.openstreetmap.org/way/94449954)
image

*Case 2: [Natural feature underground.]*
 - If the tag layer is less than 0 and the tag natural is not "water"
 - [Way: 25314718](https://www.openstreetmap.org/way/25314718)
image

*Case 3: [Highway underground and no tunnel.]*
 - Layer tag is less than 0, highway tag is not "steps", the indoor tag doesn't exist or is "no" and the tunnel tag doesn't exist or is "no".
 - [Way: 121520872](https://www.openstreetmap.org/way/121520872)
image

*Case 4: [Long highway underground and no tunnel.]*
 - The same idea as Case 3 but only if the Way is more than 3 nodes
 - [Way: 587740859](https://www.openstreetmap.org/way/587740859)
image

*Case 5: [Highway above ground and no bridge.]*
 - Layer tag is above 0, highway tag exists and is not "steps", indoor tag doesn't exist or is "no", bridge tag doesn't exist or is "no", and man_made tag is not "pier"
 - [Way: 194101434](https://www.openstreetmap.org/way/194101434)
image

*Case 6: [Long highway above ground and no bridge.]*
 - Same as Case 5 but the way has more than 3 nodes
 - [Way: 356910749](https://www.openstreetmap.org/way/356910749)
image

*Case 7: [Waterway underground and no tunnel.]*
 - The layer is less than 0, the waterway tag exists or the natural tag is "water", the tunnel tag doesn't exist or is "no", and the location tag doesn't exist or is not "underground."
 - [Way: 290833931](https://www.openstreetmap.org/way/290833931) 
image

*Case 8: [Long waterway underground and no tunnel.]*
 - Same as Case 7 but only for way with over 3 nodes
 - [Way: 23066184](https://www.openstreetmap.org/way/23066184)
image

*Case 9: [Waterway above ground and no bridge.]*
 - The layer tag is above 0, the bridge tag doesn't exist or is "no", and the waterway tag exists or the natural tag is "water".
 - [Way: 166798321](https://www.openstreetmap.org/way/166798321)
image

*Case 10: [Long waterway above ground and no bridge.]*
 - Same as Case 9 but for ways with over 3 nodes
 - [Way: 686335070](https://www.openstreetmap.org/way/686335070)
image

**References**
[OSM Wiki Issue Page](https://wiki.openstreetmap.org/wiki/Osmose/issues#4110)
[Osmose backend](https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagRemove_Layer.py)

nelkapelmenka:
Fixes #611 
### Description:

The improvement of the UnusualLayerTagsCheck. Everything below is the addition to the check. Checking if the object with key:layer less or greater than 0 has incorrect tags, or if "layer=0". 

### Test Results:

There were at least 2 versions of FPR created by the editorial team. The final FP (v3) was conducted by me, as it was easier to check each case specifically. I have found no False Positives. However, I know for sure that there are some because TunnelTag doesn't have the value _FLOODED_ in Atlas . It only affects 0-1% of all results in the countries that I ran except for EGY that contains around 8% of all data FP values due to the missing value. 

I have submitted the PR to add the value to the TunnelTag. If it is not approved, I can manually filter out that tag.


nelkapelmenka:
When I run the tests locally - they are successful, however, they are failing in the PR.

Bentleysb:
@nelkapelmenka, I've seen that error on one other build today and am not sure why it is showing up intermittently. I'll try rerunning the PR build in a little bit.
In the mean time, can you fix the bugs and code smells reported by sonar?

nelkapelmenka:
@Bentleysb the only SonarLint bugs that are left are "checking that the value is present before accessing it". However, I am checking it and I don't understand why it is still a bug. Here is an example where SonarLint shows an error: `(isTagValueValid && NaturalTag.get(object).isPresent()
                && !NaturalTag.get(object).get().equals(NaturalTag.WATER)
                && (!this.objectIsTunnel(object) && !this.objectIsBridge(object)
                        && !this.objectIsCovered(object)))`

Is there a different way to go around it?

Bentleysb:
Oh, right the sonar comment has not updated because the build broke. Sorry, that is me being stupid. 
As for the check before accessing, I think it is complaining because you are checking a different Optional instance than you are getting from. You would need to refactor `NaturalTag.get(object)` out into a variable that you can then call `isPresent` and `get` on.

nelkapelmenka:
@Bentleysb That seems to have worked! Thank you

Bentleysb (dismissed):
Looks good to me. Thanks for the great additions, and refactors where needed. I expect the flooded value will get added for the tunnel tag, so I'm not worried about the EGY numbers.

src/main/java/org/openstreetmap/atlas/checks/validation/tag/UnusualLayerTagsCheck.java
```diff
@@ -109,45 +124,223 @@ public boolean validCheckForObject(final AtlasObject object)
...
+        {
+            return Optional
+                    .of(this.createFlag(object, this.naturalNotOnGround(object, isTagValueValid)));
+        }
+        else if (!this.highwayNotOnGround(object, isTagValueValid).isEmpty())
+        {
+            return Optional
+                    .of(this.createFlag(object, this.highwayNotOnGround(object, isTagValueValid)));
+        }
+        else if (!this.waterNotOnGround(object, isTagValueValid).isEmpty())
```
sayas01, line 147: Looks like this approach is unnecessarily complicated. Couldn't we just check if the condition is true and then pass the necessary instruction  within the braces? For example, `else if (this.waterNotOnGround(object, isTagValueValid)){return Optional
                    .of(this.createFlag(object, WATERWAY_NOT_ON_GROUND_INSTRUCTION));}` and the waterNotOnGround will look like the follow:
`private String waterNotOnGround(final AtlasObject object, final boolean isTagValueValid)
    {
        final Optional<NaturalTag> natural = NaturalTag.get(object);
        return (isTagValueValid
                && (WaterwayTag.get(object).isPresent()
                        || (natural.isPresent() && natural.get().equals(NaturalTag.WATER)))
                && (!this.objectIsTunnel(object) && !this.objectIsBridge(object)
                        && !this.objectIsCovered(object) && (object.tag(LocationTag.KEY) == null
                                || !object.tag(LocationTag.KEY).equalsIgnoreCase("underground"))));
    }`

sayas01: The above approach will remove the redundant calls to the same method. Also the methods called in the else if condition will look cleaner. Let me know what you think of this.

nelkapelmenka: This makes sense! I actually first did this because some methods would return two different instructions (ie, highway above ground and highway below ground), however, since then I combined the cases. I will modify my code to look similar to above.

nelkapelmenka:
@sayas01 all done!

Bentleysb (approved):

src/main/java/org/openstreetmap/atlas/checks/validation/tag/UnusualLayerTagsCheck.java
sayas01: Got it, thanks for the clarification!

sayas01 (approved):
Thanks for addressing the comments @nelkapelmenka ! LGTM!

PR #634, opened 2021-10-29
Roundabout Highway Tag Check - Osmose3010

elaineseattle:
### Description:

This PR is  for the Highway Roundabout Check, which is related to OSMOSE issue 3010.  

OSMOSE 3010 contains 3 cases:

Case 1: "Wrong highway on roundabout":  
 it must match the highest level of connected routes, except highway and trunk. 
adjust the tag highway=* of roundabout. 
https://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout#Selection_of_the_right_highway_tag 

Living examples: 

1) This roundabout [id:127882850] (https://www.openstreetmap.org/way/127882850) has incorrect highway tag. 
2) This roundabout [id:56007062] (https://www.openstreetmap.org/way/56007062) has incorrect highway tag. 

Case 3: "Roundabout shortcut": several roads connect to one node of a roundabout. In the case input and output flow of vehicles bypassing the priority rules and traffic. The fix is to separate the junction nodes into several separate ones. 

Case 4: "Roundabout crossing": way through the roundabout without stopping. 

Both case 3 and case 4 are already covered by roundaboutValenceCheck in which roundabouts are flagged valence >= 2. Thanks to Bentley for pointing out the existing work. In fact, this PR borrows a lot of the SimpleEdgeWalker concept used in roundaboutValenceCheck.

Requirements 

Case 1: Roundabout should have tag highway=* that is equal to the highest level of classified connected routes. True positive if true. 

The importance of the highway is determined by the order shown in https://wiki.openstreetmap.org/wiki/Key:highway 
Only classified highways are considered in the OSMOSE issue.
 

We follow the highway order in atlas.(https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/HighwayTag.java) 

### Test Results: (Coming soon)

|  |    |         |     0%       |    |    |    |


brianjor:
@Bentleysb this can be closed, due to #662 being merged

Issue #408, opened 2020-11-03
[Fix Suggestions] Intersecting Buildings Check


sayas01:
**Which Atlas check is the fix suggestion related to?** 

**IntersectingBuildingsCheck:**
Description: https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/intersectingBuildingsCheck.md

Source Code: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/IntersectingBuildingsCheck.java

**Describe more details of suggested fix(es)**

*Suggested Fix 1*: 

- Type: Geometry

- Operation: Delete

- Action:  Delete one of the buildings that are equal/ fully contained by another building.
The check flags three case:
1) buildings intersecting one another
2) buildings overlapping one another
3) buildings that are equal.

Autofix action will be to delete one of the buildings among the two overlapping buildings, that belongs to (2) and (3) cases.

**Is there any code enhancement needed if adding the fix suggestion component?**

No, the three cases are already flagged individually.

brianjor:
I've started working on this one.

PR #640, opened 2021-11-16
[Fix Suggestions] Intersecting Buildings Check

brianjor:
### Description:

See: #408 

### Potential Impact:

N/A

### Unit Test Approach:

Added asserts for correct building OSM id being in the fix suggestion.

Resolves: #408


sayas01 (approved):

Bentleysb (approved):

atiannicelli:
This scares me a bit as it suggests to the mapper to delete a building instead of fixing the place where the intersection happens. I just wonder if when a new mapper starts going through this and starts deleting buildings and then the OSM community traces back where they got the idea to delete buildings from they will see that atlas checks was suggesting it. The code looks great - I guess I'm more arguing that #408 might not be the best way to go. I would also think we would want to add something in the instructions that says to only delete features if they are obviously incorrect or duplicates.

PR #643, opened 2021-12-12
[new] Highway Intersection Check

### LINKED ISSUE #618: [New Check]HighwayIntersectionCheck
**Description**
Checks cases where a highway intersects power or waterway objects.

**Requirements**
1. Check Candidate (example)
 - Is an `Edge`
 - Has `highway` tag

**Use Cases**
Example:

*Case 1: Intersects water (Vietnam)*
 - A stream shares a node with a primary road 
 - https://www.openstreetmap.org/edit?editor=id&way=146071043#map=18/11.00595/106.66726

image

*Case 2: Intersects power line (Vietnam)*
 - A power line tower is also node on a road
 - https://www.openstreetmap.org/edit?editor=id&way=226103327#map=19/11.07112/106.66167
 
image

**Further investigation**
In case of waterway objects, ways considered ford or `leisure`=`slipway` are not to be flagged.

**Supported regions**
All countries

-

**References**
https://wiki.openstreetmap.org/wiki/Tag:power%3Dline

martapako:
### Description:
close #618

The purpose of this check is to flag highways intersecting power or waterway objects.

- **dam** and **weir** waterway objects are not flagged, these type of waterways can intersect with highways.

- for waterway objects the intersection with a highway having **leisure=slipway** or **ford=yes** tag is allowed, and they are not flagged.

### Potential Impact:

No downstream impact.


Bentleysb:
@martapako, have you done any analysis on the validity of the flags produced by this check? If you have, it would be great if you could add some stats on what you have done and found to the PR description, similar to the test results section in this PR: https://github.com/osmlab/atlas-checks/pull/639
If not, then could you sample through some results from around the world and report back? We just want to have some testing to show that the crazy data that is in OSM isn't throwing something wacky at this new check.

Issue #653, opened 2022-01-14
[New Check] Separate Sidewalk Tag Check


vladlemberg:
**Description**
A check to validate that when sidewalk=* tags are used on a highway that any separately mapped sidewalk(s) are consistent with the highway’s sidewalk tags. It is important to note here that separately mapped highways is still a proposal (https://wiki.openstreetmap.org/wiki/Proposed_features/sidewalk_schema). The use of separately mapped sidewalks is not a requirement and this check will not ask mappers to add separate sidewalk ways unless the tags specify “separate” sidewalks. Existence of sidewalk=both|left|right highway tags AND separate sidewalks is discouraged, but not illegal, but this check will ask mappers to alter sidewalk tags on the way to be sidewalk=separate if there are separate sidewalks in the area.

**Requirements**
An Atlas Object to be flag must meet the following criteria to be considered:

* Object must be a car-navigable way
* Object must have sidewalk=* tag.
* Object’s sidewalk tag is inconsistent with surrounding sidewalks.
    * Sidewalks associated with the Highway Object in question must be approximately parallel to the highway Object and within 15 meters * <# of lanes>.
    * When determining if sidewalk tags are consistent with surrounding separate sidewalks use the following table. 

<img width="707" alt="image" src="https://user-images.githubusercontent.com/62575608/149589725-40647a52-fd17-46c7-9858-e7dbbb240a02.png">

1. A ‘1’ indicates an invalid combination that should be flagged as an issue. Text used in these situations should be as shown below.
    1. “The way <XYZ> has sidewalk tags that conflict with separately mapped sidewalks near this way. The way indicates that sidewalks are <sidewalk tag value>, but separately mapped sidewalks were detected on <detected sidewalk sides>. Please correct the separately mapped sidewalks and we recommend changing the sidewalk tag on this way to `sidewalk=separate`."
2. A ‘2’ indicates that the way tags specifically indicate separate sidewalks should be mapped, and the separate ways that are detected do not match with the way’s sidewalk tags. Text used in these situations is similar to number 1 but a bit different to call out the difference.
    1. “The way <XYZ> has sidewalk tags that conflict with separately mapped sidewalks near this way. The way indicates that sidewalks are <sidewalk tag value>, but separately mapped sidewalks were detected on <detected sidewalk sides>. Please correct the separately mapped sidewalks or update the ways sidewalk tags to be consistent.”
[...]

PR #654, opened 2022-01-24
SeparateSidewalkTagCheck [New Atlas Check] 

vladlemberg:
### Description:

Please refer to purpose and requirements: https://github.com/osmlab/atlas-checks/issues/653 

### Potential Impact:

List potential impact to downstream libraries here (ex. atlas-checks, atlas-generator).

### Unit Test Approach:

Describe unit tests being added with this PR.

### Test Results:

|  ARE   |    109         |    33      |     30%       |   24 |   9 | 27%   |

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
@@ -1,0 +1,274 @@
+package org.openstreetmap.atlas.checks.validation.linear.edges;
+
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+import org.openstreetmap.atlas.checks.base.BaseCheck;
+import org.openstreetmap.atlas.checks.flag.CheckFlag;
+import org.openstreetmap.atlas.geography.Heading;
+import org.openstreetmap.atlas.geography.Location;
+import org.openstreetmap.atlas.geography.PolyLine;
+import org.openstreetmap.atlas.geography.Segment;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasObject;
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.atlas.walker.OsmWayWalker;
+import org.openstreetmap.atlas.tags.FootTag;
+import org.openstreetmap.atlas.tags.FootwayTag;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.LayerTag;
+import org.openstreetmap.atlas.tags.SidewalkLeftTag;
+import org.openstreetmap.atlas.tags.SidewalkRightTag;
+import org.openstreetmap.atlas.tags.SidewalkTag;
+import org.openstreetmap.atlas.tags.oneway.OneWayTag;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.scalars.Angle;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+import avro.shaded.com.google.common.collect.Range;
+import avro.shaded.com.google.common.collect.Ranges;
+
+/**
+ * This check flags when sidewalk=* tags are used on a highway that any separately mapped
+ * sidewalk(s) are consistent with the highway’s sidewalk tags.
+ *
+ * @author Vladimir Lemberg
+ */
+public class SeparateSidewalkTagCheck extends BaseCheck<Long>
+{
+    private static final String INSTRUCTION_FORMAT = "Way {0,number,#} is tagged as sidewalk={1} but separately mapped sidewalks were detected that are not consistent with the way's sidewalk tag. Verify that the sidewalk tag for this way is correct and consistent with separately mapped ways for the entirety of the way.";
+    private static final List<String> FALLBACK_INSTRUCTIONS = Collections
+            .singletonList(INSTRUCTION_FORMAT);
+    private static final Double SIDEWALK_SEARCH_DISTANCE_DEFAULT = 15.0;
+    private static final Double EDGE_LENGTH_DEFAULT = 20.0;
+    private static final Range<Double> HEADING_DEGREE_RANGE = Ranges.closed(-20.0, 20.0);
+    private static final String MAXIMUM_HIGHWAY_DEFAULT = HighwayTag.PRIMARY.toString();
+    private final Distance searchDistance;
+    private final Distance defaultEdgeLength;
+    private final HighwayTag maximumHighwayType;
+    private final int primeMeridian = 180;
+
+    public SeparateSidewalkTagCheck(final Configuration configuration)
+    {
+        super(configuration);
+        this.searchDistance = configurationValue(configuration, "sidewalk.search.distance",
+                SIDEWALK_SEARCH_DISTANCE_DEFAULT, Distance::meters);
+        this.defaultEdgeLength = configurationValue(configuration, "edge.length",
+                EDGE_LENGTH_DEFAULT, Distance::meters);
+        this.maximumHighwayType = this.configurationValue(configuration, "maximum.highway.type",
+                MAXIMUM_HIGHWAY_DEFAULT, str -> Enum.valueOf(HighwayTag.class, str.toUpperCase()));
+    }
+
+    @Override
+    public boolean validCheckForObject(final AtlasObject object)
+    {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,309 @@
...
+    protected List<String> getFallbackInstructions()
+    {
+        return FALLBACK_INSTRUCTIONS;
+    }
+
+    /**
+     * Find the closest point between line (
+     *
+     * @param line
+     *            the ocean feature offending streets/railways offending buildings
```
atiannicelli, line 123: comment doesn't seem right

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,309 @@
...
+                        || edge.getTag(SidewalkLeftTag.KEY).isEmpty()))
+        {
+            sidewalkTagValue = "right";
+        }
+
+        return sidewalkTagValue;
+    }
+
+    /**
+     * Verifies that separate sidewalk and highway are within acceptable heading degree range
```
atiannicelli, line 193: what does it mean to be in acceptable heading range?

atiannicelli (commented):
Just a couple questions/comments - I will look into the code closer soon.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,309 @@
...
+                        .closestSegmentToPoint(sidewalk.asPolyLine(), edgeMidPoint);
+
+                if (this.isCrossing(edge.asPolyLine(), sidewalk.asPolyLine())
+                        || !LayerTag.areOnSameLayer(edge, sidewalk))
+                {
+                    continue;
+                }
+
+                if ("right".equals(sidewalkTagValue)
+                        && !this.isRightOf(edge.asPolyLine(), closestSidewalkSegment.middle())
```
atiannicelli, line 101: I think I would change this logic a tiny bit so that:
1) you call the isRightOf function once for each sidewalk and keep a count of lefts and rights. (if isRightOf then right_count++ else left_count++ )
2) then (outside of the loop) you check for all three conditions.
    "right".equals(sidewalkTagValue) && right_count > 0
    "left".equals(sidewalkTagValue) && left_count > 0
    "both".equals(sidewalkTagValue) && right_count > 0 && left_count > 0

The reason for this change is so that you don't miss flagging the situation where there are two sidewalks detected on the left and the "both" tag is set. This should be flagged but your logic below would not flag it since there are two sidewalks detected nearby. Also that would prevent you from calling isRightOf multiple times per sidewalk.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Verifies that separate sidewalk and highway are within acceptable heading degree range
+     *
+     * @param headingOne
+     *            {@link Heading} one to check
+     * @param headingTwo
+     *            {@link Heading} to check.
+     * @return {@code true} if both heading are within acceptable heading degree range
```
atiannicelli, line 199: maybe explain why this is used? I don't remember, but I remember you told me and it made sense.

atiannicelli (commented):
Just one requested change on the logic. The rest looks great.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
vladlemberg: that means highway and separate sidewalk are more or less parallel

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,343 @@
...
+
+    private boolean isSidewalkTaggingMismatch(final String sidewalkTagValue,
+            final int rightSidewalkCount, final int leftSidewalkCount)
+    {
+        switch (sidewalkTagValue)
+        {
+            // sidewalk=right but separate sidewalks on left or both sides are detected
+            case "right":
+                return rightSidewalkCount < 1 && leftSidewalkCount > 0
+                        || rightSidewalkCount > 0 && leftSidewalkCount > 0;
```
atiannicelli, line 289: Can't this be simplified to just check for left > 0?
```suggestion
                return leftSidewalkCount > 0;
```

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,343 @@
...
+            // sidewalk=left but separate sidewalks on right or both sides are detected
+            case "left":
+                return leftSidewalkCount < 1 && rightSidewalkCount > 0
+                        || leftSidewalkCount > 0 && rightSidewalkCount > 0;
```
atiannicelli, line 293: same here.
```suggestion
                return rightSidewalkCount > 0;
```

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,343 @@
...
+            // sidewalk=both but separate sidewalk on left or ride side is detected.
+            case "both":
+                return rightSidewalkCount < 1 || leftSidewalkCount < 1;
```
atiannicelli, line 296: I think this needs to check for and allow the case where right and left count are 0.
```suggestion
                return (rightSidewalkCount < 1 && leftSidewalkCount > 0)
                        || (rightSidewalkCount > 0 && leftSidewalkCount < 1);
```

vladlemberg:
Hello Everyone, I believe SeparateSidewalkCheck is ready for review. Few things:

1) I was unable to create a unit tests for alternative sidewalk mapping: sidewalk:left and sidewalk:right cases. Those tags were introduced in 6.6.1 Atlas version. Perhaps unit test suite can't understand these tags. Example: `invalidHighwayRightSidewalkLeftAlternativeMapping;`

2) Check was tested against USA region only.

vladlemberg:
@Bentleysb, @atiannicelli I'm confirming that unit tests for alternative Sidewalk mapping is broken because of new tags "sidewalk:left" and "sidewalk:right". I believe something is wrong with creating test atlas with new tags. I'm proposing to remove alternative sidewalk mapping for this PR. What would you suggest ?

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,341 @@
...
+     * existence with other
+     *
+     * @param edge
+     *            Edge to examine
+     * @return {@code true} if {@link Edge} sidewalk tag meet the criteria.
+     */
+    @SuppressWarnings("squid:S3655")
+    private boolean validSidewalkFilter(final Edge edge)
+    {
+        return edge.getTag(SidewalkTag.KEY).isPresent() && !this.isDualCarriageWay(edge)
```
Bentleysb, line 332: @vladlemberg, your testInvalidHighwayLeftSidewalkRightAlternativeMapping is failing because `edge.getTag(SidewalkTag.KEY).isPresent()` is false for all features in its test atlas.

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SeparateSidewalkTagCheck.java
```diff
@@ -0,0 +1,347 @@
...
+     *            The line to compare the location to
+     * @param location
+     *            The location
+     * @return {@code true} if the location is to the right of the line.
+     */
+    private boolean isRightOf(final PolyLine line, final Location location)
+    {
+        final PolyLine locationLine = new PolyLine(location);
+        final Segment closest = line.segments().stream()
+                .min(Comparator.comparingDouble(s -> s.shortestDistanceTo(locationLine).asMeters()))
```
Bentleysb, line 275: nit
```suggestion
                .min(Comparator.comparingDouble(segment -> segment.shortestDistanceTo(locationLine).asMeters()))
```

Bentleysb (approved):
Looks good to me.

pscrimshaw:
@atiannicelli please feel free to review when you get the chance. Thank you!

atiannicelli (approved):
Looks great to me!

PR #655, opened 2022-02-11
MapRouletteUploadCommand Challenge Country Name Parameter

Bentleysb:
### Description:

This adds a new optional parameter to the MapRouletteUploadCommand to select whether to use the ISO code or country display name in uploaded challenge titles. 
Currently challenge titles are created with the pattern "Country Display Name - Check Name". When this new parameter is true, which it is by default, this same pattern will be used. When the parameter is false the pattern will be "ISO - Check Name".

Additionally this adds some missing docs, and bumps the atlas version.

### Potential Impact:
This should be fully backwards compatible. 

### Unit Test Approach:

Added a unit test for when the param is false. 

### Test Results:
Ran three test runs with the param true, false, and not given. 

True:
Screen Shot 2022-02-11 at 2 45 38 PM

False:
Screen Shot 2022-02-11 at 2 45 27 PM

Default:
Screen Shot 2022-02-11 at 2 45 21 PM

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 project.ext.versions = [
     checkstyle: '8.18',
     jacoco: '0.8.7',
-    atlas: '6.5.7',
+    atlas: '6.6.2',
     commons:'2.6',
     atlas_generator: '5.3.6',
     atlas_checkstyle: '5.6.9',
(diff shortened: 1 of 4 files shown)


MonicaBrandeis (approved):
Looks good, LGTM

dcastrowa (approved):
Tested for bi monthly data request https://hawkeye.geo.apple.com/repos/143/dags/checks_maproulette_upload_flags/runs/48077. Thanks for these changes!

PR #664, opened 2022-03-15
Highway Intersection Check

brianjor:
### Description:

Picked up where #643 left off.

Changes are:
* Flagging the highway and the node(s) that intersects with the powerline/waterway instead of the powerline/waterway
* Flags the way instead of just the edge

### Test Results:

|  ARG   |    10         |    10      |     100%       |   10 |   0 | 0%   |
|  AUS |    9         |    9      |     100%       |   9 |   0 | 0%   |
|  TUR |    18         |    18      |     100%       |   18 |   0 | 0%   |

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/HighwayIntersectionCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/HighwayIntersectionCheck.java
@@ -36,7 +36,7 @@ import org.openstreetmap.atlas.utilities.configuration.Configuration;
 public class HighwayIntersectionCheck extends BaseCheck<Long>
 {
     private static final long serialVersionUID = -2100623356724302728L;
-    private static final String INSTRUCTION_FORMAT = "The way with id {0,number,#} has invalid intersections with {1}. A navigable way should not share nodes with non-navigable features. Either the way is inproperly tagged or is a combination of what should be two separate ways (highway and the other non-navigable feature).";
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/intersections/HighwayIntersectionCheck.java
```diff
@@ -0,0 +1,197 @@
...
+                .flatMap(Set::stream)
+                .filter(connectedEdge -> !this.hasSameOSMId(connectedEdge, edge));
+
+        final Set<Edge> invalidconnectingEdges = edgesConnectedToWay
+                .filter(connectedEdge -> TagPredicates.IS_POWER_LINE.test(connectedEdge)
+                        || !FordTag.isYes(edge))
+                .filter(connectedEdge -> TagPredicates.IS_POWER_LINE.test(connectedEdge)
+                        || this.isWaterwayToCheck(connectedEdge))
+                .filter(connectedEdge -> TagPredicates.IS_POWER_LINE.test(connectedEdge)
+                        || !this.isSlipway(edge))
```
Bentleysb, line 71: It seems like there is a lot of redundancy here, caused by having three separate filter calls.

brianjor:
> I'd be interested to see some examples of the true positive flags for this check. I am surprised there are Edges with power line or waterway tags. Those features generally become Lines.

Here is an example of a highway that has the power=line tag on it that also intersects another highway.
image
The selected way has both highway and power and intersects the "Ruta Nacional 156" 
The way "Ruta Nacional 156" is the way that gets tagged.
https://www.openstreetmap.org/edit#map=17/-32.22399/-63.17634

rheatreena (approved):

FamTal (approved):

Issue #597, opened 2021-08-06
[Fix Suggestions] AddressStreetNameCheck to consider abbreviations


herriotto:
AddressStreetNameCheck in Venezuela market has tags addr:street=Av.5 which refers to 'Avenida 5', such as here -
https://maproulette.org/challenge/18614/task/95888925 

another example is addr:street=Av. Bolivar, referring to 'Avenida Bolivar' identified in the AtlasCheck here - 
https://maproulette.org/challenge/18614/task/95891518

Are abbreviations considered/under discussion for this check?

Bentleysb:
Hi @herriotto. 
The check does not currently consider abbreviations at all, and as far as I'm aware this is the first discussion of it. 
The point of the check is to make sure that the address and surrounding name tags match, which includes identifying typos and such. So I would think that it is legitimate to flag addresses and street names that are not abbreviated in the same manor.

Bentleysb:
Actually, I take back my statement about this being the first discussion about abbreviations in AddressStreetNameCheck. I forgot about the discussion here: https://github.com/osmlab/atlas-checks/issues/409
There it has been discussed adding a standardization library to help deal with abbreviations. However, in that case it was for creating suggested fixes.

vladlemberg:
@Bentleysb, please assign this task to me.

Issue #409, opened 2020-11-03
[Fix Suggestions] Address Street Name Check


sayas01:
**Which Atlas check is the fix suggestion related to?** 

**AddressStreetNameCheck:**
Description:https://github.com/osmlab/atlas-checks/blob/dev/docs/checks/addressStreetNameCheck.md

Source Code: https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/points/AddressStreetNameCheck.java

**Describe more details of suggested fix(es)**

*Suggested Fix 1*: 

- Type: Tag

- Operation: Modify

- Action: Update the `addr:street` tag value of the point to the value of Name tag of adjacent edge. 

**Is there any code enhancement needed if adding the fix suggestion component?**

Yes, the AutoFix suggestion should be added only in the case where the total number of edges with Name tag within the given bounds is 1 and not for all flags.

If yes, provide details about what and where the enhancement is needed.  Ex. the fix requires excluding checks on relation

vladlemberg:
@sayas01, please assign it to me. thanks!

vladlemberg:
@sayas01, @Bentleysb, @atiannicelli,
I tested AddressStreetNameCheck output against US Metro Areas. Here is my observation: vast majority of flags are POIs locating in large plazas or shopping centers with physical addresses beyond default radius of 100m. The second group of flags are Address Points inside large apartment complexes. All of them most likely are FPs.  

Per our discussion with @Bentleysb, the Check is not ready for AutoFix and need additional tuning. Based on my observation, it should be some logical tuning instead of increasing radius distance. 

For example: Road Name Normalization. 
<img width="637" alt="image" src="https://user-images.githubusercontent.com/62575608/109885542-83b84500-7c33-11eb-8ed0-2db7de9f0867.png">

I've created MR challenges with different radius: 100, 500, 1000 meters that can be passed for further investigation. 
https://maproulette.org/admin/project/43124

@sayas01, @Bentleysb, please advise the next step. Do we need to pass validation to your QA team ?

Bentleysb:
@vladlemberg, what percentage of the false positives you observed would you say fall under the road name normalization use case? I think part of the original impetus for this check was to flag these spelling and capitalization discrepancies, as they can cause problems for routing engines. 
I know there was also one use case you shared with me where the false positive was caused by the search distance. Are there any other general trends you have found in the false positives?

vladlemberg:
[...]

PR #667, opened 2022-03-22
[New Atlas Check] AbbreviatedAddressStreetCheck 

vladlemberg:
### Description:

A check to validate that  Address Street `addr:street` road type  abbreviation is spelled out in full.  Please reference official OSM [good practice](https://wiki.openstreetmap.org/wiki/Names#Abbreviation_.28don.27t_do_it.29) mapping conventions.

Corresponded Issues: 
https://github.com/osmlab/atlas-checks/issues/597
https://github.com/osmlab/atlas-checks/issues/409

### Potential Impact:

* Name Standardization.
* Improving Geocoding service. 

### Unit Test Approach:

Simulate unit tests covering most common address types and false positives.     

### Test Results:

Passed. 

|  USA   |    50,000         |    5960      |     12%       |   5912 |   48 | 0.8%   |

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,0 +1,175 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer
+
+plugins {
+	id 'java'
+	id 'maven'
+	id 'maven-publish'
+	id 'idea'
+	id 'eclipse'
+	id 'signing'
+	id 'checkstyle'
+	id 'jacoco'
+	id "com.diffplug.gradle.spotless" version "3.27.0"
+	id 'org.sonarqube' version '2.8'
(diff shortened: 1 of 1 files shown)


Bentleysb (commented):
This looks great. It looks like there is just a bit of configuration and sonar clean up needed. Should this only be enabled in English speaking countries, for the moment?

atiannicelli:
This looks great @vladlemberg. Thanks! I agree with @Bentleysb that we should add a line to the configuration file that restricts this to a set of predominantly English speaking countries.  

```
    "countries": [
      "USA", "GRB", "AUS", "CAN", "JAM", "IRL", <others?>
    ],
```

@Bentleysb - Do you think that having the street name txt file is the best way to store and access that information? It seems fine with me, but I know that we were curious if that would be acceptable, architecture wise.

atiannicelli (approved):
This looks great to me.

rheatreena (approved):

PR #673, opened 2022-04-22
EdgeCrossingEdgeCheck Rework for Execution Time

Bentleysb:
### Description:

For some time now the EdgeCrossingEdgeCheck has suffered from very poor efficiency in processing features that have a significant number of invalid intersections. An example is this feature at version 5: https://www.openstreetmap.org/way/1043498633. This has regular increased the overall runtime for checks execution to 6+ times the time it normally takes. 

This stemmed from the way the check would walk a network of invalid intersections a number of times equal to the number of edges in the network. This design was originally implemented to resolve a possible nondeterminism issue, that in review did not really exist. However, there were a couple of other beneficial things added with this. Checks were made simpler, and marked the exact point of intersections. These improvements have both been well received by editors. It also seems, though, there there was a loss of some intersections points in more complex cases. 

This rework removes the redundant network walks, and attempts to build upon the improvements made to flag clarity. The primary change seen in most flags is that all involved edges are now once again shown, along with the points marking the invalid intersections.
Old basic flag:
Screen Shot 2022-04-22 at 11 24 00 AM
New basic flag:
Screen Shot 2022-04-22 at 11 24 15 AM

To keep the simplicity of flags in cases where there are networks of intersections cover a larger area, a clustering system has been implemented to split these networks in to multiple flags. Intersections in a network are clustered based on a configurable distance, defaulting to 500 meters. Each flag is comprised of the cluster of intersection points, and only the Edges connected to those points. All clusters for a network are calculated the first time any Edge in that network is encountered. The cluster the initial Edge is in is flagged and the rest are stored for latter flagging. As Edges in the other clusters are encountered, the pre-calculated clusters are pulled from the store and flagged. 
An example of the result of this is that the following flag is split into 9 flags:
Screen Shot 2022-04-22 at 11 50 30 AM
Such as these two cluster flags:
Screen Shot 2022-04-22 at 11 50 41 AM
Screen Shot 2022-04-22 at 11 51 08 AM
And a closeup of the intersections in the second cluster:
Screen Shot 2022-04-22 at 11 51 17 AM

Finally, this rework returns some intersection that were missing due to a tie break system that was used by the old multiple walker system. 
Old flag with one intersection:
Screen Shot 2022-04-22 at 11 42 28 AM
New flag with all three intersections in the network:
Screen Shot 2022-04-22 at 11 42 18 AM

### Potential Impact:

None

### Unit Test Approach:

Updated unit tests to reflect the additional features that are now part of the flag. Added a unit test to check the clustering system works to make multiple flags. 

### Test Results:

The runtime issue was tested using an Atlas where this checks was hitting the 24 hour timeout in one case. That same Atlas now completes in about a minute, and the flags are as expected.

Additional analysis was done on the counties listed below.

|  CRI   |    9       |    9      |     100%       |   9 |   0 | 0%   |
|  SGP   |    2       |    2      |     100%       |   2 |   0 | 0%   |
|  SLV   |    24       |    24      |     100%       |   24 |   0 | 0%   |

### DIFF
--- a/config/configuration.json
+++ b/config/configuration.json
@@ -308,6 +308,7 @@
     "crossing.car.navigable": true,
     "crossing.pedestrian.navigable": false,
     "indoor.mapping": "indoor->*|highway->corridor,steps|level->*",
+    "cluster.distance": 500.0,
     "challenge": {
       "description": "Tasks contain ways that do not have shared nodes but cross each other.",
       "blurb": "Crossing Ways",
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/EdgeCrossingEdgeCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/intersections/EdgeCrossingEdgeCheck.java
@@ -1,7 +1,8 @@
 package org.openstreetmap.atlas.checks.validation.intersections;
 
-import java.util.Arrays;
-import java.util.Comparator;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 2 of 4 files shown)


vladlemberg (approved):
Looks awesome!

rheatreena (approved):

Issue #585, opened 2021-07-19
[Bug] SinkIslandCheck Flags Ferry Locked Roads


Bentleysb:
**Describe the bug**
SinkIslandCheck produces false positive flags when the only car accessible route to a small network of edges is via a ferry. 

**To Reproduce**
1. Run the Check in RUS
2. Look for flags of and near https://www.openstreetmap.org/way/603953726

**Expected behavior**
The check should consider car ferries as part of the road network. However, I don't think ferry routes need to necessarily be flagged themselves. This may need to be investigated for airport connections as well.

PR #676, opened 2022-05-25
SinkIslandCheck Ferry and Car Access Update

Bentleysb:
### Description:

This update is to resolve two issues with the SinkIslandCheck.

This fixes https://github.com/osmlab/atlas-checks/issues/585. It does this by making car ferries a valid terminus for a road network. Thus ways like the following are no longer flagged:
Screen Shot 2022-05-25 at 10 03 48 AM

This also fixes an issue with the way car navigability was derived from the vehicle, motor_vehicle, and motorcar tags. These tags act in a hierarchical fashion. This was previously ignored, so that if any of these tags had a navigable value the other's values were irrelevant. 
Example, this way was flagged because it had vehicle=yes and its motor_vehicle=permit was ignored while the connected ways just had motor_vehicle=permit:
Screen Shot 2022-05-25 at 10 10 52 AM

### Potential Impact:

None

### Unit Test Approach:

Added unit tests for car ferries, pedestrian ferries, and car access hierarchy. 

### Test Results:

Analysis was done on the difference in flags from the previous version of this check to the updates here. 

Subtracted Flags:
| ISO | Total Flags | Sampled  | Sampling % | TN | FN | Differential False Negative Rate |
|  Global   |    94        |    94      |     100%       |   88 |   6 | 6.3%   |

There were 2 types of new false negative flags introduced.
The first is when ferries connect piers with missing access tags:
https://www.openstreetmap.org/way/822360979
The second is when ferries connected roads that had misplaced one way connections:
https://www.openstreetmap.org/way/604212684

Added Flags:
| ISO | Total Flags | Sampled  | Sampling % | TP | FP | Differential False Positive Rate |
|  Global   |    287        |    29     |     10.1%       |   29 |   0 | 0%   |

The new true positives added where situations where ways were connected to roads that had something like vehicle=yes & motorcar=no. Previously these connected ways were seen as car route-able, when in fact they should not have been.
https://www.openstreetmap.org/way/784709925
https://www.openstreetmap.org/way/717145350

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
@@ -252,8 +252,8 @@ public class SinkIslandCheck extends BaseCheck<Long>
 
     /**
      * Finds the car accessibility value for an {@link Edge} by returning the first value found,
-     * using a descending list of tags to check. If no values are found fort the list of tags the
-     * value is assumed to be YES.
+     * using a descending list of tags to check. If no values are found for the list of tags the
+     * default is used.
      *
      * @param edge
      *            {@link Edge}
--- a/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
+++ b/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
@@ -257,6 +257,8 @@ public class SinkIslandCheck extends BaseCheck<Long>
      *
      * @param edge
      *            {@link Edge}
+     * @param defaultValue
+     *            {@link String} to return as the default value if no others are found
      * @return {@link String} car accessibility value
      */
     @SuppressWarnings({ "squid:S3740" })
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
```diff
@@ -251,6 +250,32 @@ private boolean endOrStartNodeHasAmenityTypeToExclude(final AtlasObject object)
                 || Validators.isOfType(edge.start(), AmenityTag.class, AmenityTag.PARKING_ENTRANCE);
     }
 
+    /**
+     * Finds the car accessibility value for an {@link Edge} by returning the first value found,
+     * using a descending list of tags to check. If no values are found fort the list of tags the
+     * value is assumed to be YES.
```
atiannicelli, line 256: The default value is passed in. Seems like the default should be "yes" like the comment says instead of passing it in. Are there times you would want the default to be something else?

atiannicelli (approved):
Looks good to me. just one small question about the default value to getCarAccess

src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/SinkIslandCheck.java
Bentleysb: I forgot to update this comment when I added the default value argument. I ended up making the default value variable because it is YES for cars and NO for ferries.

rheatreena (approved):

Issue #685, opened 2022-08-23
[Bug] Atlas date parsed incorrectly in Construction check causing crash.


atiannicelli:
It looks like the Construction check is not fetching the Atlas Date correctly. The data that this validation is running on has a name "osmbase_2022-08-20". I am not sure why it is parsing this to find the date. 

I am seeing this all over the place:

```
22/08/22 15:48:19 ERROR BaseCheck: ConstructionCheck failed on feature 508294744000000 (508294744).
java.time.format.DateTimeParseException: Text 'osmbase_2022' could not be parsed at index 0
	at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2046)
	at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1948)
	at java.base/java.time.LocalDate.parse(LocalDate.java:428)
	at org.openstreetmap.atlas.checks.validation.tag.ConstructionCheck.flag(ConstructionCheck.java:140)
	at org.openstreetmap.atlas.checks.base.BaseCheck.check(BaseCheck.java:130)
	at org.openstreetmap.atlas.checks.distributed.RunnableCheck.lambda$run$0(RunnableCheck.java:77)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.openstreetmap.atlas.checks.distributed.RunnableCheck.run(RunnableCheck.java:75)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

It looks to be fetching the name of the atlas instead of the date??
I believe that the value of atlasDateString.get() is "osmbase_2022-08-20"

        // Get the date the atlas was generated, or use today's date as a fallback
        final Optional<String> atlasDateString = object.getAtlas().metaData().getDataVersion();
        final LocalDate comparisonDate = atlasDateString.isPresent()
                && !atlasDateString.get().equals("unknown")
                        ? LocalDate.parse(atlasDateString.get().split("-")[0], ATLAS_DATE_FORMATTER)
                        : TODAYS_DATE;

To reproduce run the ConstructionCheck on any area:
(Example EMR command we use)
```
[...]

PR #686, opened 2022-08-24
Construction check: Atlas date parsed incorrectly

vladlemberg:
### Description:

ConstructionCheck is trying to parse the date from the Atlas meta data expecting original pbf name format. It can't handle custom prefixes such as "daylight" or "osmbase".  As result Check can't generate flags. For more info, please refer: https://github.com/osmlab/atlas-checks/issues/685. 

### Potential Impact:
Check will will be executed using default the "today's date".  

### Unit Test Approach:

N/A

### Test Results:

N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,13 +144,21 @@ task shaded(type: Jar) {
 }
 
 shadowJar {
-    baseName = project.name
-    mergeServiceFiles()
-    transform(com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer) {
-        resource = 'reference.conf'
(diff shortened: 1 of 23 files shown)


vladlemberg:
@Bentleysb, @atiannicelli, this approach will definitely fix the bug, however, for the ideal solution I'd move getting date logic to the BaseCheck class. in this case we don't need to check date from every AtlasObject.

Bentleysb (approved):
This looks like a good improvement to me. I'd be happy to see support for this added to BaseCheck as well.

PR #691, opened 2023-03-01
fix table

atiannicelli:
Just a small patch to fix the available_checks file so the table looks right.


Bentleysb (approved):
Thanks for this!

